PR #10, opened 2023-05-01
start work on displaying track info

toloudis:
Problem
=======
Scientists wish to see information about the tracks that cells take through the time lapse.

Solution
========
Add track data to the dataset in json format.
Implement clicking on the viewport and display track for that nucleus, for the selected feature, in a plotly plot below.  

## Type of change


toloudis:
TODO:
draw vertical line on plot at current time, and add some labeling to the plot and the rest of the app to describe the current state, improve layout/spacing

src/colorizer/loaders/JsonTracksLoader.ts
frasercl, line 1: I'm somewhat wary of proliferating new loaders for different data types. `trackIds` and `trackTimes`, though given special treatment by the client, are more similar in structure to features than they are different (both large arrays of numbers meant to be indexed by cell ID). I added outlier loading by extending the capabilities of feature loaders, and I wonder if the same can be done here. Perhaps a unified data loader could return a `DataSource` object with conversion methods to whatever representations we need (texture, typed arrays, etc.)? This would also provide us an easier time later if we want to hand track data to the GPU, e.g. to highlight segmentations on a selected track.

src/colorizer/loaders/JsonFeatureLoader.ts
```diff
@@ -15,9 +15,11 @@ export default class JsonFeatureLoader implements IFeatureLoader {
     const response = await fetch(url);
     const text = await response.text();
     const { data, min, max }: FeatureDataJson = JSON.parse(nanToNull(text));
+
     const tex = isBoolArray(data)
       ? packBooleanDataTexture(data)
       : packFloatDataTexture(data.map((val) => (val === null ? Infinity : val)));
-    return { tex, min, max };
+    const dataArr = isBoolArray(data) ? new Float32Array(data.map((value) => (value ? 1 : 0))) : new Float32Array(data);
+    return { data: dataArr, tex, min, max };
```
frasercl, line 23: Another spot where an abstract data source could be helpful. Besides being a bit awkward, the sanitization code for booleans is repeated from `packBooleanDataTexture`.

src/main.ts
```diff
@@ -194,9 +215,13 @@ async function start(): Promise<void> {
...
   featureSelectEl.addEventListener("change", handleFeatureChange);
   colorRampSelectEl.addEventListener("click", handleColorRampClick);
+  canv.domElement.addEventListener("click", handleCanvasClick);
 }
 
-window.addEventListener("beforeunload", () => canv.dispose());
+window.addEventListener("beforeunload", () => {
+  canv.domElement.removeEventListener("click", handleCanvasClick);
+  canv.dispose();
+});
```
frasercl, line 224: I'm honestly not certain that I needed to add this event handler in the first place

src/colorizer/Dataset.ts
```diff
@@ -66,6 +80,18 @@ export default class Dataset {
...
+  private async loadTracks(): Promise<void> {
+    if (!this.tracksFile) {
+      return;
+    }
+    const url = this.resolveUrl(this.tracksFile);
+    const { trackIds, times } = await this.tracksLoader.load(url);
+    this.trackIds = trackIds;
+    this.times = times;
+    console.log(this.times);
+    console.log(this.trackIds);
```
frasercl, line 92: Other console.logs are useful to debug this incomplete feature, but these raw data ones probably shouldn't remain

frasercl (approved):
Other than my nebulous interface comments and elements you've already flagged for a bit of improvement, this looks like a very solid start. I experimented with the `DataSource` interface I described above on a child branch to this one, which we can have a look at after this merges, but I wouldn't be upset about the third loader remaining either.

PR #14, opened 2023-05-11
Feature/highlight selected

toloudis:
Problem
=======
When you click on a cell there is no viewport indication of what cell you clicked on.

Solution
========
I added outlining to the colorizing shader.

## Type of change

Change summary:
---------------
The outlining happens naively in the same shader.  For more robustness/flexbility it could be in a separate render pass depending on what we want to do.

I also added a "Track" class which keeps a list of all the times and id values of a given track.  This can then be used to generate plot data AND the id of the selected cell as it moves through time.  This is a refactor into a logical object that might be reusable in other areas as we continue to develop the app.

Oh I also fixed some type weirdness with respect to DataTexture vs Texture (tsc was complaining - we need to add gh actions that auto-run tsc)

Steps to Verify:
----------------
1. Click on cell, see red outline.  
2. advance through time, same cell should be outlined from frame to frame

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -5,6 +5,8 @@ import JsonArrayLoader from "./loaders/JsonArrayLoader";
 import ImageFrameLoader from "./loaders/ImageFrameLoader";
 
 import FrameCache from "./FrameCache";
+import Track from "./Track";
+
 import { FeatureDataType } from "./types";
 
 type DatasetManifest = {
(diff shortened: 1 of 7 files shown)


src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -60,6 +62,20 @@ void main() {
...
+  else if (int(id) - 1 == highlightedId) {
+    float thickness = 2.0;
+    float wStep = 1.0 / float(frameDims.x);
+    float hStep = 1.0 / float(frameDims.y);        
+    // sample around the pixel to see if we are on an edge
+    uint R = combineColor(texture(frame, sUv + vec2(thickness*wStep, 0)));
+    uint L = combineColor(texture(frame, sUv + vec2(-thickness*wStep, 0)));
+    uint T = combineColor(texture(frame, sUv + vec2(0, thickness*hStep)));
+    uint B = combineColor(texture(frame, sUv + vec2(0, -thickness*hStep)));
+    if (R == 0u || L == 0u || T == 0u || B == 0u) {
```
toloudis, line 74: do we need to do combinecolor if we are only testing for 0?

src/main.ts
```diff
@@ -198,7 +202,15 @@ async function drawLoop(): Promise<void> {
     await drawFrame(currentFrame);
     const delta = leftArrowDown ? -1 : 1;
     currentFrame = (currentFrame + delta + dataset.numberOfFrames) % dataset.numberOfFrames;
+    // update higlighted cell id if any
+    if (selectedTrack) {
+      const id = selectedTrack.getIdAtTime(currentFrame);
+      canv.setHighlightedId(id - 1);
```
toloudis, line 208: these -1 shifts make me uneasy.  Happens in shader, and when we read back on click.   Would be nice to isolate that somewhere.

src/colorizer/shaders/colorize_RGBA8U.frag
frasercl: you still need to check all the RGB components, though for testing for 0 you could skip the shifts that combineColor does and just say `r | g | b`.

src/main.ts
frasercl: Could start using the alpha channel as a background flag?

frasercl: Since the shifts only really become necessary when indexing into these big data arrays, I guess part of why we're (necessarily) creating extra confusion around this now is because tracks introduce new places outside the shader where we need to read feature data. I found subtracting 1 immediately when reading back on click to be a bit confusing - it leads to having to remember that the highlighted ID is already shifted back by 1 in the shader:
`if (int(id) - 1 == highlightedId) { ... }`
Both of the shifts you mention could be removed without affecting this comparison. It would require shifts to instead be made in getters for track ids and times, but that might be a reasonable way to, if not isolate shifts, at least standardize where they ought to happen: immediately before using a cell ID to index into a data array. (It's super possible that this would instead introduce more headache, I haven't fully investigated - e.g. it looks like a function like `getIdsOfTrack` would have to start *adding* 1 to its output to shift array indexes back to cell IDs.)
Seeing the -1 cell id used in plot titles also makes me wonder whether scientists would expect cell IDs to start at 1 or 0, or if they'll ever want to use the number currently in the plot title in the first place.

src/colorizer/Dataset.ts
```diff
@@ -170,44 +172,33 @@ export default class Dataset {
...
   }
 
-  // get the times and values of a track for a given feature
-  // this data is suitable to hand to d3 or plotly as two arrays of domain and range values
-  public buildTrack(trackId: number, feature: string): { domain: number[]; range: number[] } {
+  private getIdsOfTrack(trackId: number): number[] {
+    return this.trackIds?.reduce(function (arr: number[], elem: number, ind: number) {
+      if (elem === trackId) arr.push(ind);
+      return arr;
+    }, []) as number[];
```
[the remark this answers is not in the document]
toloudis: no good reason.. i'll make it arrow func.

src/colorizer/shaders/colorize_RGBA8U.frag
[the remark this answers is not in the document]
toloudis: ok sure.

src/main.ts
toloudis: Maybe not worth doing anything just yet but if this comes up more often we can do a single PR to resolve it all, referencing the logic you outlined here.

frasercl (approved):

Issue #4, opened 2023-04-27
display T value


toloudis:
## Use Case

show the actual time index as text

PR #15, opened 2023-05-26
add full time controls

### LINKED ISSUE #5: allow users to select t
## Use Case

provide a time slider and numeric input

toloudis:
Problem
=======
Resolves #5 and #4

Solution
========
Steal the controls from volume-viewer, encapsulate them in a little class, and hook them up to the current redraw function
The key handler now advances time a bit more slowly but never skips a frame.

## Type of change

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -35,18 +35,18 @@ class TimeControls {
     this.backBtn = document.querySelector("#backBtn")!;
     this.timeSlider = document.querySelector("#timeSlider")!;
     this.timeInput = document.querySelector("#timeValue")!;
-    this.playBtn.addEventListener("click", ()=>this.handlePlayButtonClick());
-    this.pauseBtn.addEventListener("click", ()=>this.handlePauseButtonClick());
-    this.forwardBtn.addEventListener("click", ()=>this.handleFrameAdvance(1));
+    this.playBtn.addEventListener("click", () => this.handlePlayButtonClick());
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -196,24 +313,19 @@ async function drawFrame(index: number): Promise<void> {
...
-    plot.setTime(currentFrame);
-
+    plot.setTime(timeControls.currentFrame);
+    await drawFrame(timeControls.currentFrame);
+  }
+  if (drawLoopRunning) {
     window.requestAnimationFrame(drawLoop);
-  } else {
-    drawLoopRunning = false;
   }
```
frasercl, line 329: Looks like with `TimeControls` now handling playback, `drawLoopRunning` is set to false and never changed, so it and this if statement can go away

src/main.ts
```diff
@@ -11,6 +11,139 @@ const datasetSelectEl: HTMLSelectElement = document.querySelector("#dataset")!;
...
+    if (totalFrames < 2) {
+      this.playBtn.disabled = true;
+    } else {
+      this.playBtn.disabled = false;
+    }
+    if (totalFrames < 2) {
+      this.pauseBtn.disabled = true;
+    } else {
+      this.pauseBtn.disabled = false;
+    }
```
frasercl, line 141: Both these if statements have the same condition and can be combined

src/main.ts
frasercl, line 1: This file has a couple minor spacing issues that prettier could clean up before merging

frasercl (approved):

Issue #16, opened 2023-06-01
find track


toloudis:
## Use Case

I would like to be able to enter a track id that I know about beforehand, and have the viewer display it. 

## Acceptance Criteria

Command button:  "Go to track" --> open dialog to get number entry? --> advance time to beginning of track, plot track and highlight cell

## Details

If you can't open a dialog then we should have a hide/show for the numeric input so it doesn't have to be kept up to date...?

PR #24, opened 2023-06-13
feat: Added UI for finding and seeking to a track by ID

ShrimpCryptid:
Problem
=======
Implements #16, finding and displaying by track ID.

Solution
========
Adds input field for a track ID and an action button. Clicking the button selects the track (if it exists) and seeks to its first frame.

## Type of change

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,8 +17,11 @@ class TimeControls {
   private pauseBtn: HTMLButtonElement;
   private forwardBtn: HTMLButtonElement;
   private backBtn: HTMLButtonElement;
+  private findTrackBtn: HTMLButtonElement;
   private timeSlider: HTMLInputElement;
   private timeInput: HTMLInputElement;
+  private trackInput: HTMLInputElement;
+
   private totalFrames: number;
   public currentFrame: number;
   private timerId: number;
@@ -33,12 +36,16 @@ class TimeControls {
     this.pauseBtn = document.querySelector("#pauseBtn")!;
     this.forwardBtn = document.querySelector("#forwardBtn")!;
     this.backBtn = document.querySelector("#backBtn")!;
+    this.findTrackBtn = document.querySelector("#findTrackBtn")!;
     this.timeSlider = document.querySelector("#timeSlider")!;
     this.timeInput = document.querySelector("#timeValue")!;
+    this.trackInput = document.querySelector("#trackValue")!;
     this.playBtn.addEventListener("click", () => this.handlePlayButtonClick());
     this.pauseBtn.addEventListener("click", () => this.handlePauseButtonClick());
     this.forwardBtn.addEventListener("click", () => this.handleFrameAdvance(1));
     this.backBtn.addEventListener("click", () => this.handleFrameAdvance(-1));
+    this.findTrackBtn.addEventListener("click", () => this.handleFindTrack());
+    this.trackInput.addEventListener("change", () => this.handleFindTrack());
     // only update when DONE sliding: change event
     this.timeSlider.addEventListener("change", () => this.handleTimeSliderChange());
     this.timeInput.addEventListener("change", () => this.handleTimeInputChange());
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -19,6 +21,7 @@ class TimeControls {
   private backBtn: HTMLButtonElement;
   private timeSlider: HTMLInputElement;
   private timeInput: HTMLInputElement;
+
   private totalFrames: number;
   public currentFrame: number;
```
toloudis: since you added the getter function, let's make `currentFrame` private

src/main.ts
```diff
@@ -295,6 +315,24 @@ function handleKeyDown({ key }: KeyboardEvent): void {
   }
 }
 
+async function handleFindTrack() {
+  // Load track value
+  let trackId = trackInput.valueAsNumber;
+  let newTrack = dataset!.buildTrack(trackId);
```
toloudis, line 321: Please always prefer `const` over `let` since it is more conservative against accidental modification.  (We probably should add this to our linting rules if not already)

src/main.ts
```diff
@@ -295,6 +315,24 @@ function handleKeyDown({ key }: KeyboardEvent): void {
...
+
+  // Check for track validity
+  if (newTrack.times.length > 0) {
```
toloudis, line 324: Track has a member function so you can call `newTrack.length()`, a bit less intrusive

src/main.ts
ShrimpCryptid: Fixed and left a note on the issue ticket about adding linting for const/let

src/main.ts
```diff
@@ -295,6 +321,24 @@ function handleKeyDown({ key }: KeyboardEvent): void {
   }
 }
 
+async function handleFindTrack(): Promise<void> {
+  // Load track value
+  const trackId = trackInput.valueAsNumber;
+  const newTrack = dataset!.buildTrack(trackId);
+
+  // Check for track validity
+  if (newTrack.length() > 0) {
```
frasercl, line 330: This is an arbitrary style preference, but since this is validation I might flip the condition and make it an early return, e.g.
```typescript
if (newTrack.length() < 1) {
  return;
}
```
otherwise if this function had multiple layers of validation, the code would start drifting to the right.

src/main.ts
```diff
@@ -135,6 +138,25 @@ class TimeControls {
...
     }
   }
+
+  /**
+   * Sets the current frame and updates this TimeControls UI to match.
+   * @returns true if the frame was set correctly (false if the frame is out of range).
+   */
+  public setCurrentFrame(frame: number): boolean {
+    if (this.goToFrame(frame)) {
+      this.currentFrame = frame;
```
toloudis, line 148: this.goToFrame does actually set this.currentFrame so that line is not needed.

toloudis: Interestingly, this function looks a lot like handleFrameAdvance (except it takes an absolute number instead of a delta).  Maybe could just reuse that function! handleFrameAdvance(newframe-currentFrame)

src/main.ts
```diff
@@ -295,6 +321,24 @@ function handleKeyDown({ key }: KeyboardEvent): void {
...
 
+async function handleFindTrack(): Promise<void> {
+  // Load track value
+  const trackId = trackInput.valueAsNumber;
+  const newTrack = dataset!.buildTrack(trackId);
+
+  // Check for track validity
+  if (newTrack.length() > 0) {
+    selectedTrack = newTrack;
+    timeControls.setCurrentFrame(selectedTrack.times[0]);
```
frasercl, line 332: Do we know if users will ever want to select a track and remain on the current frame, e.g. to see the position of several tracks at a time? Not a blocking concern for this PR, just curious

toloudis: good question, I'd say we would wait until they request it.  The main use case is : "someone told me to look at track X"

src/main.ts
frasercl: or `handleFrameAdvance` could just call this function but convert from relative to absolute numbers first?

ShrimpCryptid: Good catch, thank you both!

frasercl (approved):
Looks good!

src/main.ts
```diff
@@ -295,6 +317,24 @@ function handleKeyDown({ key }: KeyboardEvent): void {
...
+  const trackId = trackInput.valueAsNumber;
+  const newTrack = dataset!.buildTrack(trackId);
+
+  if (newTrack.length() < 1) {  // Check track validity
+    return;
+  }
+  selectedTrack = newTrack;
+  timeControls.setCurrentFrame(selectedTrack.times[0]);
+  plot.plot(selectedTrack, featureName, timeControls.getCurrentFrame());
+  await drawLoop();
```
toloudis, line 331: I'm not sure but drawloop is possibly redundant now as setCurrentFrame should cause a redraw too.

toloudis (approved):
This is looking great! Very clean solution.

src/main.ts
ShrimpCryptid: Fixed, thanks!

Issue #2, opened 2023-04-20
add indication of range of feature values


toloudis:
Users want to see the min and max feature value.  For prototype, we could probably just print the numbers to the left and right of the displayed colormap

ShrimpCryptid:
- Remove color ramp label
- Label should flank existing color ramp selector (left/right)

Issue #1, opened 2023-04-20
Add some diverging colormaps


toloudis:
Add at least one diverging scheme with white in the middle
See:
https://matplotlib.org/stable/tutorials/colors/colormaps.html
scroll down to the "diverging" section for examples.

ShrimpCryptid:
Add 4-6 diverging ramps

PR #26, opened 2023-06-14
feat: Added diverging color ramps, feature range labels

ShrimpCryptid:
Problem
=======
Implements range labels for features (#2) and diverging color ramps (#1).

Solution
========
- Adds 5 new diverging color ramps
- Adds two labels to the color ramp and supporting HTML changes. Both update with dataset and selected feature changes.

## Notes:
- The functions that handle dataset loading and feature changes have some redundant code. Should they be merged?

## Type of change

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -10,6 +10,9 @@ document.querySelector<HTMLDivElement>("#app")!.appendChild(canv.domElement);
 const datasetSelectEl: HTMLSelectElement = document.querySelector("#dataset")!;
 const featureSelectEl: HTMLSelectElement = document.querySelector("#feature")!;
 const colorRampSelectEl: HTMLSelectElement = document.querySelector("#color_ramp")!;
+const colorRampContainerEl: HTMLDivElement = document.querySelector("#color_ramp_container")!;
+const colorRampMinEl: HTMLLabelElement = document.querySelector("#color_ramp_min")!;
+const colorRampMaxEl: HTMLLabelElement = document.querySelector("#color_ramp_max")!;
 
 // time / playback controls
 class TimeControls {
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -185,12 +188,27 @@ const colorStops: HexColorString[][] = [
...
 
 function populateColorRampSelect(): void {
   colorRampSelectEl.innerHTML = "";
+  const width = 120;
+  const height = 25;
+  // Sets dimensions for color ramp container, as color ramp isn't inline (absolute/floating)
+  colorRampContainerEl.attributeStyleMap.set("width", `${width}px`);
+  colorRampContainerEl.attributeStyleMap.set("height", `${height}px`);
   colorRamps.forEach((ramp, idx) => {
     const rampCanvas = ramp.createGradientCanvas(120, 25);
```
toloudis, line 213: looks like this line should now use `width, height`

toloudis (approved):
one minor thing but otherwise all good!

src/main.ts
```diff
@@ -185,12 +188,26 @@ const colorStops: HexColorString[][] = [
...
 ];
 const colorRamps = colorStops.map((ramp) => new ColorRamp(ramp));
 const DEFAULT_RAMP = 4;
 
 function populateColorRampSelect(): void {
   colorRampSelectEl.innerHTML = "";
+  const width = 120, height = 25;
+  // Sets dimensions for color ramp container, as color ramp isn't inline (absolute/floating)
+  colorRampContainerEl.attributeStyleMap.set("width", `${width}px`);
+  colorRampContainerEl.attributeStyleMap.set("height", `${height}px`);
```
frasercl, line 210: This just sent me down a small rabbit hole about CSSOM - looks cool! ...but doesn't work in Firefox. \*sigh\* This is maybe more an argument to incorporate Babel into our toolchain than anything, but for now prefer `.style.width/height`.

src/main.ts
toloudis: Sorry, what I meant is `ramp.createGradientCanvas(120, 25)` --> `ramp.createGradientCanvas(width, height)` so that if the numbers ever change you only have to change them in one place

frasercl (approved):

PR #27, opened 2023-06-14
Feature/GitHub linter

ShrimpCryptid:
Problem
=======
What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

Solution
========

with @pairperson1

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,6 +30,7 @@
     "no-unused-vars": 0, // Rely on typescript-eslint rule above
     "@typescript-eslint/explicit-function-return-type": ["error", { "allowExpressions": true }],
     "@typescript-eslint/typedef": ["error", { "parameters": true }],
+    "prefer-const": 1,
     "radix": 1,
     "react/jsx-uses-vars": 1,
     "react/no-deprecated": 0,
--- a/.github/workflows/run-eslint.yml
+++ b/.github/workflows/run-eslint.yml
@@ -1,0 +1,14 @@
+name: run-eslint
+run-name: Checking PR for ${{ github.actor }}
+on: pull_request
+jobs:
+  eslint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14'
+      - run: npm install
+      - run: npm run lint .
+      - run: npm run typeCheck .
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,12 +8,14 @@
       "name": "nucmorph-colorizer",
       "version": "0.0.0",
       "dependencies": {
+        "@typescript-eslint/parser": "^5.59.11",
         "plotly.js-dist-min": "^2.22.0",
         "three": "^0.151.3"
       },
       "devDependencies": {
         "@types/plotly.js-dist-min": "^2.3.1",
         "@types/three": "^0.150.1",
+        "@typescript-eslint/eslint-plugin": "^5.59.11",
         "eslint": "^8.38.0",
         "gh-pages": "^5.0.0",
         "prettier": "^2.8.7",
@@ -378,7 +380,6 @@
       "version": "4.4.0",
       "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.4.0.tgz",
       "integrity": "sha512-1/sA4dwrzBAyeUoQ6oxahHKmrZvsnLCg4RfxW3ZFGGmQkSNQPFNLV9CUEFQP1x9EYXHTo5p6xdhZM1Ne9p/AfA==",
-      "dev": true,
       "dependencies": {
         "eslint-visitor-keys": "^3.3.0"
       },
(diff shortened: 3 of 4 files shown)


src/colorizer/ColorizeCanvas.ts
```diff
@@ -167,7 +167,7 @@ export default class ColorizeCanvas {
   }
 
   setFeature(name: string): void {
-    if (!this.dataset?.features.hasOwnProperty(name)) {
+    if (!this.dataset || !Object.prototype.hasOwnProperty.call(this.dataset.features, name)) {
```
toloudis, line 170: ouch, this got worse.  Is there anything more concise than `Object.prototype.hasOwnProperty.call`?  Should `dataset.features` be typed such that it has a name property?

src/colorizer/types.ts
```diff
@@ -32,7 +32,7 @@ export type FeatureArrayType = {
 
 type FeatureTypeSpec<T extends FeatureDataType> = {
   /** The constructor for a `TypedArray` of this numeric type */
-  arrayConstructor: { new (arr: number[]): FeatureArrayType[T] };
+  ArrayConstructor: { new (arr: number[]): FeatureArrayType[T] };
```
toloudis, line 35: why caps here?

.github/workflows/run-typecheck.yml
```diff
@@ -0,0 +1,13 @@
+name: run-typecheck
```
toloudis, line 1: I suggest putting these (run-typecheck and run-eslint) into the same workflow file (call it ci-checks or something) so that we have one line item in the Actions tab in github.  @frasercl @ShrimpCryptid do you have a strong preference for two workflow runs per check, or one?   (also I suggest running it on every push to a pull request, not sure if that is what this does)

ShrimpCryptid: Yes, it currently will re-run every time the branch is pushed to!

I don't mind merging the two actions back into one. I originally split them into two different actions so that `eslint` failing wouldn't stop `typeCheck` from running, but looking at documentation it seems like I can have both in the same job and run even if a previous step failed.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: This was flagged by eslint! From what I understand, it's because it's possible to make non-default-prototype Objects which may have undefined behavior for `hasOwnProperty`. (https://eslint.org/docs/latest/rules/no-prototype-builtins).

I could just undo this change and override the error since we know `dataset` is either null or the `Dataset` type, would that make sense?

src/colorizer/types.ts
ShrimpCryptid: This is an eslint rule (https://eslint.org/docs/latest/rules/new-cap#capisnew) that enforces capitalization on any constructor called with the new keyword. This array constructor is usually called like this:
```
const buffer = new spec.ArrayConstructor(data);
```

I can modify the eslint configuration so it doesn't flag properties (i.e., `new spec.arrayConstructor` wouldn't trigger an error), if that makes sense to you!

src/colorizer/ColorizeCanvas.ts
toloudis: I don't think it's worth it.  I'm ok with it the way it is. The real problem is not with the Dataset type, it's that you truly have to call hasOwnProperty.  I suppose you could do something like "Object.keys includes name" but I don't know if you'd run into the same problem.

src/colorizer/types.ts
toloudis: Thanks for the explanation!  Makes sense to me.  No need to change anything.

.github/workflows/run-typecheck.yml
toloudis: separate jobs, separate steps, or separate command lines in the same step are all ok by me.  But I think it's slightly better to have them both in one file just to keep the Actions log a little bit cleaner.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Looks like Object.keys().includes() works here!

toloudis (approved):
🚀

src/colorizer/types.ts
frasercl: Was unsure when I wrote this whether I ought to capitalize it like any other property of `FeatureTypeSpec` or like a `TypedArray` constructor. Glad to hear eslint has a definitive opinion on this.

.github/workflows/run-static-checks.yml
frasercl, line 1: Looks like [equivalent actions in other repos](https://github.com/allen-cell-animated/volume-viewer/blob/main/.github/workflows/ci.yml) separate lint and typecheck into entirely separate jobs that run in parallel?

...though if type checking will run regardless of lint success, this ought to be more efficient anyways since both steps can share the same `npm ci`.

frasercl (approved):
Like the style changes and the action. Not an expert in how actions work, but if anything doesn't work as expected it can be worked out as the action gets some use.

Issue #3, opened 2023-04-20
add an option to fix the extremes of the range as the min/max across colonies


toloudis:
For any feature, add some ui to allow the min and max of that feature to be locked in even when changing datasets.  
Example:
load dataset1, feature1 (F1).    F1.min and F1.max are determined and the colormap extremes are mapped to those values.
activate the "lock min/max" feature.
load dataset2.  if we are still viewing F1 (?), then do not update the min/max values and keep the colormap range clamped to that min and max.  values outside the range will be considered constant (clamped to the ends of the range) and will receive the colors at the ends of the range.

UI concerns:  how to enable/disable this?  when disabled, do things need to refresh or do we wait until some other user action?  do we want to keep the same min/max even if it's a different feature with wildly different values?  (changing feature would have to auto-disable this "lock min/max" thing).

toloudis:
check box only

PR #28, opened 2023-06-15
feat: Added checkbox for locking color map range

ShrimpCryptid:
Problem
=======
Implements #3, add an option to fix the extremes of the range as the min/max across datasets.

Solution
========
- Adds a checkbox ('Lock Color Map Range') that locks the current min/max range values of the currently selected feature.
- Unchecking the checkbox resets the labels to the range of the currently selected feature.
- When switching datasets, the currently selected feature will remain selected unless there's no equivalent in the new dataset.

## Type of change

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -17,7 +17,6 @@ import {
 
 import ColorRamp from "./ColorRamp";
 import Dataset from "./Dataset";
-import { FeatureData } from "./Dataset";
 import { FeatureDataType } from "./types";
 import { packDataTexture } from "./utils/texture_utils";
 import vertexShader from "./shaders/colorize.vert";
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -17,6 +17,7 @@ import {
 
 import ColorRamp from "./ColorRamp";
 import Dataset from "./Dataset";
+import { FeatureData } from "./Dataset";
 import { FeatureDataType } from "./types";
 import { packDataTexture } from "./utils/texture_utils";
 import vertexShader from "./shaders/colorize.vert";
@@ -166,16 +167,17 @@ export default class ColorizeCanvas {
     this.setUniform("highlightedId", id);
   }
 
-  setFeature(name: string): void {
+  getFeatureData(name: string): FeatureData | null {
     if (!this.dataset?.features.hasOwnProperty(name)) {
-      return;
+      return null;
     }
+    return this.dataset.features[name];
+  }
 
-    const { tex, min, max } = this.dataset.features[name];
-
+  setFeature(tex: Texture, min: number, max: number): void {
     this.setUniform("featureData", tex);
     this.setUniform("featureMin", min);
-    this.setUniform("featureMax", max / 1.5);
+    this.setUniform("featureMax", max);
   }
 
   async setFrame(index: number): Promise<void> {
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -178,7 +178,7 @@ export default class ColorizeCanvas {
     if(!this.dataset?.hasFeature(name)) {
       return;
     }
-    const featureData = this.dataset?.getFeatureData(name)!;
+    const featureData = this.dataset.getFeatureData(name)!;
     this.featureName = name;
     this.setUniform("featureData", featureData.tex);
     // Don't update the range values when locked
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -192,7 +192,7 @@ export default class ColorizeCanvas {
     this.setUniform("featureMax", this.colorMapRangeMax);
   }
 
-  setColorMapRangeLock(locked: boolean) {
+  setColorMapRangeLock(locked: boolean): void {
     this.isColorMapRangeLocked = locked;
     if (this.featureName) {  // trigger update for color map range
       this.setFeature(this.featureName);
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -175,10 +175,10 @@ export default class ColorizeCanvas {
   }
 
   setFeature(name: string): void {
-    const featureData = this.dataset?.getFeatureData(name);
-    if (!featureData) {
+    if(!this.dataset?.hasFeature(name)) {
       return;
     }
+    const featureData = this.dataset?.getFeatureData(name)!;
     this.featureName = name;
     this.setUniform("featureData", featureData.tex);
     // Don't update the range values when locked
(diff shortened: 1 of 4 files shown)


src/colorizer/ColorizeCanvas.ts
```diff
@@ -166,16 +167,17 @@ export default class ColorizeCanvas {
...
+  }
 
-    const { tex, min, max } = this.dataset.features[name];
-
+  setFeature(tex: Texture, min: number, max: number): void {
     this.setUniform("featureData", tex);
     this.setUniform("featureMin", min);
-    this.setUniform("featureMax", max / 1.5);
+    this.setUniform("featureMax", max);
   }
```
frasercl, line 181: This API change, and the need to replace `canv.setFeature` with the new function `updateFeature` in main.ts, makes me think that locking behavior should be handled within `ColorizeCanvas`.

The goal of the `ColorizeCanvas` API should be to capture all the actions a user should reasonably want to perform on the canvas, so that main.ts (or any other future client of the class) can primarily focus on setting up UI and attaching it to the appropriate methods. I could see a benefit of splitting up feature switching into getting and setting for error handling purposes, but if implementing locking outside `ColorizeCanvas` requires main.ts to reach into the feature abstraction and manipulate underlying data, that's a sign it should be implemented inside `ColorizeCanvas`.

frasercl: Play/pause is a notable exception, but it gets away with it because it just calls `setFrame` in a smart way. (I'm honestly starting to reconsider keeping it separate though - if we wanted more explicit support, moving the `TimeControls` class into its own file might be a good first step)

toloudis: +1 for moving TimeControls into its own file.  I should have done it when I first added it.

toloudis: @frasercl it took me a while but I think I understand your main point here.  The lock-ability means we have to provide max and min independently of the feature data texture.  It does seem to make sense that ColorizeCanvas could be responsible for holding on to the locked min/max and the flag for whether to use the "locked" min/max or the Feature's own min/max.
What's the api then?
ColorizeCanvas.setMinMax(number, number) or setLockedMinMax(number, number)?  ColorizeCanvas.setLockMinMax(bool) ?

frasercl: I think `setLockMinMax` and `setMinMax` (`setFeatureRange`/`setLockFeatureRange`?) seem reasonable, though Peyton pointed out that getters for the min/max will be more important than setters at this stage. The UI needs access to the min/max values to display now, but `ColorizeCanvas` can set the min/max off of the feature until we build in the option to set the values manually later.

toloudis: Talked to Peyton and realized you don't need min/max setters, as the internals can just grab the values when lock is set to true.

ShrimpCryptid: @toloudis I agree with what @frasercl said after we had a chance to talk about it in person-- it makes sense that the logic in main.ts should be a UI wrapper around ColorizeCanvas, rather than storing and updating the range values itself.

I added getters for the min and max color map ranges and a setter for the lock (`setColorMapRangeLock(bool)`) into ColorizeCanvas, which are accessed from main.ts. I'll eventually need to add setters for the range values too, but that's part of a different feature that will come later.

src/main.ts
```diff
@@ -291,15 +296,30 @@ function handleDatasetChange({ currentTarget }: Event): void {
 function handleFeatureChange({ currentTarget }: Event): void {
   const value = (currentTarget as HTMLOptionElement).value;
   console.log(value);
-  canv.setFeature(value);
+  updateFeature(value);
+}
+
+function updateFeature(newFeatureName: string): void {
+  const featureData = dataset?.getFeatureData(newFeatureName);
```
toloudis, line 303: not meaning to complicate this, but would dataset.featureNames.includes(newFeatureName) work here?   It seems like we could even have a shortcut Dataset.hasFeature(name:string) to hide a bit of the details. It would be used in a couple of places...

src/main.ts
```diff
@@ -260,22 +261,26 @@ async function loadDataset(name: string): Promise<void> {
   timeControls.updateTotalFrames(dataset.numberOfFrames);
   timeControls.setCurrentFrame(0);
   resetTrackUI();
-  featureName = dataset.featureNames[0];
-  console.log("features: " + dataset.featureNames[0]);
+
+  // Only change the feature if there's no equivalent in the new dataset
+  if (!dataset.featureNames.includes(featureName)) {
```
toloudis, line 266: see my other comment about Dataset.hasFeature :)

toloudis (approved):
I made one suggestion to hide/encapsulate how we determine if a dataset has a feature of a given name, but otherwise this is looking good to me.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -175,18 +175,17 @@ export default class ColorizeCanvas {
...
 
   setFeature(name: string): void {
-    const featureData = this.dataset?.getFeatureData(name);
-    if (!featureData) {
+    if(!this.dataset?.hasFeature(name)) {
       return;
     }
+    const featureData = this.dataset?.getFeatureData(name)!;
```
ShrimpCryptid, line 181: Style question here! I changed this so it calls `hasFeature` and added the `!` symbol to assert that the feature data isn't null, but I wanted feedback on if this is a good change or not.

I like that it's slightly clearer to read (it's checking that the feature exists, then getting its data), but it might be possible for the dataset to have bad feature data? Is the original check better here?

src/colorizer/Dataset.ts
```diff
@@ -17,7 +17,7 @@ type DatasetManifest = {
   times?: string;
 };
 
-type FeatureData = {
+export type FeatureData = {
```
frasercl, line 20: This export is no longer used... though it's super possible that it will be later, so it might as well stay

src/colorizer/ColorizeCanvas.ts
frasercl: Yeah, I think this is good!

To potentially overexplain my philosophy here: the idea of explicitly typing `getFeatureData` to maybe return `null` is to have the type system force you to handle whatever errors are possible before using the value. In this context, `!` means "I checked, no errors are possible." Like Dan said I don't think "bad feature data" is a thing yet, or will ever be since we make all our own data, so checking if a feature with that key exists is all you need to do to make that assertion.

frasercl (approved):
Looking good!

src/colorizer/ColorizeCanvas.ts
frasercl: In fact, you can turn the `?.` after `dataset` into a regular old `.` here, since typescript recognizes that we wouldn't pass the if statement above if `dataset` were undefined, and narrows its type for the rest of the function

ShrimpCryptid: Okay, great, thank you! Marking this as resolved 😊

Issue #18, opened 2023-06-01
capture image sequence


toloudis:
## Use Case

I want to produce a colorized movie to embed in powerpoint

## Acceptance Criteria

"capture mode" button --> press Play or auto-play when capture is clicked --> disable UI (except for pause/cancel), iterate through sequence and save out pngs one by one --> exit capture mode when done or cancelled, re-enable UI.  Name output files sequentially.

## Details

* output resolution?  canvas size or could we do original frame size?
* where to put the controls? right next to time slider?
* capture track could be interesting but was not asked for.  do we start from t=0 or current t, start of track, etc?

PR #29, opened 2023-06-20
feat: Image sequence recording

ShrimpCryptid:
Problem
=======
Solves issue #18, which requests a way to capture sequences of images.

Solution
========
- Refactored tracking of current time to ColorizeCanvas instead of TimeControls.ts
- Moved TimeControls.ts to its own file
- Added new UI elements for starting and stopping recordings.
image
- **Recording**
  - Recordings have a default name ([dataset]-[feature]-[frame_number].png) that can be modified.
  - Once the recording is started, all other UI elements are disabled. Frames are advanced one-by-one, then downloaded to the default download directory.
  - When the recording is finished, the canvas will reset to the frame that the recording was started on.
  - Note: There is a series of download settings that should be disabled. By default, Chrome/Firefox/Edge do not ask you where to save downloads on Allen Institute machines, so it should be fine, but there are disastrous results if this isn't set up correctly. (e.g., having to click through 500+ save dialogs). In the future this could probably be done with a popup dialog, but seemed like overkill for this feature.

## Type of change

Extra Note:
I did a little experimenting on my machine to see what values would work well for the delay between each downloaded frame. A set time delay is not ideal, but I haven't been able to find a workaround.
| Delay (ms) | Record Time (m:ss) | % frames dropped |
| --- | --- | --- |
| 0 | 0:30 | 47% |
| 5 | 0:32 | 45% |
| 10 | 0:35 | 39% |
| 25 | 0:43 | 26% |
| 50 | 0:57 | 4.5% |
| 75 | 1:09 | 0.1% |
| 100 | 1:22 | 0% |


src/colorizer/RecordingControls.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    this.timerId = 0;
+    this.startingFrame = 0;
+    this.redrawfn = redrawfn;
+    this.isDisabled = false;
+
+    this.startBtn.addEventListener("click", () => this.handleStartButtonClick());
+    this.abortBtn.addEventListener("click", () => this.handleAbortButtonClick());
+    this.filePrefixInput.addEventListener("changed", () => this.handlePrefixChanged());
+    this.filePrefixResetBtn.addEventListener("click", () => this.handlePrefixResetClicked());
```
toloudis, line 46: Minor nitpick: I suggest grouping all the `queryselector` and `addeventlistener` lines together so that the ui elements are clearly logically distinct from the rest of the state.  Also that would match how TimeControls looks.

src/main.ts
```diff
@@ -381,23 +247,33 @@ function resetTrackUI(): void {
...
+      const id = selectedTrack.getIdAtTime(canv.getCurrentFrame());
       canv.setHighlightedId(id - 1);
-      // console.log(`selected track: ${selectedTrack.trackId}; highlighted id ${id}`);
     }
-    // update current time in plot
-    plot.setTime(timeControls.getCurrentFrame());
-    await drawFrame(timeControls.getCurrentFrame());
   }
+
+  await canv.render();
```
toloudis, line 259: this seems like the place to do the file saving if we are in a recording state.

toloudis: Also when thinking about dropped frames, I believe maybe we should be using requestAnimationFrame in this app to just run an infinite redraw loop.

toloudis: @ShrimpCryptid and I discussed a few ways to make this better and the result of that conversation is #30 but maybe not necessary to implement in this PR unless @frasercl thinks so...  We also agreed on an improvement that will make setFrame not have to return false.

src/colorizer/RecordingControls.ts
```diff
@@ -0,0 +1,147 @@
...
+      // Get canvas as an image URL that can be downloaded
+      const dataURL = this.canvas.domElement.toDataURL("image/png");
+      const imageURL = dataURL.replace(/^data:image\/png/, "data:application/octet-stream");
+      
+      // Update our anchor (link) element with the image data, then force
+      // a click to initiate the download.
+      this.hiddenAnchorEl.href = imageURL;
+      const frameNumber: string = currentFrame.toString().padStart(maxDigits, "0");
+      this.hiddenAnchorEl.download = `${this.filePrefixInput.value}${frameNumber}.png`;
+      this.hiddenAnchorEl.click();
```
toloudis, line 96: I wonder if creating a completely new anchor element each time would also have helped the dropped frames?

ShrimpCryptid: Tested this! It ran quite a bit slower than reusing the same anchor element (about 0:48) and dropped 18% of frames. This is comparable to the other runs-- I think the overhead of creating and adding some 570 anchor elements slows down the download loop and makes it comparable to the timeout.

src/colorizer/RecordingControls.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.isDisabled = false;
+
+    this.startBtn = document.querySelector("#sequence_start_btn")!;
+    this.abortBtn = document.querySelector("#sequence_abort_btn")!;
+    this.filePrefixResetBtn = document.querySelector("#sequence_prefix_reset_btn")!;
+    this.startAtCurrentFrameChkbx = document.querySelector("#sequence_start_frame_checkbox")!;
+    this.filePrefixInput = document.querySelector("#sequence_prefix")!;
+
+    this.hiddenAnchorEl = document.createElement("a");  // Hidden element for initiating download later
+    document.body.appendChild(this.hiddenAnchorEl);
```
toloudis, line 39: I believe that appending the `a` element to document.body is unnecessary...

src/colorizer/ColorizeCanvas.ts
```diff
@@ -207,7 +220,30 @@ export default class ColorizeCanvas {
     return this.colorMapRangeMax;
   }
 
+  getTotalFrames(): number {
+    return this.totalFrames;
+  }
```
frasercl, line 225: Can this use `this.dataset.numberOfFrames` rather than duplicating state in this object?

src/colorizer/TimeControls.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.backBtn = document.querySelector("#backBtn")!;
+    this.timeSlider = document.querySelector("#timeSlider")!;
+    this.timeInput = document.querySelector("#timeValue")!;
+    this.playBtn.addEventListener("click", () => this.handlePlayButtonClick());
+    this.pauseBtn.addEventListener("click", () => this.handlePauseButtonClick());
+    this.forwardBtn.addEventListener("click", () => this.handleFrameAdvance(1));
+    this.backBtn.addEventListener("click", () => this.handleFrameAdvance(-1));
+    // only update when DONE sliding: change event
+    this.timeSlider.addEventListener("change", () => this.handleTimeSliderChange());
+    this.timeInput.addEventListener("change", () => this.handleTimeInputChange());
```
frasercl, line 36: Worth thinking about generalizing this class and `RecordingControls` to just hold handlers and letting main.ts handle the arbitrary business of getting DOM controls and binding the handlers to them. Out of scope for this PR though.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Yup, done! Added some default behavior if the dataset hasn't been loaded yet.

src/colorizer/TimeControls.ts
toloudis: I like that idea, similar in spirit to #30 with the goal of decoupling UI from business logic.

src/colorizer/TimeControls.ts
```diff
@@ -0,0 +1,132 @@
...
+    // slider from also operating
+    const loadNextFrame = async (): Promise<void> => {
+      if (this.isDisabled) {
+        // stop render loop if time controls have been disabled.
+        return;
+      }
+      let nextFrame = this.canvas.getCurrentFrame() + 1;
+      if (nextFrame >= this.canvas.getTotalFrames()) {
+        nextFrame = 0;
+      }
```
toloudis, line 61: could now use
`let nextFrame = this.wrapFrame(this.canvas.getCurrentFrame() + 1);`

src/main.ts
```diff
@@ -299,18 +164,18 @@ function handleFeatureChange({ currentTarget }: Event): void {
   updateFeature(value);
 }
 
-function updateFeature(newFeatureName: string): void {
+async function updateFeature(newFeatureName: string): Promise<void> {
```
toloudis, line 167: I forget, isn't it the case if your function is marked `async` then the Promise on the return value is implied?

toloudis (approved):
I added a couple of tiny comments but LGTM overall!

src/main.ts
ShrimpCryptid: Doesn't seem like it! This is the error message that the typescript checker spits out if I do that:
`TS1064: The return type of an async function or method must be the global Promise<T> type. Did you mean to write 'Promise<void>'?`

src/colorizer/TimeControls.ts
```diff
@@ -0,0 +1,129 @@
...
+    clearInterval(this.timerId);
+
+    // TODO: Fix this function so that it doesn't stop the
+    // slider from also operating
+    const loadNextFrame = async (): Promise<void> => {
+      if (this.isDisabled) {
+        // stop render loop if time controls have been disabled.
+        return;
+      }
+      let nextFrame = this.wrapFrame(this.canvas.getCurrentFrame() + 1);
```
frasercl, line 58: Linter says this can be `const` now

frasercl (approved):
Looks good!

Issue #17, opened 2023-06-01
filter out feature values


toloudis:
## Use Case

I would like to eliminate certain feature values from the view

## Acceptance Criteria

select a max and min value for a feature - could be interesting to do this as a two ended slider tht you can push in from either end.  
I am assuming we would just hide cells that are outside the selected range.  Requested by @niveditasa 

## Details

depends on #2
Where to put the controls?   I imagine a slider just below the color gradient and lined up with it.  Do we want to truncate the gradient or rescale the gradient to fit to the new min and max?  If the latter, then lining it up with the slider would be misleading.

toloudis:
* make min and max be input controls
* add a new checkbox for "clamp min/max" vs "hide out of range values"

ShrimpCryptid:
- Add a button to reset to default min/max once edited

PR #31, opened 2023-06-21
feat: Editable ranges and masking for color map

ShrimpCryptid:
Problem
=======
Resolves issue #17, filtering out feature values.

Solution
========
- Adds editable inputs for the min and max fields of the color ramp display, which controls the endpoints of the gradient.
- When either input is modified, the lock range checkbox is checked, preventing the range endpoints from changing when the feature/dataset does.
- Unchecking the lock range checkbox resets the inputs to the default.
- A separate checkbox ("Hide values outside range") masks values (outliers included) that are outside the range.
## Type of change

image

### DIFF
--- a/src/colorizer/shaders/colorize_RGBA8U.frag
+++ b/src/colorizer/shaders/colorize_RGBA8U.frag
@@ -13,6 +13,8 @@ uniform vec3 outlierColor;
 
 uniform int highlightedId;
 
+uniform bool hideOutOfRange;  // bool
+
 in vec2 vUv;
 
 layout(location = 0) out vec4 gOutputColor;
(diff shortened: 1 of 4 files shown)


src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -86,12 +88,17 @@ void main() {
   // Data buffer starts at 0, non-background segmentation IDs start at 1
   float featureVal = getFeatureVal(int(id) - 1);
   uint outlierVal = getOutlierVal(int(id) - 1);
+  float normFeatureVal = (featureVal - featureMin) / (featureMax - featureMin);
 
-  if (isinf(featureVal) || outlierVal != 0u) {
-    // outlier
-    gOutputColor = vec4(outlierColor, 1.0);
+  // Mask all values, including outliers, that are outside the range.
+  if (hideOutOfRange && (normFeatureVal < 0.0 || normFeatureVal > 1.0)) {
```
toloudis, line 94: I wonder if "hideOutOfRange" should also hide outliers.  This was not specified and we can wait for user feedback on that.  But I imagine that outliers would also be considered "out of range" :)

toloudis (approved):
LGTM!

src/colorizer/shaders/colorize_RGBA8U.frag
ShrimpCryptid: I found that if I just left the outlier color shading as the default, I would get a lot of random outlier flicker, hence the masking that's applied here.

frasercl (approved):

Issue #23, opened 2023-06-13
track id in url query string


toloudis:
## Use Case

It could be useful to have the track ID in the url query string so that users can send particular track urls to each other.

## Acceptance Criteria

* When a cell (or track) is selected, update the url with `?dataset=X&track=Y`
* also update url when current dataset is changed
* Initialize app from query string when dataset and track are provided. 

## Details

Bonus: add `&t=time`

PR #38, opened 2023-07-14
feature/load metadata from URL

ShrimpCryptid:
Problem
=======
Resolves https://github.com/allen-cell-animated/nucmorph-colorizer/issues/23!

Solution
========
- The URL of nucmorph colorizer now updates with the selected dataset, feature, frame #, and track ID.
- On initial load of the webpage, these same parameters can be pulled from the URL and used to seek to the right dataset/feature/frame/track.
- Some formatting fixes on `ColorizeCanvas.ts` are also included.

Note: There's a possible bug where approximately 25% of the time when loading invalid track IDs, the canvas will freeze on the first render and not update when clicked (even though the rest of the interface will show selected/deselected tracks and the correct timestamp). The canvas resets correctly if the time slider is changed, however. I have not been able to reproduce this since discovering it yesterday.

Screenshots (optional):
-----------------------
image
Example URL: `http://localhost:5173/?dataset=mama_bear&feature=NUC_shape_volume_lcc&track=11018&t=346`

### DIFF
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -157,7 +157,7 @@ export default class ColorizeCanvas {
     } else {
       this.setUniform("outlierData", packDataTexture([0], FeatureDataType.U8));
     }
-  
+
     // Force load of frame data (clear cached frame data)
     const frame = await this.dataset?.loadFrame(this.currentFrame);
     if (!frame) {
@@ -189,7 +189,7 @@ export default class ColorizeCanvas {
   }
 
   setFeature(name: string): void {
-    if(!this.dataset?.hasFeature(name)) {
+    if (!this.dataset?.hasFeature(name)) {
       return;
     }
     const featureData = this.dataset.getFeatureData(name)!;
@@ -203,7 +203,7 @@ export default class ColorizeCanvas {
     }
     this.setUniform("featureMin", this.colorMapRangeMin);
     this.setUniform("featureMax", this.colorMapRangeMax);
-    this.render();  // re-render necessary because map range may have changed
+    this.render(); // re-render necessary because map range may have changed
   }
 
   setColorMapRangeLock(locked: boolean): void {
@@ -241,7 +241,7 @@ export default class ColorizeCanvas {
       this.setUniform("featureMax", this.colorMapRangeMax);
     }
   }
-  
+
   getColorMapRangeMin(): number {
     return this.colorMapRangeMin;
   }
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -25,6 +25,7 @@ import pickFragmentShader from "./shaders/cellId_RGBA8U.frag";
 
 const BACKGROUND_COLOR_DEFAULT = 0xf7f7f7;
 const OUTLIER_COLOR_DEFAULT = 0xc0c0c0;
+export const BACKGROUND_ID = -1;
 
 type ColorizeUniformTypes = {
   aspect: number;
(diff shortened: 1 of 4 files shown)


src/colorizer/ColorizeCanvas.ts
ShrimpCryptid, line 1: This file is entirely formatting changes. I thought I'd include it in this PR because it's pretty minimal.

src/main.ts
```diff
@@ -224,6 +226,7 @@ async function handleCanvasClick(event: MouseEvent): Promise<void> {
   if (id < 0) {
     selectedTrack = null;
     plot.removePlot();
+    selectedTrack = null; // clear selected track when clicking off of cells
```
toloudis, line 229: doing it twice?  or can you remove the line before removePlot?

ShrimpCryptid: Oops, good catch!

src/main.ts
```diff
@@ -256,22 +258,55 @@ function handleKeyDown({ key }: KeyboardEvent): void {
...
   plot.plot(selectedTrack, featureName, canv.getCurrentFrame());
   await drawLoop();
+  trackInput.value = "" + trackId;
 }
 
 function resetTrackUI(): void {
   trackInput.value = "";
 }
 
+// URL STATE /////////////////////////////////////////////////////////////
```
toloudis, line 282: Because main is getting so big, this whole section could go into a new .ts file.. and that file or class could encapsulate both (1) getting datasetParam, trackParam, featureParam, and timeParam from the initial url, and (2) updating the current url when they change (even t... it's a bit odd for the url to change that fast...?).

ShrimpCryptid: Actually there might be some better ways to do that, let me experiment and get back to you

toloudis: It's probably not really affecting performance. But this might be a good question to ask UX because simularium doesn't do it this way.  They only put t into a url when a user explicitly asks to generate a url for sharing via a command button.  The url is not expected to just hold all app state all the time.

toloudis: I suggest we leave it as is (the realtime updating) and see if it's bothering anyone or hurting playback perf.  Can fix it later

toloudis: unless you came up with a good way to handle it.  lol I got ahead of your comment about experimenting with it.

ShrimpCryptid: Ah, no super elegant solutions unfortunately. I ended up implementing this anyways because I found that there was a noticeable lag if the URL was updated with every timestamp. It now updates when the pause button is pressed or whenever another feature (dataset, feature, track) changes.

toloudis (approved):
LGTM!  I see a future where dataset,feature,track,time are React props!

src/colorizer/UrlUtility.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    if (feature) {
+      params.push(`${URL_PARAM_FEATURE}=${feature}`);
+    }
+    if (track) {
+      params.push(`${URL_PARAM_TRACK}=${track}`);
+    }
+    if (time) {
+      params.push(`${URL_PARAM_TIME}=${time}`);
+    }
```
frasercl, line 34: is it possible for `track` or `time` to be 0 and skip these ifs?

frasercl (approved):
One question to clear up, but otherwise looks good!

src/colorizer/UrlUtility.ts
toloudis: good catch. definitely true for time at least.

ShrimpCryptid: I think time = 0 isn't a huge deal because that's the default loaded frame. I want to say track values of 0 are also unused but I'll update this if check to allow them. Thanks!

ShrimpCryptid: Nope I stand corrected, track=0 values are used!

Issue #7, opened 2023-04-27
display track path as overlay


toloudis:
## Use Case

for a selected track, show the path from centroid to centroid up to the current time.  this could be an overlay that draws a simple line list.

toloudis:
We do a similar line list implementation in simularium-viewer which may be instructive as reference, although it maybe more complicated than what's needed here.

toloudis:
Spec: add a checkbox "Show track path in view" or something more brief than that.  It will hide/show the line layer.

Line list should be drawn connecting centroids in the selected track, in-order.  Only show path from beginning of track up to current time (therefore path should update incrementally).

Because of this, I suggest that a "Track" object also contains a copy of the array of centroids and bounds for that track. Can collect them up at track creation time. Then it will be very convenient to use them for drawing.

For now, just make the lines some constant color, like the same color as the outline.  

For reference implementation of drawing line lists, see simularium-viewer's Paths.  It's just using a basic threejs object for lines.  Simularium viewer is doing something where it tries to dim the "older" line segments but we don't need that here.  Just a const color, keep it simple till we need it to be more complicated.  

The lines can be drawn right on top of the main view.

PR #44, opened 2023-07-24
feat: Track Path

ShrimpCryptid:
Problem
=======
Resolves [#7, display track path as overlay](https://github.com/allen-cell-animated/nucmorph-colorizer/issues/7)!

Solution
========
- Added checkbox for showing track path to UI
- Paths are drawn using the centroid data that was merged into the datasets in a previous PR. It uses ThreeJS's built-in Line renderer, where each centroid is a vertex in a list of vertices that form the line.
  - The line material has a vertex shader added to ensure that the points scale correctly with the canvas.
  - The path is shown progressively with time playback (so you only see the path up to the current frame).
  - The vertex calculations are only done once when the cell is selected for the first time (yay performance!)
- Also fixed a bug where selecting a track with only one cell would not show the selection highlight (see **track 854** for a good example.)

Screenshots (optional):
-----------------------
image

### Example Video:

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/3c3ac7f7-8570-4bcf-8ee2-5725ec23189c


src/colorizer/Dataset.ts
```diff
@@ -205,8 +221,15 @@ export default class Dataset {
     // ids now contains all cell ids that have trackId.
     // get all the times for those cells, in the same order
     const times = ids.map((i) => (this.times ? this.times[i] : 0));
+    let centroids2d;
+
+    const centroids1d = this.centroids;
+    if (centroids1d) {
```
toloudis, line 227: why is it conditional? Shouldn't centroids be required now for every dataset?

src/colorizer/shaders/line.vert
```diff
@@ -0,0 +1,15 @@
...
+
+uniform highp usampler2D frame;
+uniform float aspect;
+
+void main() {
+    // Apply scale to account for frame scaling in the canvas
+    // (copied from `colorize_RGBA8U.frag`)
+    ivec2 frameDims = textureSize(frame, 0);
+    float frameAspect = float(frameDims.x) / float(frameDims.y);
+    vec2 scale = max(vec2(aspect / frameAspect, frameAspect / aspect), 1.0);
```
toloudis, line 11: what's the difference between aspect and frameAspect?

src/main.ts
```diff
@@ -267,12 +267,13 @@ async function handleFindTrack(): Promise<void> {
 async function findTrack(trackId: number): Promise<void> {
   const newTrack = dataset!.buildTrack(trackId);
 
-  if (newTrack.length() < 1) {
+  if (newTrack.times.length < 1) {
```
toloudis, line 270: why this change? I believe it is still safer and more maintainable to call `newTrack.length()`

src/colorizer/Track.ts
```diff
@@ -2,21 +2,26 @@ export default class Track {
...
+    // sort time, id, and centroids, ascending by time
     const shouldSort = true;
     if (shouldSort) {
       const indices = [...times.keys()];
       indices.sort((a, b) => (times[a] < times[b] ? -1 : times[a] === times[b] ? 0 : 1));
       this.times = indices.map((i) => times[i]);
       this.ids = indices.map((i) => ids[i]);
+      if (centroids) {
+        this.centroids = indices.map((i) => centroids[i]);
+      }
```
toloudis, line 24: here is how you do it while keeping a 1D array:
```
this.centroids = indices.reduce(function(result, i) {
  result.push(centroids[i*2], centroids[i*2+1]);
  return result;
}, []);

```

src/colorizer/ColorizeCanvas.ts
```diff
@@ -185,8 +252,66 @@ export default class ColorizeCanvas {
...
+      return;
+    }
+    // Make a new array of the centroid positions,
+    // normalizing to screen space.
+    // We need three floats per coordinate
+    this.points = new Float32Array(track.centroids.length * 3);
+    for (let i = 0; i < track.centroids.length; i++) {
+      // Screen space is in [-1, 1] range.
+      this.points[3 * i + 0] = (track.centroids[i][0] / (4.0 * this.frameResolution.x)) * 2.0 - 1.0;
+      this.points[3 * i + 1] = -((track.centroids[i][1] / (4.0 * this.frameResolution.y)) * 2.0 - 1.0);
```
toloudis, line 270: something about this math here seems to be fighting against what is in the vertex shader.  I believe you could just pass the values as centroids in their pixel coordinates and have the shader do the rest of the transformation from pixels to (-1,1) range by dividiing by frame textureSize

src/colorizer/Dataset.ts
ShrimpCryptid: My thinking was to maintain backwards-compatibility with any imported datasets that don't include centroid data for scientists working on other projects. This was useful for some of my tests too (like in the URL dataset PR) as I didn't need to create dummy `centroid.json` files for every dataset.

src/colorizer/ColorizeCanvas.ts
toloudis: alternatively, if you transform the coordinates to the -1..1 space HERE, then you shouldn't need to transform the positions at all in the vertex shader?

src/main.ts
ShrimpCryptid: Changed, thank you!

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: So, the canvas has coordinates in [-1, 1], but the way the image is resized in the canvas element makes the vertex coordinates not map correctly. The logic in the vertex shader scales it in the same way that the canvas is scaled.

image
image

I agree that passing pixel coords in as the vertex coordinates makes more sense, though, and I'll update the logic accordingly!

src/colorizer/shaders/line.vert
frasercl: `aspect` is the aspect of the dataset, `frameAspect` is the aspect of the canvas. These lines are copied over from the colorize shader and are for fitting the image into the canvas with the correct aspect ratio. Currently the canvas has the same aspect as all our datasets, but if it didn't this would shrink to fit.

It's possible that in both cases this bit of vector math could be done outside the shader?

toloudis: I see. Until today I didn't realize the canvas could ever be a different aspect ratio than the actual image data. Couldn't we just resize the canvas to always keep that in sync?

toloudis: anyway, knowing this, I would say the vertex shader is totally fine.  I'd save the change I mention above for a separate PR.

src/colorizer/ColorizeCanvas.ts
toloudis: Yup I was totally missing the resizing aspect ratio behavior.  But yeah I still agree with just passing the coordinates in as pixels.

src/colorizer/shaders/line.vert
frasercl: We could, but I figured the problem of keeping a canvas at the right aspect ratio in the context of a webpage (where it may have to share space and play nice with other elements) would be tougher than keeping a texture at the right aspect ratio within a canvas. Peyton and I were talking about just moving this vector math out of the shader, as I suggested above. I think either one would free up the path to use Three's native `LineBasicMaterial` rather than messing with custom shaders, which feels like a worthwhile thing to do in this PR if possible.

src/colorizer/ColorizeCanvas.ts
frasercl: could something like this be accomplished by setting the line's `scale`?

toloudis: yep that's another possibility - if you use the threejs `scale` then you are integrating more closely with threejs infrastructure and then it would be natural to use the modelview transform in the vertex shader.

src/colorizer/shaders/line.vert
ShrimpCryptid: Done! I made the calculations for this a bit more verbose but with more clarity. I also added Cameron's suggestion of doing the calculation outside the shader and passing it in.

src/colorizer/Dataset.ts
ShrimpCryptid: Ended up also updating this PR to include loading bounds data into the dataset + track

src/colorizer/ColorizeCanvas.ts
```diff
@@ -185,8 +287,65 @@ export default class ColorizeCanvas {
...
+    if (!track || !track.centroids || track.centroids.length === 0 || !this.frameResolution) {
+      return;
+    }
+    // Make a new array of the centroid positions, normalizing to screen space.
+    // Points is in 3D while centroids are pairs of 2D coordinates in a 1D array
+    this.points = new Float32Array(track.length() * 3);
+    for (let i = 0; i < track.length(); i++) {
+      // Screen space is in [-1, 1] range.
+      this.points[3 * i + 0] = (track.centroids[2 * i] / (4.0 * this.frameResolution.x)) * 2.0 - 1.0;
+      this.points[3 * i + 1] = -((track.centroids[2 * i + 1] / (4.0 * this.frameResolution.y)) * 2.0 - 1.0);
```
toloudis, line 304: the only remaining thing I would change is to do this part of the math `/ (4.0 * this.frameResolution.x)) * 2.0 - 1.0` inside the shader instead of here.  
At this place in the code, you should just copy the values from track.centroids into this.points without having to modify the values.   
This is not a deal breaker but what it accomplishes is to take this bit of math that's applied to every value in this linear loop and instead run it on the gpu where it is parallelized over batches of points.  It also gives you the opportunity to optimize copying the raw values.

src/main.ts
```diff
@@ -371,8 +369,9 @@ async function start(): Promise<void> {
   trackInput.addEventListener("change", () => handleFindTrack());
   colorRampMinEl.addEventListener("change", () => handleColorRampMinChanged());
   colorRampMaxEl.addEventListener("change", () => handleColorRampMaxChanged());
-  lockRangeCheckbox.addEventListener("change", () => handleLockRangeCheckboxChange());
-  hideOutOfRangeCheckbox.addEventListener("change", () => handleHideOutOfRangeCheckboxChange());
+  showTrackPathCheckbox.addEventListener("change", () => handleShowTrackPathChanged());
+  lockRangeCheckbox.addEventListener("change", () => handleLockRangeCheckboxChanged());
+  hideOutOfRangeCheckbox.addEventListener("change", () => handleHideOutOfRangeCheckboxChanged());
```
ShrimpCryptid, line 374: Renamed these (Change vs. Changed)

toloudis (approved):
reminding us to prioritize #51 very quickly after this...

src/colorizer/ColorizeCanvas.ts
```diff
@@ -164,7 +258,11 @@ export default class ColorizeCanvas {
     if (!frame) {
       return;
     }
+    // Save frame resolution for later calculation
+    this.frameResolution = new Vector2(frame.image.width, frame.image.height);
```
frasercl, line 262: This feels like it should be the dataset's responsibility to store, not the canvas's... but there's not a satisfying way to derive it since we can't guarantee that all images in a dataset are the same size (though they absolutely should be). So I suppose it's better here.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -185,8 +287,65 @@ export default class ColorizeCanvas {
...
+    // Get highlighted id
+    if (this.track) {
+      let id;
+      // Tracks of length 1 should not be offset by 1
+      if (this.track.length() === 1) {
+        id = this.track.getIdAtTime(this.currentFrame);
+      } else {
+        id = this.track.getIdAtTime(this.currentFrame) - 1;
+      }
+      this.setUniform("highlightedId", id);
```
frasercl, line 347: Might prefer:

1. `if (!this.track) { return; }` over `if (this.track) { ...continue }`
2. Two `setUniform` calls in both "if" branches taking a const value rather than one call and a declared undefined value

...but these are minor style things.

src/colorizer/Dataset.ts
```diff
@@ -120,6 +131,24 @@ export default class Dataset {
...
+  }
+
+  private async loadBounds(): Promise<void> {
+    if (!this.boundsFile) {
+      return;
+    }
+    const url = this.resolveUrl(this.boundsFile);
+    const source = await this.arrayLoader.load(url);
+    this.bounds = source.getBuffer(FeatureDataType.U16);
+  }
```
frasercl, line 150: Out of scope here, but might be about time to work out some generic way to handle these load functions rather than this boilerplate.

src/colorizer/Track.ts
frasercl: May also possible to use `flatMap` here?

frasercl (approved):
None of my comments are necessarily blockers. Excited to get this in!

src/colorizer/ColorizeCanvas.ts
toloudis: I would be comfy with building in the assumption that all frames of a dataset are same xy size.   Can't imagine why anyone would want otherwise...?   We could even put the values in our top level json file - we were also discussing adding a downsampling factor to resolve any possible differences between coordinates of centroid data vs size of images.

src/colorizer/Dataset.ts
toloudis: Nice idea, it could at least turn into a one line `this.bounds = await this.loadToBuffer(this.boundsFile, FeatureDataType.U16);`

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Agreed that the dataset should hold this info! I'd like for centroid coordinates to match that of the image sizes, so that transformation would only need to be handled by the data generation script(s). Having this data in the manifest would be nice because then we don't have to wait for the first frame to be loaded to get its resolution... but I could also just grab the texture size for now.

src/colorizer/Track.ts
toloudis: omg yes.  flatMap is WAY better isn't it.  Thank you.

src/colorizer/Dataset.ts
ShrimpCryptid: Fixed! It's not the perfect one-liner we were hoping for but a lot of the redundant code has been removed.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -114,14 +154,38 @@ export default class ColorizeCanvas {
...
     this.pickScene.add(this.pickMesh);
 
+    // Configure lines
+    this.points = new Float32Array([0, 0, 0]);
+
+    this.lineGeometry = new BufferGeometry();
+    this.lineGeometry.setAttribute("position", new BufferAttribute(this.points, 3));
+    // Line material has its own vertex + fragment shader because they're
+    // required to make sure they resize correctly with the frame. This is because
+    // the points are drawn in a relative canvas range [-1, 1] rather than the frame.
```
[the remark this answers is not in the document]
ShrimpCryptid: Yes! Thank you for the suggestion :D

src/colorizer/ColorizeCanvas.ts
```diff
@@ -114,14 +133,33 @@ export default class ColorizeCanvas {
...
+    this.lineMaterial = new LineBasicMaterial({
```
toloudis, line 144: I think the comment above here is basically obsolete

ShrimpCryptid: oops good catch

src/colorizer/Dataset.ts
```diff
@@ -167,13 +202,25 @@ export default class Dataset {
...
+    const loadBounds = this.loadToBuffer(FeatureDataType.U16, this.boundsFile);
+
+    promises.concat([loadOutlier, loadTracks, loadTimes, loadCentroids, loadBounds]);
+    await Promise.allSettled(promises);
+
+    this.outliers = await loadOutlier;
+    this.trackIds = await loadTracks;
+    this.times = await loadTimes;
+    this.centroids = await loadCentroids;
+    this.bounds = await loadBounds;
```
frasercl, line 223: This is much more elegant than the original! But... it looks a bit odd to have `await`s that we don't expect or want to wait for anything? (Wish I knew more about JS's async/await model to know that these definitely don't yield control, even if all these promises are definitely settled by this point. But I'm sure JS implementors in their wisdom thought of that.)

I might have tried to destructure these values out of the array returned by `Promise.allSettled`? I actually tried this to confirm doing this would typecheck properly, and I have the option to make a suggestion, so I'm curious to see what that does:

```suggestion
    const featuresPromise = Promise.allSettled(this.featureNames.map(this.loadFeature.bind(this)));

    const loadOutlier = this.loadToTexture(FeatureDataType.U8, this.outlierFile);
    const loadTracks = this.loadToBuffer(FeatureDataType.U32, this.tracksFile);
    const loadTimes = this.loadToBuffer(FeatureDataType.U32, this.timesFile);
    const loadCentroids = this.loadToBuffer(FeatureDataType.U16, this.centroidsFile);
    const loadBounds = this.loadToBuffer(FeatureDataType.U16, this.boundsFile);
    const othersPromise = Promise.all([loadOutlier, loadTracks, loadTimes, loadCentroids, loadBounds]);

    const result = await Promise.all([othersPromise, featuresPromise]);
    const [[outliers, tracks, times, centroids, bounds]] = result;

    this.outliers = outliers;
    this.trackIds = tracks;
    this.times = times;
    this.centroids = centroids;
    this.bounds = bounds;
```

I think leaving these `await`s would be just fine to be clear, but something to consider.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -73,13 +82,23 @@ export default class ColorizeCanvas {
   private mesh: Mesh;
   private pickMesh: Mesh;
 
+  // Rendered track line that shows the trajectory of a cell.
+  private lineGeometry: BufferGeometry;
+  private lineMaterial: LineBasicMaterial;
+  private line: Line;
```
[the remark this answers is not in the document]
ShrimpCryptid: Fixed, thank you!

src/colorizer/Dataset.ts
```diff
@@ -203,18 +250,30 @@ export default class Dataset {
...
-    const times = ids.map((i) => (this.times ? this.times[i] : 0));
+    const times = this.times ? ids.map((i) => (this.times ? this.times[i] : 0)) : [];
+
+    let centroids: number[] = [];
+    if (this.centroids) {
+      centroids = ids.reduce((result, i) => {
+        result.push(this.centroids![2 * i], this.centroids![2 * i + 1]);
+        return result;
+      }, [] as number[]);
+    }
```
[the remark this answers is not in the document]
ShrimpCryptid: The number of centroids usually isn't huge so I'm not *super* worried about performance, but I think flatMap makes this a lot more readable. I'll implement this in the places where it's used.

src/colorizer/Dataset.ts
ShrimpCryptid: Dan and I were both looking at the await statements sideways so I like this change, thank you!

Issue #22, opened 2023-06-12
Show currently hovered cell feature value


ShrimpCryptid:
## Use Case

Even though the cell is colorized, as a user, I might want to know the actual feature value.

## Acceptance Criteria
- Show the current feature value of the hovered cell in a box off the main display. If no cell is currently hovered over, the box should be empty.

## Details

- Decide where to show output.

PR #45, opened 2023-07-24
feat: Show hovered feature values

ShrimpCryptid:
Problem
=======
Resolves [#22, Show currently hovered cell value](https://github.com/allen-cell-animated/nucmorph-colorizer/issues/22)!

Solution
========
- Adds UI text that shows the currently hovered track and feature values.
- Adds a mouse movement listener that checks the current XY of the mouse when over the canvas.
  - The feature and track boxes show the value of the last hovered cell, ignoring non-cell pixels (background).
  - Upon exiting the canvas, the track and feature value box resets.

Screenshots (optional):
-----------------------
image

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/1b7ffade-8426-4b5b-8b21-a6d7663e537a

Questions:
-----------------------
- Should there be a responsive element to hovering? (should cells be highlighted? Is this too many redraws?)
- Should I remove the track information, or reorder it?

### DIFF
--- a/index.html
+++ b/index.html
@@ -78,18 +78,26 @@ p {
       </div>
     </div>
     <div id="app"></div>
-    <div>Time (use arrow keys)
-      <p style="margin:2px;">
-        <button id="playBtn">Play</button>
-        <button id="pauseBtn">Pause</button>
-        <button id="backBtn">Back</button>
-        <button id="forwardBtn">Forward</button>
-        <input id="timeSlider" type="range" min="0" max="0" step="1" value="0"/><input type="number" id="timeValue" min="0" max="0" value="0"/>
-      </p>
-    </div>
-    <div>Find by track:
-      <input id="trackValue" type="number" value=""/>
-      <button id="findTrackBtn">Find</button>
+    <div style="display: flex; flex-direction: row; gap: 10px;">
+      <div>
+        <div>Time (use arrow keys)
+          <p style="margin:2px;">
+            <button id="playBtn">Play</button>
+            <button id="pauseBtn">Pause</button>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -78,18 +78,26 @@
...
+        </div>
+        <div>Find by track:
+          <input id="trackValue" type="number" value=""/>
+          <button id="findTrackBtn">Find</button>
+        </div>
+      </div>
+      <div>
+        <p id="mouse_track_id">Track ID:</p>
+        <p id="mouse_feature_value">Feature:</p>
+      </div>
```
ShrimpCryptid, line 100: This is the only real change in this section

src/main.ts
```diff
@@ -249,6 +250,35 @@ function handleColorRampClick({ target }: MouseEvent): void {
...
+}
+
+function onMouseMove(event: MouseEvent): void {
+  if (!dataset) {
+    return;
+  }
+  const id = canv.getIdAtPixel(event.offsetX, event.offsetY);
+  if (id === -1) {
+    // Ignore background pixels
+    return;
```
toloudis, line 268: does this return mean that you don't clear out the last feature value when you move from foreground to background?

ShrimpCryptid: Yeah, it keeps the last hovered feature unless the cursor exits the canvas (there's a video preview of it in the PR). I found that if I reset the hovered value on touching background pixels the hovered feature value would jump around a lot, especially given how small some of the cell targets are.

toloudis (approved):

frasercl (approved):

Issue #34, opened 2023-06-30
point to URL of any hosted dataset


toloudis:
## Use Case

As a data stakeholder, I have the ability to host my data on AWS and would like to use the tool to point to any cloud url.

## Acceptance Criteria

nucmorph colorizer has a url query parameter for http://viewer_url/?dataset=dataset_url in which dataset_url points to the top level dataset json file.  The remaining file paths referenced inside should be considered relative to this url.

## Details

Take url string, chop off `manifest.json` and use the remaining string as baseUrl...

We will have to figure out what to do with the dataset dropdown box.  That is TBD.

Ultimately we could have dataset=url&feature=featureName&track=trackId&t=time but this issue is only for dataset.

toloudis:
Acceptable inputs: 
dataset=url_prefix/filename.json (our code will split to extract the url_prefix so that file paths inside the dataset file)
dataset=url_prefix (our code will assume `manifest.json` by default)

toloudis:
if url_prefix does not start with `http` or `//` then assume it to be in our internal filesystem hardcoded location. (use the /allen... prefix)

toloudis:
based on #36, collection=url_prefix&dataset=dataset_name.

Scenarios:
1. nucmorph has mamabear, goldilocks, babybear... AND Derek adds a new dataset for EMT that is not associated with nucmorph at all.   But all the data can be on-prem in the same stock file server.
2. anyone can deploy any dataset to public cloud location. (future-proofing, possibly Jessica Yu)

if collection is omitted, then dataset must be a full url_prefix at least.
if collection is present, then it is a url_prefix to a location with a list of datasets. 

We don't have any tooling to create a collection yet.  Do it manually for the time being.  A json file with an array of string names `[{"path":"baby_bear", "name":"Baby Bear"}, ...]`

PR #46, opened 2023-07-26
Feature/url datasets

ShrimpCryptid:
# Please read! ⬇

Problem
=======
This resolves [#34, pointing to the URL of any hosted dataset](https://github.com/allen-cell-animated/nucmorph-colorizer/issues/34)!

Solution
========
- A new `collection` parameter can be added to the URL, which describes either a path to a directory containing a `collection.json` or a custom `.json` file. This provides names of and relative paths (or URLs) to one or more datasets.
  - The relative paths can either specify a directory containing a `manifest.json`, a path to a custom `.json` file with the same format. They can also be URLs (starting with `http(s)://` or `//`).
  - These dataset names are loaded and then shown in the UI datasets dropdown.
  - If left empty, the collection defaults to the existing nucmorph-colorizer data store.
  
Example `collection.json`:
```
[
    { "name": "Default", "path": "test1" },
    { "name": "Nested", "path": "test2/nested/" },
    { "name": "Custom Manifest", "path": "test3/custom-manifest.json" },
    { "name": "Custom URL", "path": "http://dev-aics-dtp-001.corp.alleninstitute.org/dan-data/colorizer/data/baby_bear"},
]
```
  
- The `dataset` parameter is now overloaded to accept direct URL paths to datasets.
  - If set to a URL, the `collection` parameter is ignored and no other datasets can be selected.
  - The URL can specify either the directory of a `manifest.json` or be a direct path to a custom `.json` file with the same format.
  - If left empty or if the dataset is not in the collection, the first dataset in the collection is loaded by default.
  
- When switching datasets, if the new dataset is missing data (could not be initialized, invalid path, etc.), the last known good dataset will be reloaded instead. (TODO: error popup?)
  
Other changes:
- Added a **timeout fetch** in response to a bug discovered during today's on-premises data outage. The interface would pause while datasets were being fetched, but because the request never resolved, no error message would be printed.
- The current frame number now persists between datasets.

## Type of change

Steps to Verify:
----------------
- Clone and run the repo with `npm run dev`, which should start the development server at `localhost:5173`.
- All of these URLs should load and show data:

| Collection | Dataset | URL | Note |
| --- | --- | --- | --- |
| None | None | http://localhost:5173 |
| None | Name | http://localhost:5173?dataset=Baby%20Bear |
| None | Base URL | http://localhost:5173?dataset=http%3A%2F%2Fdev-aics-dtp-001.corp.alleninstitute.org%2Fdan-data%2Fcolorizer%2Fdata%2Fbaby_bear |
| None | JSON URL | http://localhost:5173/?dataset=http%3A%2F%2Fdev-aics-dtp-001.corp.alleninstitute.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Ftest3%2Fcustom-manifest.json |
| Base URL |  None | http://localhost:5173/?collection=http%3A%2F%2Fdev-aics-dtp-001.corp.alleninstitute.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest |
| JSON URL | None | http://localhost:5173/?collection=http%3A%2F%2Fdev-aics-dtp-001.corp.alleninstitute.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Ftest.json  | Should say CUSTOM in dataset box. Also, try loading dataset 4 to see default behavior for bad datasets. |
| Base URL |  Name | http://localhost:5173/?collection=http%3A%2F%2Fdev-aics-dtp-001.corp.alleninstitute.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest&dataset=Test%202%20Nested | Should load Test 2 on startup |

Screenshots (optional):
-----------------------
image

Thanks for contributing!


src/colorizer/UrlUtility.ts
ShrimpCryptid, line 1: The diff on this got a bit messed up because I removed the UrlUtility class and just directly exported the functions used here.

ShrimpCryptid:
Suggestion from @jessicasyu: Add support so collections can also define URLs instead of just relative paths

src/main.ts
```diff
@@ -109,27 +118,55 @@ function setColorRampDisabled(disabled: boolean): void {
...
+    }
+
+    console.log(`Fetching dataset from path '${datasetPath}'`);
+    dataset = new Dataset(datasetPath);
+    await dataset.open();
+  } catch (e) {
+    console.error(e);
+    console.error(`Could not load dataset '${name}'.`);
+    console.warn(`Reloading dataset '${lastDataset}' instead.`);
+    console.timeEnd("loadDataset");
```
toloudis, line 163: do you need to call console.timeEnd just above line 159 also? (if the await succeeds)

src/main.ts
```diff
@@ -109,27 +118,55 @@ function setColorRampDisabled(disabled: boolean): void {
...
-  dataset = new Dataset(`${baseUrl}/${name}`);
-  await dataset.open();
+  datasetSelectEl.value = name;
+  try {
+    let datasetPath;
+    if (isUrl(name)) {
+      datasetPath = getDatasetPath(name);
+    } else {
+      datasetPath = getDatasetPathFromCollection(name, collection, collectionData);
+    }
```
toloudis, line 154: it looks like lines 149-154 should be the body of just one function call to a function exported from UrlUtility

src/colorizer/UrlUtility.ts
```diff
@@ -2,68 +2,328 @@ const URL_PARAM_TRACK = "track";
...
+ * @param timeoutMs timeout before the request should fail, in milliseconds. Defaults to `DEFAULT_FETCH_TIMEOUT_MS`.
+ * @param options additional
+ * @returns a Response promise, as returned by `fetch(url, options)`. The promise will reject if the timeout is exceeded.
+ */
+export function fetchWithTimeout(
+  url: string,
+  timeoutMs: number = DEFAULT_FETCH_TIMEOUT_MS,
+  options?: Object
+): Promise<Response> {
+  const controller = new AbortController();
```
toloudis, line 54: TIL that AbortController exists

src/colorizer/UrlUtility.ts
```diff
@@ -2,68 +2,328 @@ const URL_PARAM_TRACK = "track";
...
+  const signal = controller.signal;
+  setTimeout(() => controller.abort, timeoutMs);
```
toloudis, line 56: what happens after timeoutMs when the fetch has already completed?

src/main.ts
```diff
@@ -2,11 +2,19 @@ import { HexColorString } from "three";
...
+  CollectionEntry,
+  DEFAULT_COLLECTION_FILENAME,
+  DEFAULT_COLLECTION_PATH,
+  getCollectionData,
+  isUrl,
+  loadParamsFromUrl,
+  saveParamsToUrl,
+  getDefaultDatasetName,
+  getExpectedDatasetPath,
+} from "./colorizer/UrlUtility";
```
toloudis, line 15: Readability:  this would feel better to me if UrlUtility exported a small object that let you scope all these names at the point where you call them in this file, like `urlUtil.saveParamsToUrl` etc.  
If I'm the only one who thinks this then I will not fight for it :)

toloudis (approved):
my only (minor) concern is namespacing all the random exports coming out of UrlUtility for readability reasons.

src/main.ts
ShrimpCryptid: I had it as an object export before but felt like it was too cluttered 😭 I definitely see what you mean though. @frasercl any thoughts?

frasercl: I'm generally not a huge fan of creating objects purely for code organization but I agree this is overly long. I know TypeScript has a `namespace` keyword that I've been meaning to get to know better? It could make sense to use here, if not I think I'd slightly prefer a static class.

Also there's a "util" folder in src that this file could potentially go into?

ShrimpCryptid: Yeah, I was considering moving it to utils but wasn't sure if that would make the code diff messier. I'll go ahead and make both of these suggestions since I think you both have a sense of what changed.

toloudis: I don't mind namespace, exported object, or static class.  Maybe you could even do `import * as urlUtil`

frasercl: I like that last idea a fair bit actually

ShrimpCryptid: Ended up just doing a wildcard import like this: `import * as urlUtils from "./utils/url_utils";`! I think this is sufficiently tidy but it can be easily refactored down the line.

src/colorizer/UrlUtility.ts
ShrimpCryptid: So, the abort signal tells the fetch to reject its promise, but since the promise is already settled (resolved) it won't change its state. I think there's a test in RequestQueue over in volume-viewer that confirms that promise behavior.

I will, however, update this code so that it clears the timeout when the fetch completes, and I'll leave a comment about it.

frasercl (approved):
Really like breaking out all these well-documented utility functions, keeps things very clean.

src/main.ts
```diff
@@ -111,27 +109,50 @@ function setColorRampDisabled(disabled: boolean): void {
...
+  } catch (e) {
+    console.error(e);
+    console.error(`Could not load dataset '${name}'.`);
+    console.warn(`Reloading dataset '${lastDataset}' instead.`);
+    console.timeEnd("loadDataset");
+    if (hasLoadedOtherDataset) {
+      // Re-load last dataset instead of this one
+      await loadDataset(lastDataset);
+    }
+    return;
```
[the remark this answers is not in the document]
ShrimpCryptid: I wasn't sure if there would be any memory concerns with keeping the old dataset, but on a second look I think the frames are only loaded one by one! This change has been made if you want to look :)

frasercl (approved):
👍 looks good!

PR #54, opened 2023-08-10
fix: Hide track path when track completes

ShrimpCryptid:
Problem
=======
Request from @toloudis! Track paths are now hidden once the track is completed.

Solution
========
- Adjusted checks in `ColorizeCanvas.ts` for the range of the track that's visible.

### DIFF
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -283,23 +283,19 @@ export default class ColorizeCanvas {
    * frame.
    */
   updateTrackRange(): void {
-    // Do nothing if track doesn't exist or doesn't have centroid data
-    if (!this.track || !this.track.centroids) {
-      return;
-    }
-    if (!this.showTrackPath) {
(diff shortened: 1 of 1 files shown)


src/colorizer/ColorizeCanvas.ts
```diff
@@ -288,13 +288,20 @@ export default class ColorizeCanvas {
...
       return;
     }
+
+    // Show path up to current frame
     const trackFirstFrame = this.track.times[0];
-    // Clamp path to track length (don't draw non-existent vertices)
-    const range = Math.min(this.currentFrame - trackFirstFrame, this.track.length());
+    const range = this.currentFrame - trackFirstFrame;
+
+    if (range >= this.track.length()) {
```
toloudis, line 299: || range < 1

toloudis (approved):
minor suggestion but LGTM thank you!!

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Changes made :)

Issue #55, opened 2023-08-10
Highlight persists when deselecting cells


ShrimpCryptid:
## Description

When clicking on the background to deselect a cell, the highlight on the last selected cell persists. This goes away when the frame is changed, so likely the highlighted track is not being updated or cleared in `ColorizeCanvas`.

## Expected Behavior

Highlight around last selected cell should be gone.

ShrimpCryptid:
Fixed this as part of the React refactor (I thought I broke this and forgot there was an issue ticket oops)! Will tag this as part of that PR.

Issue #50, opened 2023-08-01
when saving image sequence, custom prefix is not used


toloudis:
## Description

Change prefix.
Click start to save image sequence.
See that custom prefix resets to default and is not applied to saved images.

## Expected Behavior

Custom file prefix should be used.

Issue #48, opened 2023-08-01
Convert project to React


ShrimpCryptid:
## Use Case
Preparation for future UI overhaul!

- Convert existing HTML to JSX for controls
- Convert global variables into React state, separate logic and UI

ShrimpCryptid:
First pass will be a simpler, direct conversion! Everything in `main.ts` is being moved into a single `App` function component.

List of changes to make in the second pass (WIP):
- [ ] Move ColorRamp into its own component

Issue #32, opened 2023-06-22
encapsulate the feature color ramp controls in a module


toloudis:
## Use Case

Just to clean up the code a bit, we can encapsulate all the controls around locking min/max, reset, and setting min/max into a separate file.  

## Acceptance Criteria

main.ts gets smaller and simpler and forwards calls into some kind of Feature or FeatureController class. 

## Details

Does this really help anything?

PR #63, opened 2023-08-24
feat: Convert project to React

ShrimpCryptid:
Problem
=======
Resolves Issue #48, converting Nucmorph-colorizer to use React!

(Also resolves https://github.com/allen-cell-animated/nucmorph-colorizer/issues/55 and resolves https://github.com/allen-cell-animated/nucmorph-colorizer/issues/50.)

Solution
========
This is an initial, somewhat naïve approach. Some logic still needs to be broken into its own components, specifically the `ColorRamp`, `TimeControls`, and `RecordingControls` modules. (See https://github.com/allen-cell-animated/nucmorph-colorizer/issues/32).

A lot of the changes here focused on compatibility and forcing ColorizeCanvas to play nicely with React's update cycle. Some state updating doesn't follow the React update paradigm, mostly because ColorizeCanvas is not a React component and has some async operations.

- Converts the existing code logic in `main.ts` to a new functional component, `App.tsx`.
- HTML is rendered using TSX/JSX, and updated using React's render cycle.
- Reorganized code for readability
- Changes to TimeControls, RecordingControls to avoid stale closures.

Note: There is a known issue where, on load, the plot domain will start at 0. This goes away if the time, feature, or dataset is updated.

## Type of change

Screenshots (optional):
-----------------------

Keyfiles:
-----------------------
1. App.tsx

TODO Tasks: (but out of scope of this PR)
----------------------
- [ ] Make ColorRamp a React component.
- [ ] Make ColorizeCanvas a React component.
- [ ] Make Plotting a React component.
- [ ] Make `Collection.ts` a separate class, and move related code out of `url_utils.ts`.
- [ ] Make `RecordingControls.ts` and `TimeControls.ts` into stateless utility functions or React components.


src/colorizer/Plotting.ts
```diff
@@ -50,6 +51,7 @@ export default class Plotting {
       return;
     }
     const plotinfo = this.dataset?.buildTrackFeaturePlot(track, feature);
+    console.log("Plot time: " + time);
```
toloudis, line 54: can remove?

src/App.tsx
```diff
@@ -0,0 +1,795 @@
...
+const CANVAS_PLACEHOLDER_ID = "canvasPlaceholder";
+const COLOR_RAMP_PLACEHOLDER_ID = "colorRamp";
+const PLOT_PLACEHOLDER_ID = "plot";
+const DEFAULT_COLOR_RAMP = 4;
+
+function App(): ReactElement {
+  // STATE INITIALIZATION /////////////////////////////////////////////////////////
+
+  const [plot, setPlot] = useState<Plotting | null>(null);
+  const canv = useMemo(() => {
```
toloudis, line 20: I know we talked about this but I am still trying to understand.  It seems this is more like a single-shot useEffect because it is creating the ColorizeCanvas, than a useMemo because it's not computing something that might ever need to be re-computed.  In this case maybe useEffect and useMemo have no real difference.

src/App.tsx
```diff
@@ -0,0 +1,795 @@
...
+  const drawLoop = useCallback(async (): Promise<void> => {
+    // Note: Selected track, frame number, etc. are not updated here.
+    // Those operations are async, and need to complete before a state update to be
+    // rendered correctly.
+    canv.setShowTrackPath(showTrackPath);
+    canv.setHideValuesOutOfRange(hideValuesOutOfRange);
+    canv.setColorRamp(colorRamp);
+    canv.setColorMapRangeMin(colorRampMin);
+    canv.setColorMapRangeMax(colorRampMax);
+    canv.setSelectedTrack(selectedTrack);
```
toloudis, line 97: When canv is a react component all this will be gone

src/App.tsx
```diff
@@ -0,0 +1,795 @@
...
+   */
+  const setSize = (): void => canv.setSize(Math.min(window.innerWidth, 730), Math.min(window.innerHeight, 500));
+
+  // INITIAL SETUP  ////////////////////////////////////////////////////////////////
+
+  // Memoized because the URL can update as data is loaded in
+  const initialUrlParams = useMemo(urlUtils.loadParamsFromUrl, []);
+
+  // Load database and collections data from the URL.
+  useMemo(async () => {
```
toloudis, line 168: should this have an explicit dependency on initialUrlParams?

src/colorizer/RecordingControls.ts
```diff
@@ -1,146 +1,112 @@
...
-  private startBtn: HTMLButtonElement;
-  private abortBtn: HTMLButtonElement;
   private recording: boolean;
-  private startAtCurrentFrameChkbx: HTMLInputElement;
 
-  private filePrefixInput: HTMLInputElement;
-  private filePrefixResetBtn: HTMLButtonElement;
-  private useDefaultPrefix: boolean;
-  private defaultPrefix: string;
   private hiddenAnchorEl: HTMLAnchorElement;
```
toloudis, line 7: All the HTMLElement member variables that used to be here seem to imply that this wanted to be its own parent component with all those buttons and inputs as children!

src/colorizer/TimeControls.ts
```diff
@@ -1,54 +1,35 @@
...
 const DEFAULT_TIMER_ID = -1;
 export default class TimeControls {
-  private playBtn: HTMLButtonElement;
-  private pauseBtn: HTMLButtonElement;
-  private forwardBtn: HTMLButtonElement;
-  private backBtn: HTMLButtonElement;
-  private timeSlider: HTMLInputElement;
-  private timeInput: HTMLInputElement;
-
+  // TODO: Change to be React state?
```
toloudis, line 8: Same comment I had for RecordingControls, it was almost structured to be its own component already!

toloudis (approved):
My only concern is allowing the canvas to only be redrawn within the React re-render cycle instead of being on its own loop, but I think for the performance requirements we have, this may be sufficient.  
Can't wait to see the next steps.

src/colorizer/RecordingControls.ts
ShrimpCryptid: Yeah, that's part of the planned changes I want to make down the line! I thought there was an issue for this specifically but this is the closest we have (https://github.com/allen-cell-animated/nucmorph-colorizer/issues/30). `RecordingControls` and `TimeControls` should definitely be their own components.

src/App.tsx
ShrimpCryptid: There's a good quick description of the difference between the hooks here: https://www.geeksforgeeks.org/when-to-use-usecallback-usememo-and-useeffect

Quick paraphrasing:
`useMemo` returns the memoized **value** of a function that is provided, while `useEffect` is a callback run **after all the components have been rendered** and **does not return a value.**

While both only recalculate the provided function when their dependencies change, they happen in different parts of the React lifecycle (immediately for `useMemo` vs. after the render for `useEffect`), and only one (`useMemo`) returns its results. 

Other things I could have used:
```
// This would have also been fine, but because I never set the canvas again it adds an unused variable.
const [canv, setCanv] = useState(new ColorizeCanvas());

// useEffect only makes sense if used with state, since it doesn't return any values.
// This would also lead to some weirdness with canv being undefined during the first render.
const [canv, setCanv] = useState<ColorizeCanvas>();
useEffect(() => {
  setCanv(new ColorizeCanvas());
}, []);
```

src/App.tsx
ShrimpCryptid: 🔥🔥🔥🔥🔥

src/colorizer/RecordingControls.ts
toloudis: ..as much as it makes sense to after the new UI...

src/App.tsx
toloudis: Thank you!  That does help and I forgot about needing the return value.

src/App.tsx
ShrimpCryptid: I think this is okay without the explicit dependency, since initialUrlParams will never change after the first call to `useMemo(urlUtils.loadParamsFromUrl)`, and because we only want the second `useMemo` that loads in the database/collections to run once and therefore it should have no dependencies.

I'm going to expand the comment here a bit to explain what it's doing more concretely, though.

src/App.tsx
```diff
@@ -0,0 +1,798 @@
...
+    let _collectionData;
+    // Load single dataset instead of collection
+    if (_datasetName && urlUtils.isUrl(_datasetName)) {
+      await replaceDataset(_datasetName);
+      setIsInitialDatasetLoaded(true);
+      return;
+    }
+
+    // Load collection data.
+    _collection = _collection || urlUtils.DEFAULT_COLLECTION_PATH;
```
meganrm, line 185: I think you have this as a TODO, but pulling out all the url handling and data initialization out of a render function would make this easier to follow. but I realize this is a first pass at conversion

ShrimpCryptid: You're right, that makes sense. There's a refactor I wanted to do related to the way collections are currently being handled that I think this change would slot into neatly. I'll add a TODO in the code itself about this!

toloudis: There's also something about factoring the bodies of many of the functions out into other modules that would also make them potentially more unit-testable outside of React?  Just a vague thought for future work.

src/App.tsx
```diff
@@ -0,0 +1,798 @@
...
+    setIsInitialDatasetLoaded(true);
+  }, []);
+
+  // Run only once the first dataset is loaded.
+  useEffect(() => {
+    if (!isInitialDatasetLoaded) {
+      return;
+    }
+    plot?.removePlot();
+    plot?.setDataset(dataset!);
```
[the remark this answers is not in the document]
ShrimpCryptid: Yes! `useEffect` callbacks are run in order, and `plot` is assigned in the first callback after the first render.

`plot?.removePlot();` is used to remove the axis labels and plot title that appear by default when plot is initialized:
image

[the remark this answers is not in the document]
meganrm: I guess I was wondering why you had the conditional, does typescript just not know that it exists?

meganrm (approved):
Overall looks good, and I think you have a good set of future TODOs already based on this

src/App.tsx
```diff
@@ -0,0 +1,800 @@
...
+  const disableTimeControlsUi = disableUi;
+
+  return (
+    <div>
+      {/** Top Control Bar */}
+      <div className={styles.canvasTopControlsContainer}>
+        <label htmlFor="dataset">Dataset</label>
+        <select
+          name="Dataset"
+          id="dataset"
```
frasercl, line 564: I'd prefer using `className` to `id` in React components. Converting current uses of `getElementById` to refs should eliminate the need for unique IDs.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -75,6 +75,7 @@ const getDefaultUniforms = (): ColorizeUniforms => {
   };
 };
 
+// TODO: Change into a component?
```
frasercl, line 78: Some rambling about React component conversion: I'd advocate for keeping classes within this directory relatively unchanged rather than allowing React to penetrate too deeply into the inner workings of the app, though I'm not super committed to this preference. We are at least required to draw a line between what gets handled in React's functional paradigm vs. what React hands off via effects to be resolved imperatively, and I think we should be deliberate about where we put that line. Since this class's main purpose is to own a `canvas`, an element which must be manipulated imperatively via its methods, I think the boundaries of this class are a pretty natural place for that boundary to go.

I'm less certain about `RecordingControls`, `TimeControls`, `ColorRamp`, etc., though I will say: we have the option to wrap these classes in components rather than converting them to components. This would provide all the same ergonomic benefits but might provide greater separation of concerns.

toloudis: For context on that comment line, I think there was one main reason to think about making this a component, and that was to drive redraws using props passed in from its parent. (Aka The reworking of the drawLoop stuff, for which there are other valid ways to handle it..).

src/App.tsx
```diff
@@ -0,0 +1,800 @@
...
+
+function App(): ReactElement {
+  // STATE INITIALIZATION /////////////////////////////////////////////////////////
+
+  const [plot, setPlot] = useState<Plotting | null>(null);
+  const canv = useMemo(() => {
+    const canv = new ColorizeCanvas();
+    canv.setColorRamp(colorRamps[DEFAULT_COLOR_RAMP]);
+    return canv;
+  }, []);
```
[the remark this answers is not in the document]
ShrimpCryptid: Talked through making a custom hook for this, thank you so much for the suggestion! I replaced the declaration of the canvas and time/recording controls.

src/App.tsx
```diff
@@ -0,0 +1,800 @@
...
+    timeControls?.setFrameCallback(setFrame);
+    recordingControls?.setFrameCallback(setFrame);
+  }, [setFrame]);
+
+  // Update the default recording controls prefix.
+  useMemo(() => {
+    if (useDefaultPrefix && datasetName && featureName) {
+      setImagePrefix(`${datasetName}-${featureName}-`);
+    }
+  }, [useDefaultPrefix, datasetName, featureName]);
```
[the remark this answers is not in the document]
ShrimpCryptid: Hmm, I think there isn't any particular reason to favor `useMemo` over `useEffect` here. It probably doesn't matter where in the render lifecycle this update happens...

Honestly this is probably a meaningless optimization and it might be fine to run this on every render without caching the results? (timeControls/recordingControls don't do any computation for this beyond assigning a field.)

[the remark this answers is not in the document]
frasercl: Worth testing, but I think `useEffect` would probably be best here (or no hook at all). I also definitely agree that long term `TimeControls` and `RecordingControls` ought to be available as React components and that `setFrame` ought to be passed down to them as a prop, though I'm unsure how those components ought to look under the hood (per my comment below).

src/App.tsx
```diff
@@ -0,0 +1,800 @@
...
+    colorRampMin,
+    colorRampMax,
+    timeControls?.isPlaying(), // updates URL when timeControls stops
+    updateUrl,
+  ]);
+
+  // Call drawLoop whenever its dependencies change.
+  useMemo(() => {
+    drawLoop();
+  }, [drawLoop]);
```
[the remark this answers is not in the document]
ShrimpCryptid: You're totally right that this works as a `useEffect`!

...but now I'm confused as to why this works as a `useEffect` 😂

I thought that React wouldn't trigger a repaint unless some state has changed, but none of the operations in drawLoop update state? I did find a [blog post about how sometimes React will run `useEffect` before the render](https://blog.thoughtspile.tech/2021/11/15/unintentional-layout-effect/), so for now I've changed this to a `useLayoutEffect` to guarantee that it runs pre-render.

src/App.tsx
```diff
@@ -0,0 +1,800 @@
...
+import RecordingControls from "./colorizer/RecordingControls";
+import TimeControls from "./colorizer/TimeControls";
+import * as urlUtils from "./colorizer/utils/url_utils";
+
+import styles from "./App.module.css";
+import { useDebounce } from "./colorizer/utils/react_utils";
+
+const CANVAS_PLACEHOLDER_ID = "canvasPlaceholder";
+const COLOR_RAMP_PLACEHOLDER_ID = "colorRamp";
+const PLOT_PLACEHOLDER_ID = "plot";
```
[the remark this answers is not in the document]
ShrimpCryptid: Fixed!

src/App.tsx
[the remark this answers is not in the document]
ShrimpCryptid: Removed the hook here for now.

src/App.tsx
ShrimpCryptid: Fixed!

src/colorizer/ColorRamp.ts
```diff
@@ -40,3 +41,51 @@ export default class ColorRamp {
     this.texture.dispose();
   }
 }
+
+// https://developers.arcgis.com/javascript/latest/visualization/symbols-color-ramps/esri-color-ramps/
+const colorStops: HexColorString[][] = [
+  // Esri color ramps - Red 5
+  ["#fee5d9", "#fcae91", "#fb6a4a", "#de2d26", "#a50f15"],
```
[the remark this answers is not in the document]
ShrimpCryptid: Fixed, thank you! I have it in a new file (`constants.ts`) but the name of it is definitely up for debate.

src/App.tsx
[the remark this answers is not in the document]
ShrimpCryptid: We talked through this! This works because ColorizeCanvas is directly manipulating a DOM element and bypasses React's virtual DOM.

src/App.tsx
```diff
@@ -0,0 +1,766 @@
...
+   * Update the canvas dimensions based on the current window size.
+   */
+  const setSize = (): void => canv.setSize(Math.min(window.innerWidth, 730), Math.min(window.innerHeight, 500));
+
+  // INITIAL SETUP  ////////////////////////////////////////////////////////////////
+
+  // Use a memoized value here and only retrieve parameters once, because the URL can be updated
+  // by state updates and lose information (like the track, feature, time, etc.) that isn't
+  // accessed in the first render.
+  const initialUrlParams = useMemo(urlUtils.loadParamsFromUrl, []);
```
frasercl, line 160: This actually also looks like a great place to use `useConstructor`

src/App.tsx
```diff
@@ -0,0 +1,766 @@
...
+    await findTrack(parseInt(findTrackInput, 10));
+  }, [findTrackInput, findTrack]);
+
+  // RECORDING CONTROLS ////////////////////////////////////////////////////
+
+  // Update the callback for TimeControls and RecordingControls if it changes.
+  // TODO: TimeControls and RecordingControls should be refactored into components
+  // and receive setFrame as props.
+  timeControls.setFrameCallback(setFrame);
+  recordingControls.setFrameCallback(setFrame);
```
frasercl, line 533: This will hopefully be a bit nicer when these are their own components

frasercl (approved):
Looking good! My remaining comments are about `useMemo`, which I think can be removed entirely, but this is now in a state that I would feel good about merging with or without that work.

src/App.tsx
```diff
@@ -537,10 +529,8 @@ function App(): ReactElement {
...
   // and receive setFrame as props.
-  useMemo(() => {
-    timeControls?.setFrameCallback(setFrame);
-    recordingControls?.setFrameCallback(setFrame);
-  }, [setFrame]);
+  timeControls.setFrameCallback(setFrame);
+  recordingControls.setFrameCallback(setFrame);
 
   // Update the default recording controls prefix.
   useMemo(() => {
```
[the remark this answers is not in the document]
ShrimpCryptid: Fixed! Thanks for looking through this!

Issue #81, opened 2023-08-30
Document data format


ShrimpCryptid:
## Use Case
Add documentation for the data format so users can add new datasets!

## Acceptance Criteria

- Markdown data format descriptor, either in README or linked from README
- Include example of existing dataset(?)
- Instructions for modifying existing data conversion scripts?

PR #86, opened 2023-08-31
doc: Documented data format

ShrimpCryptid:
Problem
=======
Resolves https://github.com/allen-cell-animated/nucmorph-colorizer/issues/81!

Solution
========
- Adds new data format documentation, with examples and images.
- Updates README with new links to the data format, and updates outdated information.

**You can preview the new file here: https://github.com/allen-cell-animated/nucmorph-colorizer/blob/doc/data-format/documentation/DATA_FORMAT.md**

Discussion Points
========
- Where should this file live in the project?
- Should I include the steps to update `generate_data` in the section for Getting Started? (I think I was pretty shaky on the list of changes that are needed, so it may be better to leave this out until we can do a pass over `generate_data` to make it more modification-friendly.) 
- Are the example dropdowns sufficiently visible?
- @toloudis is there a stable place we can link to as an example dataset?

## Type of change

Screenshots (optional):
-----------------------
image


DATA_FORMAT.md
```diff
@@ -0,0 +1,330 @@
...
+}
+```
+
+---
+
+</details>
+
+### 1. Tracks
+
+Every segmented object in each time step has a **track ID**, an integer identifier that's unique across all time steps. To recognize a single track across multiple frames, these track IDs must be grouped together with a single **track number**.
```
toloudis, line 101: Every segmented object belongs to a track but also has an id that is unique across all segmented objects in the entire time series.  These object ids are required to be sequential starting from 0.  For example: frame 1 may have objects 0, 1, 2.  Frame 2 has objects 3,4,5.  The tracking algorithm has determined that objects 0 and 3 are the same object, and belong to track 149.   Objects 1 and 4 are track 163.  And so on.  The track ids are not required to be sequential.

DATA_FORMAT.md
```diff
@@ -0,0 +1,330 @@
...
+
+A **track JSON file** consists of a JSON object with a `data` array, where for each track ID `i`, `data[i]` is the track number that track ID is part of.
```
toloudis, line 103: for each unique object id i, data[i] is the track number that object id is part of...

DATA_FORMAT.md
```diff
@@ -0,0 +1,330 @@
...
+}
+```
+
+---
+
+</details>
+
+### 2. Times
+
+The times JSON is similar to the tracks JSON. It also contains a `data` array that maps from track IDs to the frame number that they appear on.
```
toloudis, line 149: basically i think you need to subsitute "object id" instead of "track id".

toloudis, line 149: here too "track IDs" should really be "object IDs", I think

DATA_FORMAT.md
```diff
@@ -0,0 +1,330 @@
...
+    ],
+    "min": <min value for all features>,
+    "max": <max value for all features>
+}
+```
+
+### 5. Centroids
+
+The centroids file defines the center of each track ID in the dataset. For each index `i`, the coordinates are `(x: data[2i], y: data[2i + 1])`.
+Coordinates are defined in pixels in the frame, where the upper left corner of the frame is (0, 0).
```
toloudis, line 196: Even though the example should be clear, I might just also state that the centroids must follow the same format as a feature, except that the x y coordinates must occur sequentially so the list of values will be twice as long as for a feature.

DATA_FORMAT.md
```diff
@@ -0,0 +1,330 @@
...
+        <upper left x for track id 1>,
+        <upper left y for track id 1>,
+        ...
+    ]
+}
+```
+
+### 7. Outliers
+
+The outliers file stores marks whether a given track ID should be marked as an outlier using an array of booleans (`true`/`false`). Indices that are `true` indicate outlier values, and are given a unique color in Nucmorph-Colorizer.
```
toloudis, line 235:
```suggestion
The outliers file stores whether a given track ID should be marked as an outlier using an array of booleans (`true`/`false`). Indices that are `true` indicate outlier values, and are given a unique color in Nucmorph-Colorizer.
```

toloudis: I wonder if this should be considered optional data for some datasets.  It seemed silly to generate a file with all false's in it.  Also, this is likely to be sparse data and could possibly be stored as such.  We can wait to hear from users on that...

DATA_FORMAT.md
toloudis: Typically our data sources are spreadsheets with one row per unique segmented object, in which case the most convenient and natural way to assign object ids is simply by row number.

DATA_FORMAT.md
```diff
@@ -0,0 +1,330 @@
...
+
+---
+
+</details>
+
+## Collections
+
+Collections are defined by an optional JSON file and group one or more datasets for easy access. By default, they should be named `collection.json`.
+
+Collections are an array of JSON objects, each of which define the `name` (an **alias**) and the `path` of a dataset. This can either be a relative path on a file server, or a complete URL.
```
toloudis, line 283: I would say somewhere that the datasets in a collection will be presented in the app for easy switching/selection of dataset within collection.

toloudis:
Rather than linking to a stable dataset, I think we could just add a very trivial tiny fake dataset to the repo.  Something with like 3 time steps, just a couple of tracked objects, so the arrays are very short.  Could modify the generate_test_data script to generate it.

toloudis:
I think this document could be linked from README as you did, or live at the root level or inside the scripts subdirectory, or a new docs/ subdir but either place seems to make sense to me.

DATA_FORMAT.md
ShrimpCryptid: Got it, I'll replace all the instances of "track ID" with "object ID" to clarify. 

Are object ID's required to be sequential? I thought that was typical but not guaranteed.

toloudis: They have to be sequential because that's how we index into the feature arrays.  Otherwise we'd have to have some other kind of mapping.  It keeps things much simpler.  If users really needed to have more meaningful object id's we'd probably introduce a new json file that contained that list.  But for the purposes of our app, the object id is its row index.

ShrimpCryptid: I see, I think I was mixing up object IDs with tracks here.
So across all tracks, object ID's must be sequential, but a track is not guaranteed to have sequential/adjacent object IDs?

toloudis: right, if a track is the set of object id's across all frames that track the same REAL object, then it may appear to be a random sequence of unique object ids.   And no object id can appear in more than 1 track.

DATA_FORMAT.md
ShrimpCryptid: Fixed!

DATA_FORMAT.md
ShrimpCryptid: Thanks for catching this!

It actually turns out that quite a few of the files *are* optional, and I went back and marked them in the documentation. Centroids, bounds, and outliers are all optional. 
(Technically, so are tracks and times, but I think the app will stop working as expected without them.)

DATA_FORMAT.md
```diff
@@ -0,0 +1,372 @@
...
+---
+
+For example, if there were the following two tracks in some dataset, the track file might look something like this.
+
+| Track # | Object IDs |
+| ------- | ---------- |
+| 1       | 0, 1, 4    |
+| 2       | 2, 3, 5    |
+
+Note that the object IDs are not guaranteed to be sequential!
```
toloudis, line 132:
```suggestion
Note that the object IDs in a track are not guaranteed to be sequential!
```

DATA_FORMAT.md
```diff
@@ -0,0 +1,372 @@
...
+
+_Example frame:_
+![](./documentation/frame_0.png)
+_Each unique color in this frame is a different object ID._
+
+**Frames** are image textures that store the object IDs for each time step in the time series. Each pixel in the image can encode a single object ID in its RGB value (`object ID = R + G*256 + B*256*256 - 1`), and background pixels are `#000000` (black).
+
+Additional notes:
+
+- There is currently no way to indicate overlaps in the data-- your script will need to flatten 3D data and segmentations.
```
toloudis, line 178: Nit: I would consider removing this line as it is already captured in an issue ticket and might confuse - as in "what overlaps?"

DATA_FORMAT.md
```diff
@@ -0,0 +1,372 @@
...
+- Encoded object ID's in the frame data start at `1` instead of `0`, because `#000000` (black) is reserved for the background.
+- The highest object ID that can currently be represented is `16,843,007`.
```
toloudis, line 180: This is an interesting limitation to note, and I might also stress that we can remove that limit if the _total number of segmented objects for an entire time series_ ever exceeds it.

DATA_FORMAT.md
```diff
@@ -0,0 +1,372 @@
...
+
+![](./documentation/frame_example.png)
+
+---
+
+</details>
+
+### 4. Features
+
+Datasets can contain any number of `features`, which are a numeric value assigned to each object ID in the dataset. Each feature file usually corresponds to a single column of data.
```
toloudis, line 210:
```suggestion
Datasets can contain any number of `features`, which are a numeric value assigned to each object ID in the dataset. Each feature file corresponds to a single column of data.
```

DATA_FORMAT.md
```diff
@@ -0,0 +1,372 @@
...
+
+### 7. Outliers (optional)
+
+The outliers file stores whether a given object ID should be marked as an outlier using an array of booleans (`true`/`false`). Indices that are `true` indicate outlier values, and are given a unique color in Nucmorph-Colorizer.
+
+```
+--outliers.json--
+{
+    "data": [
+        <whether track id 0 is an outlier>,
```
toloudis, line 279:
```suggestion
        <whether object id 0 is an outlier>,
```

DATA_FORMAT.md
```diff
@@ -0,0 +1,372 @@
...
+}
+```
+
+---
+
+</details>
+
+## Collections
+
+Collections are defined by an optional JSON file and group one or more datasets for easy access. Nucmorph-Colorizer can parse collection files and present its datasets for easier comparison and analysis from the UI.
```
toloudis, line 319: Nit/future task: we really need to rename this thing as it's not only for nucmorph anymore 😀

README.md
```diff
@@ -55,8 +70,6 @@ by the web client.
 ```
 python timelapse-colorizer-data/generate_data.py --output_dir {output_dir} --dataset {dataset_name}
 ```
-Note: Currently, Nucmorph-Colorizer only searches one on-prem storage location for data, though adding support for external datasets via URL is on our roadmap of future changes. _(See [Issue #34](https://github.com/allen-cell-animated/nucmorph-colorizer/issues/34).)_
-To add new datasets to the web client, you'll need to upload them to the existing folder. More details can be found in [`generate_data.py`](./scripts/timelapse-colorizer-data/generate_data.py)
```
toloudis, line 59: 🚀

DATA_FORMAT.md
toloudis: We can't do "show path" without centroids.  So I'd only say it's optional if that is handled gracefully.  And yes, it seems to me that tracks and times SHOULD be required!  And at least one feature (until we support color-by-track-id)!

ShrimpCryptid: The show path option currently just won't do anything without centroid data. In the future I could disable the checkbox in the UI and add a tooltip indicating why it's disabled...

ShrimpCryptid:
> Rather than linking to a stable dataset, I think we could just add a very trivial tiny fake dataset to the repo. Something with like 3 time steps, just a couple of tracked objects, so the arrays are very short. Could modify the generate_test_data script to generate it.

Here's a preview of the modifications to `generate_test_data.py`, and the new example dataset that's been added.

http://dev-aics-dtp-001.corp.alleninstitute.org/nucmorph-colorizer/dist/?dataset=http%3A%2F%2Fdev-aics-dtp-001.corp.alleninstitute.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Fdataset1

DATA_FORMAT.md
ShrimpCryptid: Fixed, thank you!

DATA_FORMAT.md
```diff
@@ -0,0 +1,374 @@
...
+    "bounds": <relative path to bounds JSON>        //< optional
+}
+
+
+*Note: all paths are relative to this JSON file.
+```
+
+Note that the `outliers`, `centroids`, and `bounds` files are optional, but certain features of Nucmorph-Colorizer won't work without them.
+
+A complete example dataset is also available in the [`documentation`](./documentation/example_dataset) directory of this project, and can be [viewed on Nucmorph-Colorizer](http://dev-aics-dtp-001.corp.alleninstitute.org/nucmorph-colorizer/dist/?dataset=https://raw.githubusercontent.com/allen-cell-animated/nucmorph-colorizer/main/documentation/example_dataset/manifest.json).
```
ShrimpCryptid, line 47: This link will only work once this PR is merged into main.

toloudis (approved):
LGTM!  Maybe now the DATA_FORMAT.md could go in documentation?

documentation/DATA_FORMAT.md
```diff
@@ -0,0 +1,374 @@
...
+
+---
+
+</details>
+
+## Getting Started
+
+For most datasets, the easiest way to get started is to modify one of the existing data generation scripts, like [`generate_data.py`](../scripts/timelapse-colorizer-data/generate_data.py).
+
+(Check with your team or one of the developers on the Animated Cell team to see if there's already a data generation script for your project!)
```
frasercl, line 374: This might be better placed at the start of the file so that people know immediately whether they need to read the rest or not

documentation/DATA_FORMAT.md
```diff
@@ -0,0 +1,374 @@
...
+
+---
+
+</details>
+
+### 4. Features
+
+Datasets can contain any number of `features`, which are a numeric value assigned to each object ID in the dataset. Each feature file corresponds to a single column of data.
+
+Features must also provide a `min` and `max` range property.
```
frasercl, line 214:
```suggestion
Features should also provide a `min` and `max` range property.
```
Technically, the app will derive `min` and `max` as needed, it'll just have to iterate over the whole array to do so

documentation/DATA_FORMAT.md
```diff
@@ -0,0 +1,374 @@
...
+
+![](./frame_example_simple.png)
+
+---
+
+</details>
+
+### 4. Features
+
+Datasets can contain any number of `features`, which are a numeric value assigned to each object ID in the dataset. Each feature file corresponds to a single column of data.
```
frasercl, line 212: Might add a sentence here to say that features are what this colorizer actually colorizes. While it's not necessarily this document's job to express this, we might as well make it clear here that displaying this data is the primary purpose of the app.

documentation/DATA_FORMAT.md
```diff
@@ -0,0 +1,374 @@
...
+
+- **Dataset**: A dataset is a single time-series, and can have any number of tracked objects and features.
+- **Collection**: An arbitrary grouping of datasets.
+- **Object ID**: Every segmentation object in every frame has an integer identifier that is unique across all time steps. This identifier will be used to map an object to relevant data. Object IDs must be sequential, starting from 0, across the whole dataset.
+
+## Dataset
+
+A dataset consists of a group of files that describe the segmentations, tracks, feature data, processed images, and additional metadata for a single time-series.
+
+The most important file is the **manifest JSON**, which describes all the files in the dataset. (Manifests should be named `manifest.json` by default.)
```
frasercl, line 19: nit: I might just call it **manifest**?

frasercl (approved):
Overall, this is an excellent piece of documentation! None of my suggestions are blocking.

Issue #72, opened 2023-08-25
Set frame skip in export


ShrimpCryptid:
## Use Case

Requested by @jessicasyu!
When working with large datasets, sometimes it's redundant to export every frame.

## Acceptance Criteria

- Add a numeric input to the export dialog that allows the user to specify the number of frames to skip when exporting image sequences.

## Details
- Should the exported image names reflect the original frame #'s (0, 2, 4, 6) or always be in 0, 1, 2, sequence?

jessicasyu:
I vote for keeping the original frame numbers. It helps to be able to match exported frames back to the original sequence and I think most methods for "set of images to movie" can handle non-sequential numbering.

toloudis:
Yeah, the big question was whether tools like photoshop or whatever can read in image sequences that are disjointed like that.

Issue #30, opened 2023-06-20
RecordingControls should use TimeControls


ShrimpCryptid:
## Use Case

Internal-only, non-user affecting. Reduce code complexity by having TimeControls and RecordingControls' advancement loops use the same underlying logic, just with separate callback behavior.

## Acceptance Criteria
- When recording, the interval loop should be completely contained in TimeControls.
- Any side-effects during recording should be done as callbacks.
- Wrapping or termination behavior should be handled via callbacks.

## Details
e.g.:
```private playTimeSeries(onNewFrameCallback: () => void, onCompleted: () => void, intervalMs: number): void {}```
- Note: ok to have separate loops for timeout vs. interval if needed
- separate public entrypoint for recording

toloudis:
also to consider:
TimeControls was intended to encapsulate all time handling.  It also encapsulates the user controls and the handlers we create. Consider a separate module called TimeController that just has the playback loop logic (with some convenient api on top of it) and TimeControls is ONLY a container for the controls and setting up their handlers, which then manipulate the TimeController.

toloudis:
This ticket is also a good opportunity to pull out the code that gets DOM controls and binds handlers, and put it back into main, so that TimeControls and RecordingControls just contain the handlers themselves.  In which case I would change the class names to Controller instead of Controls.

ShrimpCryptid:
Can we use just one timeout? (timeout instead of setInterval?)

PR #102, opened 2023-09-26
feat: UI Update 4 - Export Modal

### LINKED ISSUE #79: Refactor TimeControls and RecordingControls
## Acceptance Criteria
- Separate out the UI elements for time-related inputs (seeking, forward/backward movement) into a component.
- RecordingControls should also be separated, but likely into a modal (see https://github.com/allen-cell-animated/nucmorph-colorizer/issues/37)
- Move existing logic for handling time/recordings into their own utility files or classes?

## Details
- Semi-related to https://github.com/allen-cell-animated/nucmorph-colorizer/issues/30.

ShrimpCryptid:
Problem
=======
Adds functionality for the Export feature, allowing image sequences to be exported from the currently loaded dataset.

Closes #79, #30, and #72.

Solution
========
- Rewrote `RecordingControls` to be simpler! It's now agnostic to `ColorizeCanvas`; instead, an options object and callbacks are passed to it when a recording is started. 
- Implemented UI interactions for the `ExportButton.tsx` modal, including canceling/aborting operations and range selection.

Also:
- Renamed `ThemeContext` to `AppThemeContext` for clarity and to avoid conflicts with an Antd export by the same name.
- Some CSS cleanup in various components.

## Type of change

Screenshots (optional):
--------------

I did a voiceover for this one! Let me know if that's more helpful than me pointing my mouse everywhere 😊
*Note: There's been some changes to the Stop/Cancel popups based on feedback from Lyndsay. The most recent version will always be available via the interactive demo.*

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/5cf1f1ca-144b-496a-8734-c9315fb37425

INTERACTIVE VERSION: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-102/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Fmain%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=feature_0

-----------------------
1. ExportButton.tsx
2. RecordingControls.ts

Thanks for contributing!


src/colorizer/RecordingControls.ts
```diff
@@ -1,110 +1,121 @@
-import ColorizeCanvas from "./ColorizeCanvas";
+export type RecordingOptions = {
+  /** Frame to start recording from, inclusive. 0 by default. */
+  min: number;
+  /** Highest frame number to record, inclusive. 0 by default. */
+  max: number;
```
ShrimpCryptid, line 5: Should start take in the range as required parameters rather than making them part of the options? This feels a bit cleaner to me but thought I should ask.

src/App.tsx
```diff
@@ -575,9 +588,17 @@ function App(): ReactElement {
...
+            totalFrames={dataset?.numberOfFrames || 0}
+            setFrame={setFrame}
+            currentFrame={currentFrame}
+            startRecording={(options: Partial<RecordingOptions>) => {
+              recordingControls.start(setFrameAndRender, getCanvasImageAsUrl, options);
+            }}
+            stopRecording={() => recordingControls.abort()}
+            defaultImagePrefix={datasetKey + "-" + featureName + "-"}
+            disabled={dataset === null}
+          />
```
ShrimpCryptid, line 601: App.tsx acts as a bridge between RecordingControls and the export button here. Both are mostly agnostic to the other besides shared parameter typing.

ShrimpCryptid:
@lynwilhelm 👀

src/components/ExportButton.tsx
```diff
@@ -0,0 +1,354 @@
...
+        // Close modal once recording finishes and show completion notification
+        setPercentComplete(100);
+        notification.success({
+          message: "Export complete.",
+          placement: "bottomLeft",
+          duration: 4,
+          icon: <CheckCircleOutlined style={{ color: theme.color.text.success }} />,
+        });
+        // Close the modal after a small delay so the success notification can be seen
+        setTimeout(() => stopRecording(true), 750);
```
toloudis, line 197: Nit: this kind of thing always scares me.  It makes me wonder how our app would respond to "crazy clicking" where you just go around clicking on everything randomly, really fast and see what breaks.

src/components/ExportButton.tsx
toloudis, line 1: Nit: feels a bit weird to call this ExportButton when it has the dialog with all the controls too.  Maybe Export.tsx ?

src/colorizer/RecordingControls.ts
toloudis: My take: I don't think it's too different either way.  We can always change it later if there's a good reason.

toloudis (approved):
Looks good, love the video.  As I said earlier, this is a great example of a tutorial-type video that we should make available to our USERS!!

src/components/ExportButton.tsx
ShrimpCryptid: Valid! I added an extra flag for when the closing animation is running because it was possible to stop and restart the recording (and then have the modal close on you :X).

It is _also_ possible to have the Cancel confirmation stay open after the window has been closed, but I don't think this causes any aberrant behavior (clicking either option won't reopen the Export window) so it's probably fine.

src/components/Export.tsx
```diff
@@ -0,0 +1,358 @@
...
+  // If dataset changes, update the max range field with the total frames.
+  useEffect(() => {
+    setCustomMax(props.totalFrames - 1);
+  }, [props.totalFrames]);
+
+  useEffect(() => {
+    if (useDefaultImagePrefix) {
+      setImagePrefix(props.defaultImagePrefix);
+    }
+  }, [props.defaultImagePrefix, useDefaultImagePrefix]);
```
frasercl, line 96: I'd recommend using a deriver rather than an effect here:

```typescript
const getImagePrefix = () => useDefaultImagePrefix ? props.defaultImagePrefix : imagePrefix;
```

I believe it's recommended to avoid using effects to update state, since state updates will trigger a second render and a second dependency check for all your effects.

src/components/Export.tsx
```diff
@@ -0,0 +1,358 @@
...
+
+  // Store the current frame whenever the modal opens so we can reset to it when
+  // the modal is closed.
+  useEffect(() => {
+    if (isLoadModalOpen) {
+      setOriginalFrame(props.currentFrame);
+    }
+  }, [isLoadModalOpen]);
```
frasercl, line 104: Again setting state in effects, though here I'd recommend replacing this with a ref. A rule of thumb I've started thinking about for state is: if somehow this value changed for no reason, would I expect the appearance of the component to change? If the answer is "yes" or even "maybe," it belongs in state. If not, a ref is a fine place for it.

In place of this effect, I'd change `originalFrame` to a ref and set it in the `isLoadModalOpen` setter:

```typescript
const originalFrameRef = useRef(props.currentFrame);
const [isLoadModalOpen, _setIsLoadModalOpen] = useState(false);
const setIsLoadModalOpen = (isOpen) => {
  _setIsLoadModalOpen(isOpen);
  if (isOpen) {
    originalFrameRef.current = props.currentFrame;
  }
}
```

As a bonus, you could then remove the value from the dependency list of `stopRecording` below, since the latest value of the ref will always be accessible there.

src/components/Export.tsx
```diff
@@ -0,0 +1,358 @@
...
+                <Progress
+                  style={{ marginRight: "8px", verticalAlign: "middle" }}
+                  type="circle"
+                  size={theme.controls.heightSmall - 6}
+                  percent={percentComplete}
+                  showInfo={false}
+                  strokeColor={percentComplete === 100 ? theme.color.text.success : theme.color.theme}
+                  strokeWidth={12}
+                />
+              </Tooltip>
```
frasercl, line 247: Kind of wild of Ant to make you put the code for what goes at the bottom of the modal before the code for the content, but what can you do

frasercl (approved):
Couple of React style comments, but this is looking really good overall!

src/components/Export.tsx
ShrimpCryptid: Is there a difference between making this a function vs. a constant value?

src/components/Export.tsx
ShrimpCryptid: Ant moment lol 🐜

src/components/Export.tsx
ShrimpCryptid: Thanks for the code example here too!

src/App.tsx
```diff
@@ -515,7 +517,14 @@ function App(): ReactElement {
   // TODO: TimeControls and RecordingControls should be refactored into components
   // and receive setFrame as props.
   timeControls.setFrameCallback(setFrame);
-  recordingControls.setFrameCallback(setFrame);
+
+  const setFrameAndRender = useCallback(
+    async (frame: number) => {
+      await setFrame(frame);
+      await canv.render();
```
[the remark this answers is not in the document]
ShrimpCryptid: I wish render was awaitable...

Issue #69, opened 2023-08-25
Grey values out of range instead of hiding


ShrimpCryptid:
## Use Case

Requested by @jcass11 and @jessicasyu!
The requested feature is to indicate the existence of cells even if they're not currently being colored, giving them an alternate color instead of hiding them.

## Acceptance Criteria

- Cells outside of the range should have an alternate color applied to them instead of color ramp values.
- Add some UI element to turn on/off this feature.

toloudis:
end goal is that values out of range can be either hidden OR use an assigned color.

Issue #91, opened 2023-09-05
Threshold using multiple features


ShrimpCryptid:
## Use Case

Consider only cells that meet certain criteria defined by one or more feature ranges.

## Acceptance Criteria

- TBD, needs UX design

## Details

- Prerequisite: https://github.com/allen-cell-animated/nucmorph-colorizer/issues/90
- Needs UX
- Dropdown panel? Query box?
- Does the filtering dropdown include every feature type in every dataset? How do we keep ranges across datasets?

ShrimpCryptid:
Discussion with Antoine:
- Fast data visualization, slicing data
ex: What is the effect of environmental conditions on cell speed? -> looking at different criteria (cells w/ 3+ neighbors vs. cells w/ <3 neighbors)

- Looking at cells over time based on speed of cells, for JUST cells 
"Assigning classes to cells using multiple criteria" -> filtering cells by feature value to validate perceived grouping/classes
- Value threshold
  - Dynamic filter (whatever objects CURRENTLY match the filter)
  - Lock a certain subset of tracks at a defined frame and ONLY see those across the rest of the dataset

- Lots of future nice-to-haves, but don't want to slow down development. Ideally something that works NOW, and can be improved later.
- No immediate need to do more than single min + max for a feature
- Ideally no limit to number of filters/thresholds across features :)

toloudis:
what if user changes color-by feature to a feature that is already in our filter set?   how do we handle that?

ShrimpCryptid:
First implementation ideas:
1. pass in an array of sampler2D's into the shader, along with an accompanying min/max range array. 
2. Pre-calculate out-of-range values for all object ID's according to the current filter, and pass it in as a sampler2D.
  1. Naive approach is to use the CPU for this, do operations on a buffer
  2. GPU-based approach could use a buffer texture that is updated with each feature + its min/max every time the filters change.

Issue #90, opened 2023-09-05
Threshold using a separate feature


ShrimpCryptid:
## Use Case

Use a separate feature threshold than the one that's currently being visualized to mask/grey out cells. Requested by @antoineborensztejn!
ex: Threshold by distance from colony but visualize # of neighbors

## Acceptance Criteria
- Users can define a set feature range and apply thresholds for the feature, outside of which cells/objects are invisible or greyed out ("hide out of range")
- Users should still be able to see the color ramp from a different feature.

## Details

- Needs UX

ShrimpCryptid:
Closing this, since it's now redundant with #91

PR #105, opened 2023-09-29
feat: UI Update 5 - Custom coloring for out-of-range and outlier values.

ShrimpCryptid:
Problem
=======
Closes https://github.com/allen-cell-animated/nucmorph-colorizer/issues/69, "Grey values out of range instead of hiding".

Solution
========
- Adds a new `DrawModeSelector`, which lets users pick between three drawing options (hidden, solid color, and color map) for outliers and out-of-range values.
- Updates the ColorizeCanvas API to accept a DrawMode and Color for outliers + out-of-range values.

Misc:
- Fixes a bug where color map range min/max would not update correctly when swapping between datasets.
- Adds a sampling method for the ColorRamp. This was part of a discontinued feature (showing the current color ramp in the color picker) but it felt like it was worth keeping in the codebase.

## Type of change

Screenshots (optional):
-----------------------

Interactive preview: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-105/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Fmain%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=feature_0

(Turn audio on!)

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/856b86e9-ba4d-43ce-8081-b2e2180ca746

-----------------------
1. `DrawModeSelector.tsx`
2. `App.tsx`

Thanks for contributing!


src/App.tsx
```diff
@@ -412,9 +429,10 @@ function App(): ReactElement {
       }
       setFeatureName(newFeatureName);
 
-      if (!isColorRampRangeLocked) {
-        setColorRampMin(newDataset?.features[newFeatureName].min || colorRampMin);
-        setColorRampMax(newDataset?.features[newFeatureName].max || colorRampMax);
+      const featureData = newDataset.getFeatureData(newFeatureName);
+      if (!isColorRampRangeLocked && featureData) {
+        setColorRampMin(featureData.min);
+        setColorRampMax(featureData.max);
```
ShrimpCryptid, line 435: Before, datasets with a min feature value of 0 would not update the `colorRampMin` (because 0 is falsy).

src/colorizer/ColorRamp.ts
```diff
@@ -39,4 +39,32 @@ export default class ColorRamp {
...
+    if (maxIndex === minIndex) {
+      return new Color(this.colorStops[minIndex]);
+    }
+
+    // Get a new normalized t value between the min and max indices, range [0, 1]
+    const tNormalized = (tIndex - minIndex) / (maxIndex - minIndex);
+    const minColor = new Color(this.colorStops[minIndex]);
+    const maxColor = new Color(this.colorStops[maxIndex]);
+    return minColor.lerp(maxColor, tNormalized);
+  }
```
ShrimpCryptid, line 69: This isn't used anywhere, but I wanted to keep it from some experimental changes I'd made because it seems like it would come in handy.

src/components/LabeledDropdown.tsx
ShrimpCryptid, line 1: I made some changes to LabeledDropdown (which DrawModeSelector uses) so that it actually shows the `label` of the selected item, not just the `key` values. I also made the `label` prop optional, and it's now hidden if nothing is provided.

toloudis:
@lynwilhelm @ShrimpCryptid I want to give a headsup for #91 and #90 to consider where this functionality belongs in the layout.  #90 will be the very next thing to implement, that affects ui significantly.

Also it just occurred to me that all the options result in a single color being applied for values above and below the range.  Except for "use color map" which results in different colors for values above versus values below!

src/colorizer/ColorizeCanvas.ts
```diff
@@ -248,8 +266,24 @@ export default class ColorizeCanvas {
     this.setUniform("backgroundColor", color);
   }
 
-  setOutlierColor(color: Color): void {
-    this.setUniform("outlierColor", color);
+  setHideValuesOutOfRange(hide: boolean): void {
+    this.hideValuesOutOfRange = hide;
+    this.setUniform("hideOutOfRange", this.hideValuesOutOfRange);
```
toloudis, line 271: this whole function looks like it could be deleted?  all the functionality is in setOutOfRangeDrawMode

src/colorizer/ColorizeCanvas.ts
```diff
@@ -248,8 +266,24 @@ export default class ColorizeCanvas {
...
+      this.setUniform("outlierColor", color);
+    }
+  }
+
+  setOutOfRangeDrawMode(mode: DrawMode, color?: Color): void {
+    this.setUniform("outOfRangeDrawMode", mode);
+    if (mode === DrawMode.USE_COLOR && color) {
+      this.setUniform("outOfRangeColor", color);
+    }
+    console.log("Mode: " + mode);
```
toloudis, line 286: can remove console log

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -9,7 +9,15 @@ uniform float featureMax;
 uniform vec2 canvasToFrameScale;
 uniform sampler2D colorRamp;
 uniform vec3 backgroundColor;
+
+const uint DRAW_MODE_HIDE = 0u;
```
toloudis, line 13: it might be useful to have a comment above these to say they must be kept in sync with the enum in ColorizeCanvas (and vice versa)

src/components/DrawModeSelector.tsx
```diff
@@ -0,0 +1,105 @@
...
+`;
+
+/**
+ * UI element for choosing between different drawing modes, and provides callbacks for when
+ * changes are made to selections.
+ * - `HIDE`: Hide an object type
+ * - `USE_COLOR`: Use a custom, solid color. (When selected, also shows a ColorPicker.)
+ * - `USE_RAMP`: Use the color ramp.
+ */
+export default function DrawModeSelector(propsInput: DrawModeSelectorProps): ReactElement {
```
toloudis, line 40: I think `Selector` is an overloaded term that is used in `Redux` to mean something totally different that is not a component.  I'll defer to others who know more about React/Redux as to how important it is to rename this. (@frasercl @meganrm )

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Oops, I was not as careful during my cleanup pass as I thought I was apparently 😅

toloudis (approved):

src/components/DrawModeSelector.tsx
frasercl: Library authors definitely like to use the generic word "selector" to mean lots of different non-user-facing things, but I don't think it would be disastrously confusing to see it as the name of a component. (I probably wouldn't ever expect a selector as Redux uses it to appear as a component.) If I were to rename the component I think I'd go with `DrawModeDropdown` to specify the generic term?

ShrimpCryptid: There's also a ColorRampSelector currently that I could also rename (that is also a dropdown). Thank you both for the input!

src/App.tsx
```diff
@@ -28,6 +28,8 @@ import IconButton from "./components/IconButton";
 import SpinBox from "./components/SpinBox";
 import HoverTooltip from "./components/HoverTooltip";
 import Export from "./components/Export";
+import DrawModeDropdown from "./components/DrawModeDropdown";
+import { Color } from "three";
```
frasercl, line 32: World's smallest nit: move this external import up to live with the react/antd imports

src/App.tsx
```diff
@@ -115,11 +125,16 @@ function App(): ReactElement {
    * render. Also update other UI elements, including the plot and URL.
    */
   useEffect(() => {
+    // TODO: This is getting larger... ColorizeCanvas needs to be refactored to take
+    // all of these in as props.
```
frasercl, line 129: Seeing this makes me think of website-3d-cell-viewer, where these prop updates each have their own effect so that one change doesn't call every update method... But several of volume-viewer's props are expensive to update, and pretty much all of the updaters below make cheap updates to shader uniforms. Sounds like this would be a convenient and appropriate pattern here, definitely worth keeping in mind (or in a TODO) for the future!

frasercl (approved):
Looks good!

Issue #106, opened 2023-10-03
UI Formatting Fixes


ShrimpCryptid:
## Description
Submitted by @lynwilhelm!

https://www.figma.com/file/DcFEsrszbUOK0qU1E9wrH0/2023-Aug-NucMorph-Viz-App?type=design&node-id=0-1&mode=design&t=2vKRlNZpBtoXvk8S-0

Tasks:
- [x] Change Color Map to sentence case (Color map)
- [x] Slider text should be 12px, spaced from slider
- [x] Increase size of icons in playback buttons and center
- [x] Change slider background to #f2f2f2
- [x] Change Search hint text to #A3A4A5
- [x] Change border colors to #cbcbcc
  - [x] Color Map
  - [x] Feature dropdown
  - [x] Checkboxes
  - [x] Input areas
- [x] Remove box shadow from all primary buttons
- [x] Adjust primary + default button hover + active styling
- [x] Adjust primary dropdown hover + active colors
- [x] Adjust default dropdown hover + active colors
- [x] Adjust color map hover + active colors
- [x] Add spacing to dropdown styling
- [x] Adjust colors in dropdown styling
- [x] Add tooltips to dropdowns(????)
- [x] Decrease max text length in dropdowns (text clipping)
- [x] ~~Change text alignment/cutoff in Dataset/Feature dropdowns~~ Ignoring for now

Extras:
- [x] Make color ramp work with tab selection + click
- [x] Make dropdowns work with tab selection + click

Issue #98, opened 2023-09-13
UI Overhaul Accessibility Issues


ShrimpCryptid:
## Description

- [ ] Arrow key bindings for time still activate when selecting/editing text in input boxes
- [x] Dropdowns do not work with tab navigation
- [x] Color ramp selector does not work with tab navigation
- [ ] ~~Plotly doesn't work with tab navigation (unsure what expected behavior is though)~~
- [ ] Hover Tooltip may cause issues with screen readers?

ShrimpCryptid:
Reopened as #308 and #309.

PR #107, opened 2023-10-05
feat: UI Update 6 - Styling Fixes and Accessibility

ShrimpCryptid:
Problem
=======
Resolves https://github.com/allen-cell-animated/nucmorph-colorizer/issues/106, UI Formatting Fixes!

Also fixes some known accessibility issues I previously noted in https://github.com/allen-cell-animated/nucmorph-colorizer/issues/98.

Solution
========

See https://github.com/allen-cell-animated/nucmorph-colorizer/issues/106 for the full list!

- Refactors `LabeledDropdown` to strip out Ant's default menu handler. The new custom buttons respond to focus navigation and have custom styling.
- Refactors `ColorRampDropdown` (previously `ColorRampSelector.tsx`) to update button styling and make it responsive to focus navigation.
- General styling and sizing changes to buttons.

## Type of change

Screenshots (optional):
-----------------------
https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/8d217448-8392-473a-94ba-c8f96d1adbfc

https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-107/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Fmain%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=feature_0

base version for comparison: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-108/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Fmain%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=feature_0

-----------------------
1. `AppStyle.tsx`
3. `LabeledDropdown.tsx`
4. `ColorRampDropdown.tsx`

Thanks for contributing!


src/components/ColorRampDropdown.tsx
```diff
@@ -0,0 +1,118 @@
...
+      if (!doesContainTarget(event.relatedTarget)) {
+        setForceOpen(false);
+      }
+    };
+
+    componentContainerRef.current?.addEventListener("focusout", handleFocusLoss);
+    return () => {
+      componentContainerRef.current?.removeEventListener("focusout", handleFocusLoss);
+    };
+  }, [forceOpen]);
```
ShrimpCryptid, line 55: This is most of the new content in `ColorRampDropdown`, plus some className manipulation below.

ShrimpCryptid: (sorry that the diff is ugly, it didn't figure out that I renamed this file)

src/components/AccessibleTooltip.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type AccessibleTooltipProps = {
+  disabled?: boolean;
+  title?: string;
+  placement?: TooltipPlacement;
+};
+
+/**
+ * Layout-safe, focus-responsive tooltip! This is a wrapper around the Ant Tooltip,
+ * fixing some of the biggest issues with it.
```
toloudis, line 13: could be interesting to check antd's issues and see if others have brought it up

toloudis (approved):
Nit: It might be an optical illusion but it still looks to me like the baseline of the text for the Dataset and Feature dropdown labels is slightly different than the dropdown contents.  Like maybe by 1px?

Otherwise LGTM!  The comments are very helpful for knowing why some of the weird extras were added in the code.

src/components/AccessibleTooltip.tsx
ShrimpCryptid: 🙃 just realized there's a focus trigger mode in Tooltip that would have worked for this use case actually and I did not need to rewrite this from scratch

The layout issue still persists though when disabling tooltips, but this might let me fix the bug I noted with Dropdown...

ShrimpCryptid: This was actually a great comment because it turns out the issue I was having with Tooltips changing layout doesn't happen anymore? I'm not sure if it got fixed in an update ((or if Ant is just gaslighting me)) but I've removed `AccessibleTooltip.tsx` and simplified code in a few places.

ShrimpCryptid:
@toloudis Re-requesting based on the conversation we had in the Retro earlier (I removed a big chunk of code based on one of your earlier comments)

toloudis (approved):

src/components/LabeledDropdown.tsx
```diff
@@ -64,41 +97,89 @@ export default function LabeledDropdown(inputProps: LabeledDropdownProps): React
...
+          {item.label}
+        </Button>
       </Tooltip>
     );
-  }
+  });
+
+  // Can't use AccessibleTooltip because Ant passes props from Dropdown
+  // to the Button through the Ant Tooltip component >:(
+  // TODO: Fix AccessibleTooltip in this use case?
```
toloudis, line 144: is this comment obsolete now?

ShrimpCryptid: Good catch, yes!

src/components/ColorRampDropdown.tsx
```diff
@@ -41,20 +73,20 @@ const ColorRampSelector: React.FC<ColorRampSelectorProps> = (propsInput): ReactE
...
-      let className = "";
       // Manipulate class names for rounding at start and end of dropdown list
+      let className = "";
       if (i === 0) {
-        className += " " + styles.dropdownFirst;
+        className = styles.dropdownFirst;
       }
       if (i === props.colorRamps!.size - 1) {
-        className += " " + styles.dropdownLast;
+        className = styles.dropdownLast;
```
frasercl, line 82: tip: this can also be accomplished with the CSS `:first-child` and `:last-child` pseudo-classes

src/components/LabeledDropdown.module.css
```diff
@@ -5,11 +5,39 @@
   gap: 6px;
 }
 
-.labeledDropdown button {
+/* Make button and dropdown have the same width */
+#dropdownButton,
+.dropdownContent {
```
frasercl, line 10: Why is `dropdownButton` an id rather than a class? Ids should be unique items which appear only once in the document.

src/components/ColorRampDropdown.tsx
ShrimpCryptid: oh wait nvm GitHub figured it out!

src/components/LabeledDropdown.module.css
```diff
@@ -25,9 +53,41 @@
...
+
+.dropdownContent {
+  padding: 4px;
+  display: flex;
+  flex-direction: column;
+  gap: 2px;
+}
+
+#dropdownItem {
+  text-align: left;
```
frasercl, line 72: Same as above

src/components/ColorRampDropdown.tsx
```diff
@@ -23,12 +23,44 @@ const ColorRampSelector: React.FC<ColorRampSelectorProps> = (propsInput): ReactE
...
+        setForceOpen(false);
+      }
+    };
+
+    componentContainerRef.current?.addEventListener("focusout", handleFocusLoss);
+    return () => {
+      componentContainerRef.current?.removeEventListener("focusout", handleFocusLoss);
+    };
+  }, [forceOpen]);
+
```
frasercl, line 55: This is nice! Had a bit of trouble understanding the logic at first, but it makes the component feel like it "just works" for multiple input methods. I might at least mess around with making it a custom hook since it's also used in `LabeledDropdown`, but honestly I'm leaning towards it being fine to live in two places.

frasercl (approved):
Couple comments but overall this is looking good (both the code and the rendered app)!

src/components/ColorRampDropdown.tsx
ShrimpCryptid: I'll write a TODO to potentially refactor it into a hook later, but I'll leave it for now. Thanks for the suggestion!

ShrimpCryptid: Also improved the comments here, sorry that they were unclear!

src/components/LabeledDropdown.module.css
frasercl: `!important` is kind of a last-resort escape hatch, which is forgivable in certain cases (I've used it a couple times over in website-3d-cell-viewer). Ids let the browser assume that an element is unique and treat it differently, so it's probably worse practice not to preserve that. If turning these back into classes means you instead have to stick `!important` onto every style rule, that's not nice and we should try to find a secret third option, but I think I'd prefer to merge this with `!important` all over the place rather than assigning the same id to a list of elements.

Issue #33, opened 2023-06-29
Download image sequence as video/MP4


ShrimpCryptid:
## Use Case

@jcass11 and @vianamp requested the ability to directly download MP4 video for use on presentations and sharing.

## Acceptance Criteria

- Users can click a button to initiate a recording.
- Once the recording is finished, an encoded MP4 video will be downloaded to the user's Download directory (or the user will be prompted for a download location).
- Naming should default to the same as the existing RecordingControls.

## Details

- Similar to https://github.com/simularium/simularium-viewer/issues/252, which is running into technical issues with built-in browser security policies.

toloudis:
Also potentially relevant: https://github.com/simularium/simularium-viewer/pull/251

PR #109, opened 2023-10-09
feat: Export MP4 Video Files

ShrimpCryptid:
Problem
=======
Closes https://github.com/allen-cell-animated/nucmorph-colorizer/issues/33, "Download image sequence as video/MP4"!

Solution
========
Adds a new option in the Export modal for exporting to an MP4 video. This uses the WebCodecs browser API which provide faster recording speed and more control over video quality, but is unfortunately not yet supported in Firefox.

- Renamed `RecordingControls` to `CanvasRecorder`. It's now an abstract class that provides lifecycle methods for implementing classes to override.
- Added the `ImageSequenceRecorder`, which contains the previous logic for saving image sequences directly.
- New `Mp4VideoRecorder`, which saves each frame to a video.
  - Also adds quality (filesize) controls and video speed.
- The `Export` component now handles all details related to recording, and simply reports its status via a setter that's passed to it (`setIsRecording`). Also adds additional video controls and an export mode switcher.

Note:
- I also tried using the MediaRecorder API with ffmpeg to convert it from .webm to .mp4 and it was... not great. The video recording was really slow because it records in real-time, and the video would get corrupted or have lots of visible compression artifacts. Using ffmpeg also required me to download a 31 MB WASM file to the client browser and it took several minutes to process large video. WebCodecs is not as widely supported, but has a much smoother user experience and better video quality.

## Type of change

Screenshots (optional):
-----------------------
<p float="left">

</p>

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/98a88eae-bf36-429e-91ab-36aca2349fdd

Example exported video (Low quality, 30fps, ~11 MB)

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/1f403ba2-3f38-463f-ae6c-0b2bff67b985

This is probably one of the PRs where **testing the changes locally makes the most sense!** This feature really shines with a larger dataset, and the default example collection is only available over HTTP.

## Keyfiles

- `CanvasRecorder.ts`
- `Mp4VideoRecorder`
- `Export.tsx`


src/colorizer/recorders/CanvasRecorder.ts
```diff
@@ -0,0 +1,246 @@
...
+   * and `{prefix}.{ext}` for videos. */
+  prefix: string;
+  /** Delay between each recorded frame, in milliseconds. 100 ms by default. */
+  delayMs: number;
+  /** Frames per second for video output. Default is 30 fps.*/
+  fps: number;
+  /** Average video bitrate in bits/second. */
+  bitrate: number;
+  /** Width, height of output video or image. */
+  outputSize: [number, number];
```
ShrimpCryptid, line 32: fps, bitrate, outputSize, and onError are new!

src/colorizer/recorders/CanvasRecorder.ts
ShrimpCryptid, line 1: RecordingControls got renamed and changed here. A few new lifecycle steps (`setup`, `recordFrame`, and `onRecordingCompleted`) were added as protected methods to be overridden by child objects.

.eslintrc
```diff
@@ -33,9 +33,7 @@
     "@typescript-eslint/typedef": ["error", { "parameters": true }],
     "prefer-const": 1,
     "radix": 1,
-    "react/jsx-uses-vars": 0, // Project does not use react
-    "react/no-deprecated": 0,
-    "react/prop-types": 0,
+    "react/prop-types": 0, // This relies specifically on React PropTypes and ignores other typing methods
```
toloudis, line 36: I think all these changes make sense except for why remove "react/no-deprecated" ?

ShrimpCryptid: This removes the rule that disables the no-deprecated check! So in this case it's actually enabling it :)

src/colorizer/recorders/Mp4VideoRecorder.ts
```diff
@@ -0,0 +1,150 @@
+import { ArrayBufferTarget, Muxer } from "mp4-muxer";
+import CanvasRecorder, { RecordingOptions, defaultRecordingOptions } from "./CanvasRecorder";
+import { sleep } from "../utils/timing_utils";
+
+// Eslint doesn't recognize the WebCodecs API yet
+/* global VideoEncoder, VideoFrame, VideoEncoderConfig*/
```
toloudis, line 6: is this still meant to be here?

src/components/Export.tsx
```diff
@@ -196,138 +290,288 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
...
+    // It scales linearly with the bitrate until a maximum filesize is hit,
+    // which seems to depend on the video dimensions.
+
+    // Video quality is bitrate in bits/second
+    const maxVideoBitsDuration = totalSeconds * videoBitsPerSecond;
+
+    // Experimentally-determined compression ratio (bits per pixel). This may not apply for all
+    // datasets, depending on image complexity.
+    // For high bitrates, the video filesize will be constrained by duration.
+    const compressionRatioBitsPerPixel = 2.77;
```
toloudis, line 361: shouldn't this be relative to number and spacing of keyframes and also the bit rate (videoBitsPerSecond)?   Also, just wondering how accurate the final estimate is in practice.

src/colorizer/recorders/CanvasRecorder.ts
```diff
@@ -0,0 +1,255 @@
...
+        } else if (e instanceof ErrorEvent) {
+          this.options.onError((e as ErrorEvent).error);
+        } else {
+          // May throw Event type if the error is from an event listener.
+          // This happens most often when a resource fails to load
+          // (e.g., network issues or user not on VPN)
+          // TODO: Update error message if this tool becomes public!
+          this.options.onError(
+            new Error(
+              "Encountered an unknown error while exporting. See the console for more details. For Institute users, please check VPN status."
```
toloudis, line 157: I like the todo.  This message seems like it's making big assumptions about who is reading it.  Might want to replace "the console" with "the browser's Javascript console"...

src/components/Export.tsx
ShrimpCryptid: Updated this comment block to illustrate why this is needed and increased the value to overshoot estimates.

src/colorizer/recorders/Mp4VideoRecorder.ts
ShrimpCryptid: Yes, updated the comment to mention the specific linter errors that get thrown without it 😞

toloudis (approved):
Thank you for this great new feature! Output ready to be dropped into a Powerpoint!

I'm still skeptical about the file size estimate - I wonder if @lynwilhelm would say it's better to provide an inaccurate estimate if it's within a factor of 2-3, or to not give an estimate at all.

frasercl (approved):
Really nicely structured and designed, both in the code and the interface!

Issue #77, opened 2023-08-30
Refactor Plotting into a react component


ShrimpCryptid:
## Acceptance Criteria

- Plot should be invoked like a React component, where fields are passed in as props.
```
<Plotting
   dataset={...}
   selectedTrack={...}
   feature={...}
   time={...}
   ...
/>
```
- Can be a component-like wrapper around the existing class.

Issue #76, opened 2023-08-30
Refactor ColorizeCanvas into a react component


ShrimpCryptid:
## Acceptance Criteria
- ColorizeCanvas should be invoked from App like a component (e.g., current setters are replaced with props that are updated as part of React's render cycle).
```
<ColorizeCanvas
   dataset={...}
   featureName={...}
   selectedTrack={...}
   ramp={...}
   rampMin={...}
   rampMax={...}
   ...
/>
```
- Internally, ColorizeCanvas can still break the React render paradigm and update its `HTMLCanvas` separately.

## Details
- Unsure if this will break things like timeControls/recordingControls loops?

PR #111, opened 2023-10-10
refactor: Cleanup Internals

### LINKED ISSUE #32: encapsulate the feature color ramp controls in a module

ShrimpCryptid:
Problem
=======
Closes #32, #76, and #77! 

This refactor cleans up the internals of the app, moving some code into separate modules for simplicity.

Solution
========
- `ColorizeCanvas` now has a `CanvasWrapper.tsx` component! The canvas is still owned by `App` and passed down because it has async operations that are important for timing, but a lot of basic uniform updating/interaction event handling is now done by the wrapper.
- `Plotting` now has a `PlotWrapper.tsx` component. This completely encapsulates the plot, handling timing and track updates.
- Add a new component `LabeledRangeSlider.tsx` to replace identical code in `App`. (Also has some minor improvements for input handling.)
- `App.tsx` has been cleaned up! Refactoring the canvas and plot let me remove a lot of otherwise trivial, single-use functions and turn them into lambdas.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-111/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Fmain%2Fdocumentation%2Fexample_dataset%2Fmanifest.json

Keyfiles (in order of suggested review):
-----------------------
1. `CanvasWrapper.tsx`
2. `PlotWrapper.tsx`
3. `LabeledRangerSlider.tsx`
4. `App.tsx`


src/App.module.css
ShrimpCryptid, line 1: Moved CSS inline to `LabeledRangerSlider.tsx`

src/App.tsx
```diff
@@ -279,48 +220,10 @@ function App(): ReactElement {
...
-  useEffect(() => {
-    canv.domElement.addEventListener("click", handleCanvasClick);
-    // Returned callback is fired if/when App is removed from the DOM,
-    // or before running useEffect again. This prevents memory issues with
-    // duplicate event listeners.
-    return () => {
-      canv.domElement.removeEventListener("click", handleCanvasClick);
-    };
-  }, [handleCanvasClick]);
-
```
ShrimpCryptid, line 323: See `onTrackClicked` prop in render. Removed all references to plot here, and canvas now takes in the track as a prop, so `handleCanvasClick` reduces to two lines of code 😊

src/App.tsx
```diff
@@ -437,23 +335,9 @@ function App(): ReactElement {
...
-  );
-
-  const handleFeatureChange = useCallback(
-    (value: string): void => {
-      console.log(value);
-      if (value !== featureName && dataset) {
-        updateFeature(dataset, value);
-      }
-    },
-    [featureName, dataset, updateFeature]
```
ShrimpCryptid, line 456: Moved to lambda function since it's only referenced once

src/App.tsx
ShrimpCryptid, line 1: I'd start with `PlotWrapper.tsx` and `CanvasWrapper.tsx`! All references to the plot were removed, and setters for a lot of canvas functionality were removed and changed to a more React-style passing of props to the component.

src/components/LabeledRangeSlider.tsx
```diff
@@ -0,0 +1,155 @@
...
+    // Bit of a hack to override font size by increasing specificity
+    font-size: var(--font-size-label-small);
+    left: 0;
+  }
+
+  &:last-child {
+    font-size: var(--font-size-label-small);
+    right: 0;
+  }
+`;
```
ShrimpCryptid, line 70: Copied directly from `App.module.css`

src/App.module.css
ShrimpCryptid, line 1: Moved to LabeledRangeSlider.tsx

src/App.tsx
```diff
@@ -31,25 +31,18 @@ import SpinBox from "./components/SpinBox";
...
   // STATE INITIALIZATION /////////////////////////////////////////////////////////
   const theme = useContext(AppThemeContext);
 
-  const canvasRef = useRef<HTMLDivElement>(null);
-  const plotRef = useRef<HTMLDivElement>(null);
-
-  const [plot, setPlot] = useState<Plotting | null>(null);
   const canv = useConstructor(() => {
     return new ColorizeCanvas();
   });
```
ShrimpCryptid, line 44: Mentioned this in the description, but canv is still kept top-level here since some of its functionality is important for timing. I guess I *could* pass it up via a ref from `CanvasWrapper.tsx` but I think this might be cleaner?

src/App.tsx
```diff
@@ -276,51 +217,6 @@ function App(): ReactElement {
     setupInitialParameters();
   }, [isInitialDatasetLoaded]);
 
-  // Add event listeners for unloading and resizing on startup.
-  useEffect(() => {
-    window.addEventListener("beforeunload", () => {
-      canv.domElement.removeEventListener("click", handleCanvasClick);
-      canv.dispose();
-    });
```
ShrimpCryptid, line 284: Removing this because the browser is likely taking care of this, and the `beforeunload` event is unreliable in certain browser contexts (https://developer.chrome.com/articles/page-lifecycle-api/#the-beforeunload-event)

src/App.tsx
```diff
@@ -187,23 +142,13 @@ function App(): ReactElement {
...
+    [canv, dataset, featureName, currentFrame]
   );
 
-  /**
-   * Update the canvas dimensions based on the current window size.
-   *
-   * TODO: Find calculation for margin magic number
-   */
-  const setSize = (): void => canv.setSize(Math.min(window.innerWidth - 75, 730), Math.min(window.innerHeight, 500));
-
```
ShrimpCryptid, line 206: Now handled by `CanvasWrapper.tsx`

src/App.tsx
```diff
@@ -467,37 +344,8 @@ function App(): ReactElement {
...
-  useEffect(() => {
-    canv.domElement.addEventListener("mousemove", onMouseMove);
-    canv.domElement.addEventListener("mouseleave", onMouseLeave);
-    return () => {
-      canv.domElement.removeEventListener("mousemove", onMouseMove);
-      canv.domElement.removeEventListener("mouseleave", onMouseLeave);
-    };
-  }, [onMouseMove, onMouseLeave, canv]);
-
```
ShrimpCryptid, line 499: Handled by `CanvasWrapper.tsx` now

src/components/Export.tsx
```diff
@@ -5,6 +5,7 @@ import SpinBox from "./SpinBox";
 import { RecordingOptions } from "../colorizer/RecordingControls";
 import { AppThemeContext } from "./AppStyle";
 import { CheckCircleOutlined } from "@ant-design/icons";
+import { clamp } from "three/src/math/MathUtils";
```
toloudis, line 8: Nit: Split imports into local dependencies vs node_modules dependencies -- helps readability.  I suspect there might be a Prettier setting for this but not sure?

toloudis (approved):
This looks ok to me

src/components/Export.tsx
ShrimpCryptid: I couldn't find a prettier rule for this without installing a plugin, but there's an eslint rule for this, it looks like! If I enable it, it will, unfortunately, make me fix every import that's out of order across the whole project.

I'm going to add this as a TODO for now so I don't clutter up the PR, and I'll fix this instance manually 😅

toloudis: Thanks! Low priority but nice to have... 👍   (I feel like we have this in effect in at least one codebase but I might be mistaken)

src/components/CanvasWrapper.tsx
```diff
@@ -0,0 +1,183 @@
...
+  useMemo(() => {
+    const settings = props.outlierDrawSettings;
+    canv.setOutlierDrawMode(settings.mode, settings.color);
+  }, [props.outlierDrawSettings]);
+
+  // Updated track-related settings
+  useMemo(() => {
+    canv.setSelectedTrack(props.selectedTrack);
+    canv.setShowTrackPath(props.showTrackPath);
+  }, [props.selectedTrack, props.showTrackPath]);
```
frasercl, line 88: I'd again prefer these to be `useEffect`s, though if there's a good lifecycle reason not to do that (e.g. you found from testing that `useEffect` happens too late for interactions to feel nice) then this is an acceptable approach.

src/App.tsx
frasercl: This is cleaner.

frasercl (approved):
Top-level comment: any idea what's causing the `Warning: findDOMNode is deprecated in StrictMode` error I'm seeing in the console? Looks to be something within Ant...?

Other than that (and my one other comment), this looks good! Nice and clean.

src/components/CanvasWrapper.tsx
ShrimpCryptid: Because `canv` breaks the React lifecycle model, these updates need to happen as part of the render or else React won't redraw the canvas (and then things like changing the color ramp will happen one update too late ;-;).

src/App.tsx
toloudis: These event listeners that hold on to references to things like canv, and would prevent canv from ever being garbage collected.   In this case, I think we generally expect the listener and the canvas to last for the whole browser session and won't get re-created over and over, right?

src/components/CanvasWrapper.tsx
toloudis: Maybe you could add a comment saying why this should be a useMemo?

toloudis (approved):

src/App.tsx
ShrimpCryptid: Yeah, I can't see a situation where we'd need to remake the canvas.

ShrimpCryptid:
> Top-level comment: any idea what's causing the `Warning: findDOMNode is deprecated in StrictMode` error I'm seeing in the console? Looks to be something within Ant...?
> 
> Other than that (and my one other comment), this looks good! Nice and clean.

I found out what was causing this! It's an issue with how I'm using the Tooltip, I think.

```
      <Dropdown...>
        <Tooltip ...>
          {dropdownButton}
        </Tooltip>
      </Dropdown>
```

I think the only workaround I can see would be kind of gross (make an invisible div with the same position and size as the dropdown button)...

PR #112, opened 2023-10-13
refactor: Cleanup for data scripts

ShrimpCryptid:
Problem
=======
Cleans up data generation scripts for easier use by scientists!

Solution
========
1. Moves most of the file writing logic to a new `data_writer_utils.py` file, with improved docstring and inline commenting.
2. Updates project Python dependencies and documentation.
3. `generate_data.py` and `generate_data_2.py` (renamed to `convert_nucmorph_data.py` and `convert_emt_migration_data.py`) are now greatly simplified and nearly identical. I recommend using VSCode's comparison tool to diff them.

Partly peer programmed with @toloudis !

## Type of change

Steps to Verify:
----------------
I have successfully run the whole data pipeline for the `baby_bear` script without crashing, and output images are identical to the original dataset. None of the object IDs have changed, and all data seems to be included correctly.

If you'd like to see the dataset, you can run the client locally and load the dataset from this link: `http://dev-aics-dtp-001.corp.alleninstitute.org/dan-data/colorizer/data/test/new-dataset/baby_bear/`


scripts/timelapse-colorizer-data/data_writer_utils.py
ShrimpCryptid, line 1: Most of this logic is copied from `generate_data.py`.

README.md
```diff
@@ -57,6 +57,8 @@ Utilities are included in this repository to convert time-series data generated
 
 For loading of datasets to work correctly, you'll need to run these commands from a device that has access to Allen Institute's on-premises data storage. If running off of shared resources, remember to initialize your virtual environment first! This may look like `conda activate {my_env}`.
 
+NOTE: You will also need to run Python version 3.8 due to the project's dependencies. For conda, you can create an environment with `conda create --name {my_env} python=3.8`.
```
toloudis, line 60: I think we can manage this dependency differently.  It really only matters for generate_data.py and not for data_writer_utils or generate_data_2.  I'm not sure there's anything to change here except maybe to note exactly when/why you'd need 3.8.

scripts/timelapse-colorizer-data/generate_data.py
```diff
@@ -19,7 +22,7 @@
 # python timelapse-colorizer-data/generate_data.py --output_dir /allen/aics/animated-cell/Dan/fileserver/colorizer/data --dataset mama_bear
 # python timelapse-colorizer-data/generate_data.py --output_dir /allen/aics/animated-cell/Dan/fileserver/colorizer/data --dataset goldilocks
 
-# DATASET SPEC:
+# DATASET SPEC: See DATA_FORMAT.md for more details.
```
toloudis, line 25: I guess most of the lines below could be moved out of here and into data_writer_utils

scripts/timelapse-colorizer-data/generate_data.py
```diff
@@ -139,107 +117,62 @@ def make_frames(grouped_frames, output_dir, dataset, scale: float):
...
-        json.dump(trjs, f)
-
-    logging.info("Writing times.json...")
-    times = a["index_sequence"].to_numpy()
-    tijs = {"data": times.tolist()}
-    with open(outpath + "/times.json", "w") as f:
-        json.dump(tijs, f)
-
-    logging.info("Writing centroids.json...")
+    tracks = a[TRACK_ID_COLUMN].to_numpy()
```
toloudis, line 128: why isn't "is_outlier" ( the line above this one)  a symbolic constant?  I'm wondering whether mapping the strings up above just adds noise to someone trying to quickly understand the code

toloudis: same for centroid_x and centroid_y

scripts/timelapse-colorizer-data/generate_data.py
```diff
@@ -61,76 +64,51 @@
...
-                bbox_data[write_index + 2 : write_index + 4] = bbox_max
+        # Scale the image and format as integers.
+        seg2d = writer.scale_image(seg2d)
+        seg2d = seg2d.astype(np.uint32)
 
-        # convert data to RGBA
-        seg_rgba = np.zeros(
-            (seg_remapped.shape[0], seg_remapped.shape[1], 4), dtype=np.uint8
+        # Remap the frame image so the IDs are unique across the whole dataset.
+        seg_remapped, lut = writer.remap_segmented_image(
```
toloudis, line 104: tiny naming nit: a writer isn't a thing that remaps images or scales them.

scripts/timelapse-colorizer-data/generate_data.py
```diff
@@ -139,107 +117,62 @@ def make_frames(grouped_frames, output_dir, dataset, scale: float):
...
+    reduced_dataset[INITIAL_INDEX_COLUMN] = reduced_dataset.index.values
+    grouped_frames = reduced_dataset.groupby(TIMES_COLUMN)
 
+    # Make the features, frame data, and manifest.
     nframes = len(grouped_frames)
-
     features = ["NUC_shape_volume_lcc", "NUC_position_depth"]
-    make_features(a, features, output_dir, dataset, scale)
-
+    make_features(full_dataset, features, writer)
```
toloudis, line 172: nit: i still wonder if this would be more lightweight if we passed reduced_dataset in (making sure it has the feature columns already) 🤷

toloudis:
I think it's probably time to rename the generate scripts with nucmorph and emt migration in their names.

scripts/timelapse-colorizer-data/generate_data.py
ShrimpCryptid: I think I'll remove this for now and continue pointing people to DATA_FORMAT.md via link! It'll be easiest if all the documentation lives in once place and doesn't have to be replicated... 😅

scripts/timelapse-colorizer-data/generate_data.py
ShrimpCryptid: I do think the symbolic names are a bit more clear than the original labels for some columns (ex: `label_img` vs. `OBJECT_ID_COLUMN`). I didn't initially include outliers because the EMT Migration project doesn't include the column, but I'll add them where they're used.

toloudis: The people preparing these scripts "should" be the ones who know what their data columns mean. It just felt like an extra level of indirection that one has to read.   But as long as it's consistent I think it's ok.

scripts/timelapse-colorizer-data/generate_data.py
ShrimpCryptid: Moved the remapping code and image scaling out of the writer class :)

scripts/timelapse-colorizer-data/generate_data.py
ShrimpCryptid: I think the performance difference might be pretty negligible? This part of the code runs pretty fast (1-2sec) compared to `make_frames`, which takes several hours.

README.md
ShrimpCryptid: The issue happens when you install the `pyproject.toml`, which means that the EMT Migration script is also affected.  I guess they could be moved into separate folders with their own dependency files...?

toloudis: In a way, the actual packageable library is just data_writer_utils now, which could have a `>=3.8` requirement, and then the convert scripts are just separate.

scripts/pyproject.toml
```diff
@@ -9,21 +9,21 @@ authors = [
 ]
 description = "data prep scripts"
 # readme = ""
-requires-python = ">=3.10"
+requires-python = ">=3.8, <3.9"
```
toloudis, line 12: probably this is a good place to have a comment that the nuc-morph-analysis dependency is only for convert_nucmorph_data and that's what's keeping this python version at 3.8

toloudis (approved):
🚀

scripts/timelapse-colorizer-data/convert_nucmorph_data.py
ShrimpCryptid, line 1: Previously `generate_data.py`! You can do a manual diff between the files to look at the changes, which mostly involved moving logic into `data_writer_utils.py`.

scripts/timelapse-colorizer-data/convert_emt_migration_data.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # Remap the frame image so the IDs are unique across the whole dataset.
+        seg_remapped, lut = remap_segmented_image(
+            seg2d,
+            frame,
+            OBJECT_ID_COLUMN,
+        )
+
+        writer.update_and_write_bbox_data(grouped_frames, seg_remapped, lut)
+        writer.write_image(seg_remapped, frame_number)
```
meganrm, line 69: is there a reason to have to call these successively? could you make one function in writer that that calls these both?

scripts/timelapse-colorizer-data/convert_emt_migration_data.py
```diff
@@ -0,0 +1,215 @@
...
+        c = a.loc[a["Image_Metadata_Plate"] == int(plate)]
+        c = c.loc[c["Image_Metadata_Position"] == int(position)]
+        make_dataset(c, output_dir, dataset, do_frames, scale)
+    else:
+        # for every combination of plate and position, make a dataset
+        b = a.groupby(["Image_Metadata_Plate", "Image_Metadata_Position"])
+        collection = []
+        for name, group in b:
+            dataset = str(name[0]) + "_" + str(name[1])
+            print(dataset)
```
meganrm, line 170: nit, but you seem to be using a logger other places, so replace this, or delete?

scripts/timelapse-colorizer-data/convert_emt_migration_data.py
```diff
@@ -0,0 +1,215 @@
...
+            collection.append({"name": dataset, "path": dataset})
+            c = a.loc[a["Image_Metadata_Plate"] == name[0]]
+            c = c.loc[c["Image_Metadata_Position"] == name[1]]
+            make_dataset(c, output_dir, dataset, do_frames, scale)
+        # write the collection.json file
+        with open(output_dir + "/collection.json", "w") as f:
+            json.dump(collection, f)
+
+
+parser = argparse.ArgumentParser()
```
meganrm, line 180: not for this PR, but we switched from using argparser to [fire](https://google.github.io/python-fire/guide/) and I like the simplicity of it

scripts/timelapse-colorizer-data/convert_emt_migration_data.py
```diff
@@ -0,0 +1,215 @@
...
+"""Column of frame number that the object ID appears in."""
+SEGMENTED_IMAGE_COLUMN = "OutputMask (CAAX)"
+"""Column of path to the segmented image data or z stack for the frame."""
+CENTROIDS_X_COLUMN = "R0Nuclei_AreaShape_Center_X"
+"""Column of X centroid coordinates, in pixels of original image data."""
+CENTROIDS_Y_COLUMN = "R0Nuclei_AreaShape_Center_Y"
+"""Column of Y centroid coordinates, in pixels of original image data."""
+
+
+def make_frames(grouped_frames, scale: float, writer: ColorizerDatasetWriter):
```
meganrm, line 37: type for `grouped_frames`?

scripts/timelapse-colorizer-data/convert_emt_migration_data.py
```diff
@@ -0,0 +1,215 @@
...
+
+        time_elapsed = time.time() - start_time
+        logging.info(
+            "Frame {} finished in {:5.2f} seconds.".format(
+                int(frame_number), time_elapsed
+            )
+        )
+
+
+def make_features(dataset: pd.DataFrame, features, writer: ColorizerDatasetWriter):
```
meganrm, line 79: type here too for `features`

meganrm (approved):

Issue #65, opened 2023-08-25
User controls for playback speed


toloudis:
## Use Case

For some data, playback is actually too fast.  User Derek has requested making it adjustable.

## Acceptance Criteria

UX

ShrimpCryptid:
- Requested the ability to slow down specifically for presentations, or trying to set a timing for how long it takes the sequence to complete.
- Graham and Jessica Yu both expressed interest in adding timing controls.
- Jess noted that it's a little fast for her too by default.
- Slider seems like it meets the use case for most folks

PR #113, opened 2023-10-13
feat: Playback speed controls

ShrimpCryptid:
Problem
=======
Closes https://github.com/allen-cell-animated/nucmorph-colorizer/issues/65!

Solution
========
- Refactors `TimeControls.tsx` to support passing in an FPS setting.
- Adds a new element, `PlaybackSpeedControl.tsx`, which displays speed as a multiplier (0.5x, 1x, 2x, etc.).

Note: The original default speed is now 10 FPS (was previously 25 FPS). The slider ranges from 0.25x to 2.5x speed, which means max speed is the same as the original.
(On my browser + internet, the maximum speed I could get was approximately 30 FPS, so keeping it to 25 FPS felt appropriate.)

## Type of change

Steps to Verify:
----------------
Please check "Hide whitespace"!
image

Screenshots (optional):
-----------------------
image

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/b0cc8ae8-7919-4340-aae3-99d98cc77d88


src/components/PlaybackSpeedControl.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+const defaultProps: Partial<PlaybackSpeedControlProps> = {
+  fps: 30,
+  onChange: () => {},
+  baselineFps: DEFAULT_PLAYBACK_FPS,
+  disabled: false,
+};
+
+export default function PlaybackSpeedControl(inputProps: PlaybackSpeedControlProps): ReactElement {
+  const props = { ...defaultProps, ...inputProps } as Required<PlaybackSpeedControlProps>;
```
frasercl, line 20: you could, if you wanted, change the type of `defaultProps` from `Partial<_>` to `Required<_>` and avoid the cast on line 20

src/colorizer/TimeControls.ts
```diff
@@ -8,23 +9,35 @@ export default class TimeControls {
...
   public setFrameCallback(fn: (frame: number) => void): void {
     this.setFrameFn = fn;
   }
 
+  public setPlaybackFps(fps: number): void {
+    this.playbackFps = fps;
+    if (this.timerId !== DEFAULT_TIMER_ID) {
+      // restart playback with new fps
+      clearInterval(this.timerId);
+      this.pauseCallbacks.every((callback) => callback());
```
frasercl, line 36: `forEach` would make more semantic sense here than `every` and avoid a potential bug.

`every` is for checking that every element of an array satisfies some condition: it expects the function it receives to return a boolean, and returns the result of running that function on every element of the array and `&&`ing all the results. (It has a counterpart, `some`, which `||`s the results together. Other languages call these methods `all` and `any`, respectively.) But like the `&&` operator, `every` also *short-circuits*: once any predicate returns a falsy value (including `undefined`), it halts. So this method is likely to only run the first callback in `pauseCallbacks` and then stop, because it knows at that point that not every callback in the list returns `true`.

(Looks like this is a pattern established previously in this file? Sorry for not catching this before!)

frasercl (approved):
Looks great! Left some comments, though only one requires addressing.

src/App.module.css
```diff
@@ -191,11 +191,12 @@
...
   max-width: calc(100vw - 60px);
   margin-top: 10px;
+  flex-wrap: wrap;
+  align-items: center;
 }
 
 .timeSliderContainer {
-  width: 100%;
-  max-width: 300px;
+  width: calc(min(50vw, 300px));
```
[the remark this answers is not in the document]
ShrimpCryptid: This was a quick fix for scaling behavior: the time slider would previously stay at the fixed 300px width, which meant that on REALLY narrow window dimensions it would extend off the screen. Because I was also changing the layout for the speed controls, I threw this in too.

src/components/PlaybackSpeedControl.tsx
ShrimpCryptid: I ended up just removing the two required props from the default ones!

src/colorizer/TimeControls.ts
ShrimpCryptid: Thanks for the in-depth explanation! This is fixed :')

toloudis (approved):

PR #114, opened 2023-10-16
feat: Added units and feature scaling to `convert_nucmorph_data.py`

ShrimpCryptid:
Problem
=======
The data units the nucmorph dataset generation script uses are in pixels and don't reflect real-world values (μm, μm<sup>3</sup>, etc.). 

Solution
========
- Uses the lookup table provided by `nuc_morph_analysis` to get scaling factors, human-readable labels, and units.
- Modifies the base utility scripts to accepts feature metadata (in this case, units).

The default datasets have not been updated with the new units yet, though I have the newly-generated feature and manifest files ready. (These will likely not replace those in the existing dataset until the UI is updated to handle units and decimals elegantly; see https://github.com/allen-cell-animated/nucmorph-colorizer/pull/120).

Here's a preview that includes the converted data:
https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-114/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ffull-resolution%2Fcollection.json&dataset=Baby%20Bear&feature=Volume&t=75

(Nicer version with the new units UI changes):
https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-120/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ffull-resolution%2Fcollection.json&dataset=Baby%20Bear&feature=Volume&t=75

## Type of change


scripts/timelapse-colorizer-data/data_writer_utils.py
```diff
@@ -161,6 +161,7 @@ def write_manifest(
         self,
         num_frames: int,
         feature_names: List[str],
+        feature_metadata: List[object] = [],
```
toloudis, line 164: At least, this should have stricter typing (enforcing `unit`) and at best, maybe the api should just accept a list of feature units as List[str] instead of asking users to structure it into some not-very-well-documented metadata object.   The utils code could take care of the structuring.

scripts/timelapse-colorizer-data/convert_nucmorph_data.py
```diff
@@ -168,13 +172,22 @@ def make_dataset(output_dir="./data/", dataset="baby_bear", do_frames=True, scal
     reduced_dataset[INITIAL_INDEX_COLUMN] = reduced_dataset.index.values
     grouped_frames = reduced_dataset.groupby(TIMES_COLUMN)
 
+    # Get the units and human-readable label for each feature; we include this as
+    # metadata inside the dataset manifest.
+    features = ["NUC_shape_volume_lcc", "NUC_position_depth"]
```
toloudis, line 177: I feel like this string list is now defined twice in the file and should be DRY factored up?

toloudis:
Any particular reason why not to update the other data prep script too?  That's the one where the units are in parens and the script could do a little magic to figure them out.

ShrimpCryptid:
> Any particular reason why not to update the other data prep script too? That's the one where the units are in parens and the script could do a little magic to figure them out.

Added! Updated the EMT dataset too, preview is here: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-120/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2FEMT%2Fcollection.json&dataset=3500005821_3&feature=Mean%20migration%20speed%20per%20track&track=100&t=30

You should be able to tell that it's different because there will be `μ` symbols where there used to be `u`!

scripts/timelapse-colorizer-data/data_writer_utils.py
ShrimpCryptid: Keeping this as a dictionary because I think this will keep things a bit tidier if the API updates in the future?

scripts/timelapse-colorizer-data/convert_nucmorph_data.py
```diff
@@ -168,13 +175,28 @@ def make_dataset(output_dir="./data/", dataset="baby_bear", do_frames=True, scal
...
+    # metadata inside the dataset manifest.
+    feature_labels = []
+    feature_metadata = []
+    formatted_units = {
+        "($\mu m$)": "µm",
+        "($\mu m^3$)": "µm³",
+        "($\mu m^3$/hr)": "µm³/hr",
+        "(min)": "min",
+        "($\mu m^{-1}$)": "µm⁻¹",
+    }
```
toloudis, line 188: why is this different from line 165 above?  I think this would be much much simpler if you hardcode the unit strings alongside the list of feature names.  You could have FEATURE_COLUMNS, FEATURE_NAMES, FEATURE_UNITS all together up at the top.

toloudis: Having said that, I get that it's nice to auto-generate the names and units from the original column names for this dataset.   I'm mostly motivated by wanting these scripts to be as simple as possible for varying levels of python users to be able to dive in and modify them.

toloudis (approved):

scripts/timelapse-colorizer-data/data_writer_utils.py
toloudis: It's a small point for now, but my take is: this is tidier for devs but not necessarily any easier for data entry.  my goal is the lowest possible friction for a scientist to throw their own script together.

scripts/timelapse-colorizer-data/convert_emt_migration_data.py
```diff
@@ -142,21 +154,23 @@ def make_dataset(
     reduced_dataset[INITIAL_INDEX_COLUMN] = reduced_dataset.index.values
     grouped_frames = reduced_dataset.groupby(TIMES_COLUMN)
 
+    # Get the units and human-readable label for each feature; we include this as
+    # metadata inside the dataset manifest.
+    feature_labels = []
+    feature_metadata: List[FeatureMetadata] = []
+    for i in range(len(FEATURE_COLUMNS)):
+        (label, unit) = extract_units_from_feature_name(FEATURE_COLUMNS[i])
```
ascibisz, line 162: nit: since we're just iterating through a list, this could just be
```
for feature in FEATURE_COLUMNS:
    (label, unit) = extract_units_from_feature_name(feature)
    ...
```
which feels slightly more pythonic to me

scripts/timelapse-colorizer-data/data_writer_utils.py
```diff
@@ -202,8 +228,21 @@ def write_manifest(
             "centroids": "centroids.json",
             "bounds": "bounds.json",
         }
+
+        # Merge the feature metadata together and include it in the output if present
+        if len(feature_metadata) != 0:
```
ascibisz, line 233: nit: this can just be `if feature_metadata:` since empty lists are false in python

scripts/timelapse-colorizer-data/convert_nucmorph_data.py
```diff
@@ -133,9 +139,10 @@ def make_features(
     centroids_y = dataset[CENTROIDS_Y_COLUMN].to_numpy()
 
     feature_data = []
-    for i in range(len(features)):
-        # TODO normalize output range excluding outliers?
-        f = dataset[features[i]].to_numpy()
+    for i in range(len(feature_names)):
```
ascibisz, line 142: same nit as above, can just do `for feature_name in feature_names:`

scripts/timelapse-colorizer-data/convert_nucmorph_data.py
```diff
@@ -168,13 +175,28 @@ def make_dataset(output_dir="./data/", dataset="baby_bear", do_frames=True, scal
...
+    feature_labels = []
+    feature_metadata = []
+    formatted_units = {
+        "($\mu m$)": "µm",
+        "($\mu m^3$)": "µm³",
+        "($\mu m^3$/hr)": "µm³/hr",
+        "(min)": "min",
+        "($\mu m^{-1}$)": "µm⁻¹",
+    }
+    for i in range(len(FEATURE_COLUMNS)):
```
ascibisz, line 189: same nit, `for feature in FEATURE_COLUMNS:`

ascibisz (approved):
I left a couple nits, but they're so mild I won't be offended if you ignore. Looks good to me!

PR #117, opened 2023-10-17
fix: Replace HTTP resource links with HTTPS

ShrimpCryptid:
Replaces the links to HTTP resources with those to HTTPS. This is both for security and to enable certain features like video export, which are only available in secure contexts.

Note:
- I purposefully am not including links to the HTTP version of Nucmorph Colorizer, because some functionality (like video export) doesn't work on it, but does if the project is installed locally.
- The github pages preview link will now successfully show our default datasets on VPN!

New README: https://github.com/allen-cell-animated/nucmorph-colorizer/tree/doc/use-https

### DIFF
--- a/src/components/LoadDatasetButton.tsx
+++ b/src/components/LoadDatasetButton.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useCallback, useContext, useEffect, useRef, useState } from "react";
+import React, { ReactElement, useCallback, useContext, useRef, useState } from "react";
 import { Button, Input, InputRef, Modal } from "antd";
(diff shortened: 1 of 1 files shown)


toloudis (approved):

ShrimpCryptid:
@toloudis re-requesting your review because I threw in a quick extra feature!

image

You can trigger it by loading this resource on the preview page: http://dev-aics-dtp-001.corp.alleninstitute.org/dan-data/colorizer/data/demo-collection.json

toloudis (approved):

frasercl (approved):

Issue #60, opened 2023-08-15
units


toloudis:
## Use Case

As a quantitative scientific tool, we need to have units on numeric measured features.

## Acceptance Criteria

Feature data should have units.  This needs to come from the data sources.

In the viewer, we should see units on all places that display feature values.  
In particular: the editable feature range values, the hovered feature value, and the plot Y axis.  (I may have missed something here).

ShrimpCryptid:
Here's a quick overview of all the places units should be shown.
![Image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/1d379afb-7a3d-4afa-813b-92e360e99444)

Written out:
- [ ] Min/max slider labels (but not input boxes)
- [ ] Hover tooltip
- [ ] Plot axis label
- [ ] Feature dropdown, in parentheses/italics after the unit name?

PR #120, opened 2023-10-20
feat: Units support and labeling

ShrimpCryptid:
Problem
=======
Closes https://github.com/allen-cell-animated/nucmorph-colorizer/issues/60!

Solution
========
There's two big themes in this change! The first is support for units, and the second is improvements to how numbers are displayed.

**Units:**
- Refactored `Dataset` to add support for two ways of getting units for a dataset. You can either:
  1. have the units be pulled automatically from the feature names that ended in parentheses, so `volume (in^3)` would become `volume` with units `in^3`, or...
  2. define the units in a new `featureMetadata` section in the manifest file, which overrides the previous behavior.
- Units are shown in the tooltip, feature dropdown, plot axes, and feature header.
- Added a new font (Carlito) as Lato does not support Greek characters.

**Numeric Display**
- Adds formatting for decimals (currently shows only three decimal places; approved by Saurabh).
- Slider only shows numbers as formatted decimals, but supports full-precision floats.

## Type of change

Steps to Verify:
----------------
Dataset where units are defined in the manifest file: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-120/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ffull-resolution%2Fcollection.json&dataset=Baby%20Bear&feature=Height&t=75

Dataset where units are part of the feature names: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-120/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2FEMT%2Fcollection.json&dataset=3500005820_1&feature=mean%20migration%20speed%20per%20track&t=31

Example dataset manifest with units included: https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/full-resolution/baby_bear/manifest.json

Video walkthrough (includes audio)
-----------------------

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/f62aa734-51a4-4c86-98f0-7d44d7776f8e

-----------------------


src/App.tsx
```diff
@@ -328,12 +329,16 @@ function App(): ReactElement {
...
-    (id: number): number => {
+    (id: number): string => {
       if (!featureName || !dataset) {
-        return -1;
+        return "";
       }
       // Look up feature value from id
-      return dataset.getFeatureData(featureName)?.data[id] || -1;
+      const featureData = dataset.getFeatureData(featureName);
+      const featureValue = featureData?.data[id] ?? -1;
```
toloudis, line 338: the the data value is 0 will this return -1?

toloudis: Minor/ nit: this whole block of code could actually go inside of dataset.  You already have `dataset.getFeatureNameWithUnits` - you could also have `dataset.getFeatureValueWithUnits(featureName, id)`

src/App.tsx
```diff
@@ -397,9 +402,21 @@ function App(): ReactElement {
...
+    // Add units to the dataset feature names if present
+    return dataset.featureNames.map((name) => {
+      return { key: name, label: dataset.getFeatureNameWithUnits(name) };
+    });
+  }, [dataset]);
+
   const disableUi: boolean = isRecording || !datasetOpen;
   const disableTimeControlsUi = disableUi;
 
+  const featureUnits = dataset?.featureHasUnits(featureName) ? dataset.getFeatureUnits(featureName) : "";
```
toloudis, line 418: Nit: getFeatureUnits could do the featureHasUnits check internally and return empty string as needed...?

src/App.tsx
```diff
@@ -494,7 +509,10 @@ function App(): ReactElement {
               tooltipContent={
                 <>
                   <p>Track ID: {lastHoveredId && dataset?.getTrackId(lastHoveredId)}</p>
-                  <p>Feature: {lastHoveredId && getFeatureValue(lastHoveredId)}</p>
+                  <p>
+                    {featureName}:{" "}
+                    <span style={{ whiteSpace: "nowrap" }}>{lastHoveredId && getFeatureValue(lastHoveredId)}</span>
```
toloudis, line 514: this is a nice touch

src/colorizer/Dataset.ts
```diff
@@ -25,6 +26,11 @@ export type FeatureData = {
   tex: Texture;
   min: number;
   max: number;
+  units?: string;
+};
+
+export type FeatureMetaData = {
```
toloudis, line 32: it looks weird to define this after the thing that uses it.  javascript is funny.

src/App.tsx
ShrimpCryptid: The `??` is a [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)! It's similar to using `||` but checks only for `null` or `undefined`, so 0 or other falsy values are safe. I'll add a comment about it though!

toloudis: my bad, thanks

src/App.tsx
ShrimpCryptid: Maybe horrid idea, what if someone includes empty quotes for their unit string? Should we allow that...?

toloudis: If they actually put a string consisting of two consecutive quotation marks (`"\"\""`), then we should print it as such.  But if they specify empty string, then that's the same as "no units provided".

src/colorizer/Dataset.ts
```diff
@@ -186,24 +211,53 @@ export default class Dataset {
...
+        // could be found.
+        metadata.units = metadataUnits;
+      } else {
+        // Strip the units from the feature name and save to metadata instead
+        newFeatureName = featureName.replace(detectedUnits[0], "").trim();
+        metadata.units = detectedUnits[1]; // Inner part (inside parentheses)
+      }
+
+      newFeaturesToFiles[newFeatureName] = this.featureFiles[featureName];
+      featuresToMetadata[newFeatureName] = metadata;
```
toloudis, line 248: most of the contents of the for loop could be its own function.  I'm not super excited about using these parens as a way to get units in.   I suggest that we disregard parentheses here and don't even try to parse for them, but in the data prep scripts we could do that parsing and populate featureMetadata that way.

toloudis: But since the code is already written, I think just making this code block its own function would be a nice cleanup/improvement.

toloudis: Possible side effect: someone might name their feature with a parenthetical suffix that was not meant to be units.  Python scripts could warn about this...

src/components/AppStyle.tsx
```diff
@@ -65,8 +65,10 @@ const theme = {
     },
   },
   font: {
-    family: "Lato",
-    resource: "https://fonts.googleapis.com/css2?family=Lato&display=swap",
+    // Lato does not provide Greek characters so Carlito (which is derived from Lato
+    // and provides Greek + Latin + Cyrillic) is used as a fallback.
+    family: "Lato, Carlito, sans-serif",
+    resource: "https://fonts.googleapis.com/css2?family=Lato&family=Carlito&display=swap",
```
toloudis, line 71: does this change affect the appearance of any text other than the Greek characters?

tests/Dataset.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    expect(dataset.featureHasUnits("feature2")).to.be.true;
+    // Should not strip out label if there is a unit override provided
+    expect(dataset.featureHasUnits("feature3 (μm/s(1))")).to.be.true;
+    expect(dataset.featureHasUnits("feature4 (unit)")).to.be.false;
+
+    expect(dataset.getFeatureUnits("feature1 (m)")).to.equal("meters");
+    expect(dataset.getFeatureUnits("feature2")).to.equal("sq. meters");
+    expect(dataset.getFeatureUnits("feature3 (μm/s(1))")).to.equal("");
+    expect(dataset.getFeatureUnits("feature4 (unit)")).to.be.undefined;
+  });
```
toloudis, line 120: good tests 👍

src/components/AppStyle.tsx
ShrimpCryptid: Yes, kinda. It won't affect standard text characters, as they'll still default to Lato. The font family attribute here acts as a series of fallbacks, where if a character is not found in a font the browser will try to use the next one, eventually going to some predefined default (I think it's usually Times New Roman).

Carlito provides a LOT more glyphs than Lato does, including groups other than Greek characters.
image
image

There *are* versions of Lato that do have complete character sets (https://www.latofonts.com/lato-free-fonts/#download) but we'd have to include that in the website files and it adds like 250 kB to the webpage download size. It also won't hook into the browser caching that Google Fonts provides. I specifically chose Carlito for this use case (mu) because it looks relatively consistent and it's easy to add in via Google Fonts, but if we start to see weird font deviances this might be the better option.

ShrimpCryptid: Okay wait actually I'm just going to use the Lato extended font I don't know why I didn't think about this earlier

src/colorizer/Dataset.ts
frasercl: Agree here - finding units in the formatting of name strings could be problematic, but with it already here I'd find it more readable as a separated and named function.

ShrimpCryptid: Ended up removing this since I made a related change at the data scripts layer! Thank you for the feedback.

public/Lato-Regular.ttf
toloudis, line 1: This is missing the woff, woff2 and eot files

src/components/AppStyle.tsx
```diff
@@ -88,6 +93,11 @@ export const AppThemeContext = createContext(theme);
 const CssContainer = styled.div`
   @import url("https://fonts.googleapis.com/css2?family=Lato&display=swap");
 
+  @font-face {
+    font-family: LatoExtended;
+    src: url(Lato-Regular.ttf);
```
toloudis, line 98: I think this needs the woff2,woff,eot in order of precedence.  Also see https://web.dev/articles/reduce-webfont-size#web_font_formats

toloudis (approved):
Apart from the font file inclusion this looks good to me!

src/components/AppStyle.tsx
ShrimpCryptid: Done, thank you!

frasercl (approved):
Looks good!

Issue #115, opened 2023-10-17
Remove plot controls + layout fixes


ShrimpCryptid:
## Use Case

Requested by @lynwilhelm and @vianamp!

- Remove controls on the plot
- Reduce vertical gap between plot and label
- Make plot heading left-aligned

![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/4b7cdb18-8b70-4fde-8b7f-b0e30222680d)

ShrimpCryptid:
@lynwilhelm how far left did you want the plot title?

ShrimpCryptid:
I tried fiddling with plotly and couldn't get it to remove the generous margins it uses on the plot layout. Need to keep looking into this one...

The plot controls have been removed in PR #123.

ShrimpCryptid:
Closing for now!

PR #123, opened 2023-10-26
feat: Feature Thresholding

ShrimpCryptid:
Problem
=======
Implements #91 and completes part of #115! 

Solution
========
Adds canvas support and UI for multi-feature thresholding. Updates the App layout to use a tabbed pane view of the plot and feature thresholds.

This is a big one! This touches code from the shaders to the UI, so there's lots of files included in this change.

**Canvas support for thresholding:**
- The main shader (`colorize_RGBA8U`) now takes in a mask texture of 0/1's, with a value for every object ID index. The texture is 1 if the object should be in range; otherwise it will either be colored with a solid color or hidden.
  - Also tweaks the color ramp min/max values so they only color and don't apply thresholding.
-  `ColorizeCanvas.tsx` can generate the mask texture from an array of `FeatureThreshold` objects, which give a min and max bound.
- `CanvasWrapper.tsx` uses React's `useTransition` hook to recalculate the mask texture-- this way it won't block UI updates!

**UI for thresholding:**
- New component, `FeatureThresholdPanel.tsx`!
  - Uses the Ant Select component for selecting features in the current dataset.
  - Each selected feature shows up as a slider with a min/max range.
  - Features can be removed via a button.
- Changing feature thresholding for the currently selected feature causes the color ramp to also update and snap to that range.

**App Layout:**
- The Plot and Filters content is now in a tabbed view that can be switched between.

**Misc Changes:**
- Adds a new button style in `IconButton` for minimalist buttons (like the remove/delete button in the new interface).
- Adds resizing support to the plot in `PlotWrapper` and `Plotting`, so that it plays nicely with page resizing.

There's also a number of future improvements I want to make that are currently captured in this issue: https://github.com/allen-cell-animated/nucmorph-colorizer/issues/124.

## Type of change

Steps to Verify:
----------------
https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-123/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2FEMT%2Fcollection.json

Screenshots (optional):
-----------------------
image

-----------------------
1. `ColorizeCanvas.tsx`
2. `FeatureThresholdPanel.tsx`
3. `App.tsx`

Thanks for contributing!


ShrimpCryptid:
Fixes from today's design discussion:
- [ ] Fix overflow on filters list
- [ ] Remove "Plot" label in Plot tab

src/colorizer/ColorizeCanvas.ts
```diff
@@ -367,9 +376,40 @@ export default class ColorizeCanvas {
...
+    // TODO: Optimize memory by using a true boolean array? Bit-level manipulation to fit it within Uint8Array?
+    const inRangeIds = new Uint8Array(this.dataset.numObjects);
+    inRangeIds.fill(1);
+
+    for (const threshold of thresholds) {
+      const featureData = this.dataset.getFeatureData(threshold.featureName);
+      if (!featureData) {
+        continue;
+      }
+      inRangeIds.forEach((value, i) => {
```
ShrimpCryptid, line 405: Replace with for

src/colorizer/Dataset.ts
```diff
@@ -182,6 +182,10 @@ export default class Dataset {
     return Object.keys(this.featureFiles);
   }
 
+  public get numObjects(): number {
+    return this.times?.length || this.trackIds?.length || this.features[this.featureNames[0]]?.data.length || 0;
```
toloudis, line 186: Is there one of these that you MUST have, to have a data set with items in it?  Just wondering why go through that chain.

src/components/CanvasWrapper.tsx
```diff
@@ -52,6 +55,9 @@ const defaultProps: Partial<CanvasWrapperProps> = {
 export default function CanvasWrapper(inputProps: CanvasWrapperProps): ReactElement {
   const props = { ...defaultProps, ...inputProps } as Required<CanvasWrapperProps>;
 
+  // Used to delay updates for operations like updating thresholding!
+  // Transitions are cancellable too, so if is interrupted by a new render
+  // the old work will be discarded.
```
toloudis, line 60: not sure what this comment refers to - is it in the right place?

src/components/DrawModeDropdown.tsx
```diff
@@ -45,7 +45,7 @@ export default function DrawModeSelector(propsInput: DrawModeSelectorProps): Rea
   const items = [
     { key: DrawMode.HIDE.toString(), label: "Hide values" },
     { key: DrawMode.USE_COLOR.toString(), label: "Use custom color" },
-    { key: DrawMode.USE_RAMP.toString(), label: "Use color map" },
+    // { key: DrawMode.USE_RAMP.toString(), label: "Use color map" },
```
toloudis, line 48: delete

src/components/PlotWrapper.tsx
```diff
@@ -48,5 +48,23 @@ export default function PlotWrapper(inputProps: PlotWrapperProps): ReactElement
...
+    }
+    const width = plotDivRef.current.clientWidth;
+    const height = plotDivRef.current.clientHeight;
+    console.log(`Resizing (${width}, ${height})`);
+    plot?.setSize(width, height);
+  };
+  useEffect(() => {
+    updatePlotSize();
+    // window.addEventListener("resize", updatePlotSize);
+    // return () => window.removeEventListener("resize", updatePlotSize);
```
toloudis, line 66: why do you want the resize handlers in there? is there something not working that would work better with them?

src/components/CanvasWrapper.tsx
ShrimpCryptid: Oops, this is a holdover from when I was using `useTransition`. I'll remove this.

src/App.tsx
```diff
@@ -60,15 +67,36 @@ function App(): ReactElement {
   const [colorRampMin, setColorRampMin] = useState(0);
   const [colorRampMax, setColorRampMax] = useState(0);
   const [outOfRangeDrawSettings, setoutOfRangeDrawSettings] = useState({
-    mode: DrawMode.USE_RAMP,
+    mode: DrawMode.USE_COLOR,
```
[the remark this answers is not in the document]
ShrimpCryptid: The only thing I can think about is if a user wants to ignore outliers? But I think from our conversations with scientists outliers are marked for a reason. I can strip out the handling code from the shader.

src/colorizer/Dataset.ts
ShrimpCryptid: Hmm, features is probably the safest, since having a dataset without a feature doesn't make sense?

src/components/PlotWrapper.tsx
ShrimpCryptid: Yeah, resizing of the plot can lag behind the window dimensions changing by a few seconds. Having window resize event listeners would let it respond instantly to window size changes, but it seems to get stuck at the maximum width when they're enabled.

toloudis: I suggest making a ticket for that if you haven't already. That way it will at least be documented for later attention.

toloudis (approved):
The resize thing is puzzling but overall LGTM!  This is a killer feature that our scientists will definitely use to explore their data.

src/components/PlotWrapper.tsx
ShrimpCryptid: Made the ticket: https://github.com/allen-cell-animated/nucmorph-colorizer/issues/131

src/components/LabeledRangeSlider.tsx
```diff
@@ -129,6 +139,14 @@ export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps):
   stepSize = clamp(stepSize, 0, 1);
   stepSize = setMaxDecimalPrecision(stepSize, 3);
 
+  let marks: undefined | Record<number, ReactNode> = undefined;
+  if (props.marks) {
+    marks = {};
+    props.marks.forEach((value) => {
+      marks![value] = <></>;
```
frasercl, line 146: Would recommend using `null` or `undefined` rather than a fragment here (both should be valid values of `ReactNode`)

src/components/FeatureThresholdPanel.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const selectedFeatures = props.featureThresholds.map((t) => t.featureName);
+  const featureOptions =
+    props.dataset?.featureNames.map((name) => ({ label: props.dataset?.getFeatureNameWithUnits(name), value: name })) ||
+    [];
+
+  const renderListItems = (item: FeatureThreshold, index: number): ReactNode => {
+    const featureData = props.dataset?.features[item.featureName];
+    if (!featureData) {
+      return <></>;
```
frasercl, line 99: can return `null` here too, I think

src/components/FeatureThresholdPanel.tsx
```diff
@@ -0,0 +1,154 @@
...
+      const existingThreshold = props.featureThresholds.find((t) => t.featureName === featureName);
+      if (existingThreshold) {
+        newThresholds.push(existingThreshold);
+      } else {
+        const featureData = props.dataset?.features[featureName];
+        if (featureData) {
+          newThresholds.push({ featureName, min: featureData.min, max: featureData.max });
+        }
+      }
+    });
```
frasercl, line 73: Suggestion: this `forEach` could become a `map` from `string[]` to `FeatureThreshold[]`.

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -45,6 +52,12 @@ uint getOutlierVal(int index) {
   return texelFetch(outlierData, featurePos, 0).r;
 }
 
+bool getInRange(int index) {
+  int width = textureSize(inRangeIds, 0).x;
+  ivec2 featurePos = ivec2(index % width, index / width);
+  return texelFetch(inRangeIds, featurePos, 0).r == 1u;
+}
+
```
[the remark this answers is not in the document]
toloudis: I think you can pass the texture as a function arg

src/colorizer/ColorizeCanvas.ts
```diff
@@ -367,9 +374,41 @@ export default class ColorizeCanvas {
...
+    inRangeIds.fill(1);
+
+    for (const threshold of thresholds) {
+      const featureData = this.dataset.getFeatureData(threshold.featureName);
+      if (!featureData) {
+        continue;
+      }
+      for (let i = 0, n = inRangeIds.length; i < n; i++) {
+        if (inRangeIds[i] === 1 && (featureData.data[i] < threshold.min || featureData.data[i] > threshold.max)) {
+          inRangeIds[i] = 0;
```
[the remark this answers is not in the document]
ShrimpCryptid: Yeah, it's a small optimization. I saw it in a discussion about `for` vs. `forEach`, but let me know if it impacts readability.

src/components/LabeledRangeSlider.tsx
ShrimpCryptid: I think Antd checks for null/undefined values here, and ignores them. In this case, we want to override Antd's normal labels for the mark values, and cause it to still render the marks but without any accompanying text. I'll update the comment to explain why it's using a fragment.

src/App.tsx
```diff
@@ -60,15 +67,36 @@ function App(): ReactElement {
...
 
+  const [featureThresholds, _setFeatureThresholds] = useState<FeatureThreshold[]>([]);
+  const setFeatureThresholds = useCallback(
+    // Change the current feature min + max on the color ramp if that feature's threshold moved.
+    (newThresholds: FeatureThreshold[]): void => {
+      // Check if the current feature name is being thresholded on, and if that threshold
+      // has changed. If so, snap the current min + max color ramp values so they match the new
+      // threshold values.
+      const currThreshold = featureThresholds.find((t) => t.featureName === featureName);
+      const newThreshold = newThresholds.find((t) => t.featureName === featureName);
```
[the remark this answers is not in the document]
ShrimpCryptid: Ah, so that's intended behavior. The thought there is that the color map range should by default show the range of what's being thresholded.

I updated the `updateFeature` method so it now sets the main range slider to match the threshold on initial load. Let me know if this makes sense?

src/components/FeatureThresholdPanel.tsx
```diff
@@ -0,0 +1,154 @@
...
+ */
+export default function FeatureThresholdPanel(inputProps: FeatureThresholdPanelProps): ReactElement {
+  const props = { ...defaultProps, ...inputProps } as Required<FeatureThresholdPanelProps>;
+
+  // Clear thresholds for features that don't exist when the dataset changes.
+  // TODO: Show these thresholds as disabled, rather than removing them.
+  useEffect(() => {
+    const newThresholds = props.featureThresholds.filter((t) => props.dataset?.featureNames.includes(t.featureName));
+    props.onChange(newThresholds);
+  }, [props.dataset]);
```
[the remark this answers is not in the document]
ShrimpCryptid: This gets removed in PR #128! I'm going to just remove this useEffect() block for now; thresholds for features that don't exist in the current dataset will just not show up in the thresholds list.

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -40,22 +40,16 @@ uint combineColor(uvec4 color) {
...
   ivec2 featurePos = ivec2(index % width, index / width);
-  return texelFetch(outlierData, featurePos, 0).r;
-}
-
-bool getInRange(int index) {
-  int width = textureSize(inRangeIds, 0).x;
-  ivec2 featurePos = ivec2(index % width, index / width);
-  return texelFetch(inRangeIds, featurePos, 0).r == 1u;
+  return texelFetch(tex, featurePos, 0);
 }
```
frasercl: suggestion: shrink names to `getUintFromTex` and `getFloatFromTex`

src/components/LabeledRangeSlider.tsx
frasercl: Gotcha, that makes sense!

frasercl (approved):
Looking good! Let's get this feature in!

Issue #82, opened 2023-08-30
Store selected color ramp in URL


ShrimpCryptid:
## Use Case

Include color ramps when sharing via URL.

## Acceptance Criteria

- Color ramp can be stored to and fetched from the URL parameters, e.g.:
`&colormap=Sequential_Forest`
- Fall back to a default if the name is incorrect.
- Optionally, add a way to invert color ramps (`&colormap=Sequential_Forest/inverted`)

## Details
- Originally part of https://github.com/allen-cell-animated/nucmorph-colorizer/issues/71 but spun off into a new issue ticket.

toloudis:
suggestion for inversion:  use a negative sign.  so `&colormap=Sequential_Forest` would be `&colormap=-Sequential_Forest` ! (Or could even suffix the negation instead of prefix.)

PR #127, opened 2023-10-28
feat: Color Map reversing, URL saving

### LINKED ISSUE #84: UI to reverse color ramp
## Use Case

Sometimes lighter colors make more intuitive sense for higher values (for example, when the depth feature is selected, and the tallest cells have the highest depth). There should be some way to reverse a currently selected color ramp.

## Acceptance Criteria

- Checkbox? Button? that, when clicked, reverses the direction of the currently selected color ramp. Depends on UX result

## Details
- Could use some design help from @lynwilhelm whenever we decide to prioritize this.
- Related to https://github.com/allen-cell-animated/nucmorph-colorizer/issues/82

ShrimpCryptid:
Problem
=======
Closes #84 and #82! 

Solution
========
Adds the UI and plumbing to support reversing the color map, as well as saving to and loading the color map from the URL (and whether it's reversed).

Note that if you switch color map, the new map will not be reversed (e.g. resets when swapping maps).

## Type of change

## Testing
- Open the link, change to a different color map, and then hit the reversal button next to the map.
- Copy the URL and paste it into a new tab. The same color map should be applied.

Screenshots (optional):
-----------------------
image

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -107,8 +107,10 @@ function App(): ReactElement {
       feature: featureName,
       track: selectedTrack?.trackId,
       time: currentFrame,
+      colorRamp: colorRampKey,
+      colorRampReversed: colorRampReversed,
     });
-  }, [collection, datasetKey, dataset, featureName, selectedTrack, currentFrame]);
+  }, [collection, datasetKey, dataset, featureName, selectedTrack, currentFrame, colorRampKey, colorRampReversed]);
 
   // Update url whenever the viewer settings change
   // (but not while playing/recording for performance reasons)
@@ -152,6 +154,17 @@ function App(): ReactElement {
   const initialUrlParams = useConstructor(() => {
     return urlUtils.loadParamsFromUrl();
   });
+
+  // Load URL parameters into the state that don't require a dataset to be loaded.
+  // This reduces flicker on initial load.
+  useEffect(() => {
+    if (initialUrlParams.colorRamp) {
+      setColorRampKey(initialUrlParams.colorRamp);
+    }
+    if (initialUrlParams.colorRampReversed) {
+      setColorRampReversed(initialUrlParams.colorRampReversed);
+    }
+  }, []);
 
   // Attempt to load database and collections data from the URL.
   // This is memoized so that it only runs one time on startup.
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -414,6 +414,10 @@ function App(): ReactElement {
   const disableUi: boolean = isRecording || !datasetOpen;
   const disableTimeControlsUi = disableUi;
 
+  let colorRamp = colorRampData.get(colorRampKey)?.colorRamp;
+  if (colorRampReversed && colorRamp) {
+    colorRamp = colorRamp.reverse();
+  }
   const featureUnits = dataset?.getFeatureUnits(featureName) || "";
 
   return (
@@ -530,7 +534,7 @@ function App(): ReactElement {
                 showTrackPath={showTrackPath}
                 outOfRangeDrawSettings={outOfRangeDrawSettings}
                 outlierDrawSettings={outlierDrawSettings}
-                colorRamp={colorRampData.get(colorRampKey)?.colorRamp!}
+                colorRamp={colorRamp!}
                 colorRampMin={colorRampMin}
                 colorRampMax={colorRampMax}
                 selectedTrack={selectedTrack}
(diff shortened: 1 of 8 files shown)


src/components/IconButton.tsx
ShrimpCryptid, line 1: Added new `link` styling and a way to override styles with a new `style` prop.

src/App.tsx
```diff
@@ -152,6 +155,17 @@ function App(): ReactElement {
     return urlUtils.loadParamsFromUrl();
   });
 
+  // Load URL parameters into the state that don't require a dataset to be loaded.
+  // This reduces flicker on initial load.
```
toloudis, line 159: reduces? 😄 honestly, I haven't really seen any flickering that was noticeable

ShrimpCryptid: It'll sometimes jump between a few temporary states on load! You'll see the canvas go grey and them sometimes jump from `t=0` to whatever time is in the URL.

src/colorizer/utils/url_utils.ts
```diff
@@ -164,11 +174,22 @@ export function loadParamsFromUrl(): UrlParams {
   // This assumes there are no negative timestamps in the dataset
   const timeParam = parseInt(urlParams.get(URL_PARAM_TIME) || "-1", base10Radix);
 
+  const colorRampRawParam = safeDecodeString(urlParams.get(URL_PARAM_COLOR_RAMP));
+  let colorRampParam: string | null = colorRampRawParam;
+  let colorRampReversedParam: boolean = false;
+  //  Color ramps are marked as reversed by adding ! to the end of the key
+  if (colorRampRawParam && colorRampRawParam.charAt(colorRampRawParam.length - 1) === "!") {
```
toloudis, line 181: Optional: I might be a tiny bit inclined to have a `const COLOR_RAMP_REVERSED_SUFFIX="!";` defined at the top of the file.

src/components/IconButton.tsx
```diff
@@ -36,6 +38,20 @@ const StyledButton = styled(Button)<{ $type: IconButtonProps["type"] }>`
...
+          color: var(--color-button);
+          fill: var(--color-button);
+
+          &:disabled {
+            border: 0;
+            background-color: transparent;
+            color: var(--color-text-disabled);
+            fill: var(--color-text-disabled);
+          }
+        `;
```
toloudis, line 54: not sure I get this.  What's it for?

src/App.tsx
toloudis: ah, I rarely am loading urls with nontrivial settings in them

src/components/IconButton.tsx
ShrimpCryptid: It's part of the style definition for the `IconButton`, which can have three types now, `primary | outlined | link`. The color ramp reversal button doesn't have a background or outline, and I thought it would be easier to extend `IconButton` since it already handles making icons a nice size.

The CSS here basically just disables the borders and background color so it's just colored text, and overrides the default behavior when the button is disabled so it doesn't add a background then either.

ShrimpCryptid: It doesn't, though it does handle disabled colors similarly. They differ when hovering or clicking.

Copy URL follows designs given in the Figma, while for the reversal button there wasn't an explicit design given for the hover/click behavior.

toloudis (approved):
LGTM!

src/components/IconButton.tsx
toloudis: just want to ensure we have consistent/predictable button behaviors.

src/App.tsx
```diff
@@ -413,6 +427,10 @@ function App(): ReactElement {
   const disableUi: boolean = isRecording || !datasetOpen;
   const disableTimeControlsUi = disableUi;
 
+  let colorRamp = colorRampData.get(colorRampKey)?.colorRamp;
+  if (colorRampReversed && colorRamp) {
+    colorRamp = colorRamp.reverse();
+  }
```
meganrm, line 433: I feel like more of this could be done by the ColorRamp class. Maybe a static function that takes in `colorRampKey` and `reversed` and does this logic in there

meganrm: it would also make this more testable. it looks like you have tests for the ui, but not for the ColorRamp class

ShrimpCryptid: Fixed, thank you!

meganrm (approved):

Issue #124, opened 2023-10-27
Feature thresholding: UI improvements


ShrimpCryptid:
Continued UI improvements on the Feature Thresholding Panel: 

- [x] Show disabled/enabled filters on the interface, and disable features that are not present in the current dataset instead of deleting them.
- [x] Copy feature thresholds to the URL + load them from the URL on startup
- [ ] Replace Ant Select with a Dropdown -> extend LabeledDropdown so it can support multi-selection
  - [x] Alternatively, troubleshoot Ant select layout issues
  - [x] Add Ant Select close all button? Change icons?

PR #128, opened 2023-10-29
feat: Disabling feature thresholds

ShrimpCryptid:
Problem
=======
Partially resolves action items in #124

Solution
========
- Allows thresholds on features to persist when switching datasets, even if the feature doesn't exist in the new dataset.
  - Features that are not available are greyed out/disabled.
- Various improvements to formatting on the Select component in `FeatureThresholdingPanel.tsx`
  - Fixed a spacing issue for selected feature names
  - Replaced dropdown and search icons to be consistent with the rest of the UI
- Various improvements to Slider formatting in `LabeledRangeSlider.tsx`
  - Greyed out min/max slider bound text labels when disabled. 
  - Removed tooltips when disabled.

## Type of change

Steps to Verify:
----------------
1. Open the preview link
2. Load this URL: `https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/`
3. Set a threshold on the feature `feature`.
4. Switch to another dataset (Test 5 or Test 6)
5. See that the original feature threshold is now disabled
6. Add additional feature thresholds.
7. Switch back.  All thresholds should persist

NEW: Verifying that thresholds are differentiated by feature + unit
1. Open the preview link. `Test 1 Default` should have a feature with a unit (`feature (test unit)`). If you don't see the unit label, you may need to clear your cache. 
3. Add a filter for that feature.
4. Switch to dataset `Test 2 Nested`. The original feature (`feature (test unit)`) should be greyed out.
5. You should now be able to add a new filter for the feature `feature`.

Demo Video (turn on sound 🔊)
-----------------------

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/867de740-8d79-42ec-8175-7c0daa43ea99

### DIFF
--- a/src/components/FeatureThresholdPanel.tsx
+++ b/src/components/FeatureThresholdPanel.tsx
@@ -192,7 +192,7 @@ export default function FeatureThresholdPanel(inputProps: FeatureThresholdPanelP
     // If the feature is no longer in the dataset, use the saved min/max bounds.
     const savedMinMax = featureMinMaxBoundsFallback.current.get(thresholdToKey(item)) || [0, 1];
     const sliderMin = disabled ? savedMinMax[0] : featureData.min;
-    const sliderMax = disabled ? savedMinMax[1] : featureData!.max;
+    const sliderMax = disabled ? savedMinMax[1] : featureData.max;
 
     const featureLabel = item.unit ? `${item.featureName} (${item.unit})` : item.featureName;
 
--- a/src/components/FeatureThresholdPanel.tsx
+++ b/src/components/FeatureThresholdPanel.tsx
@@ -117,7 +117,7 @@ export default function FeatureThresholdPanel(inputProps: FeatureThresholdPanelP
       } else {
         const featureData = props.dataset?.features[featureName];
         if (featureData) {
-          newThresholds.push({
+          newThresholds.unshift({
             featureName,
             min: featureData.min,
             max: featureData.max,
--- a/src/components/FeatureThresholdPanel.tsx
+++ b/src/components/FeatureThresholdPanel.tsx
@@ -9,6 +9,7 @@ import { FeatureThreshold } from "../colorizer/ColorizeCanvas";
 import LabeledRangeSlider from "./LabeledRangeSlider";
 import { Dataset } from "../colorizer";
 import IconButton from "./IconButton";
+import { thresholdMatchFinder } from "../colorizer/utils/data_utils";
 
 const PanelContainer = styled.div`
   flex-grow: 1;
@@ -139,9 +140,7 @@ export default function FeatureThresholdPanel(inputProps: FeatureThresholdPanelP
     const featureData = props.dataset?.features[featureName];
     const newThresholds = [...props.featureThresholds];
     if (featureData) {
-      const index = props.featureThresholds.findIndex(
-        (t) => t.featureName === featureName && t.unit === featureData.units
-      );
+      const index = props.featureThresholds.findIndex(thresholdMatchFinder(featureName, featureData.units));
       if (index !== -1) {
         // Delete saved min/max bounds for this feature
         const thresholdToRemove = props.featureThresholds[index];
--- a/src/components/FeatureThresholdPanel.tsx
+++ b/src/components/FeatureThresholdPanel.tsx
@@ -1,7 +1,7 @@
 import React, { ReactElement, ReactNode, useContext, useMemo, useRef } from "react";
 import { FeatureThreshold } from "../colorizer/ColorizeCanvas";
 import { Dataset } from "../colorizer";
-import { Card, List, Select, Tooltip } from "antd";
+import { Button, Card, List, Select, Tooltip } from "antd";
 import styled from "styled-components";
 import LabeledRangeSlider from "./LabeledRangeSlider";
 import { CloseOutlined, FilterOutlined } from "@ant-design/icons";
@@ -14,6 +14,17 @@ const PanelContainer = styled.div`
   flex-direction: column;
   gap: 6px;
   height: 100%;
+`;
+
+const SelectContainer = styled.div`
+  & .ant-select-selector {
+    padding: 0 2px;
+  }
+
+  & .ant-select-item-option-content {
+    font-weight: normal;
+    color: var(--color-button);
+  }
 `;
 
 const FiltersCard = styled(Card)`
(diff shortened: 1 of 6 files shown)


src/components/FeatureThresholdPanel.tsx
ShrimpCryptid, line 1: Changed max line length from 120 to 80 in this file, so some lines have been shifted as a result.

toloudis:
Just so I understand, this greyed-out thing only happens when I switch datasets within a collection?  And not if I just load a new dataset?

ShrimpCryptid:
> Just so I understand, this greyed-out thing only happens when I switch datasets within a collection? And not if I just load a new dataset?

This is great question! It will currently keep features if you switch to a totally new collection, in addition to keeping them between datasets within the same collection. I'm not actually sure what expected behavior would be, but I can add in a simple reset for switching between collections if that makes sense to you.

toloudis:
> > Just so I understand, this greyed-out thing only happens when I switch datasets within a collection? And not if I just load a new dataset?
> 
> This is great question! It will currently keep features if you switch to a totally new collection, in addition to keeping them between datasets within the same collection. I'm not actually sure what expected behavior would be, but I can add in a simple reset for switching between collections if that makes sense to you.

Yeah it seems like the list of features would just pile up. (Maybe the grey ones could be moved to the bottom). Any two datasets could have totally different feature sets.  It makes more sense if you only do it when switching between datasets in a collection, where it's much more likely to be consistent.

src/components/FeatureThresholdPanel.tsx
toloudis: this type of change probably should have its own PR where you just reformat everything rather than folding it in with real changes.

src/components/FeatureThresholdPanel.tsx
```diff
@@ -85,31 +141,46 @@ export default function FeatureThresholdPanel(inputProps: FeatureThresholdPanelP
...
+    featureMinMax.current.delete(props.featureThresholds[index].featureName);
     props.onChange(newThresholds);
   };
 
+  ////// RENDERING ///////////////////
+  // TODO: Possible bug where selected features with the same name but different units/scaling across
+  // datasets will use the same filter values.
+  // Might not be a problem if collections have similarly-structured datasets...?
+  // The alternative is to use the feature name + units as the key, but that may cause unexpected behavior
+  // for users (such as if one dataset does not have units for a feature but another does)
```
toloudis, line 154: suggestion: if feature name matches but units don't, then you just can't keep it

ShrimpCryptid: Okay, this has been done! It was a bit more involved than I thought it would be. I've updated the verification steps with instructions to test this.

src/components/FeatureThresholdPanel.tsx
```diff
@@ -47,30 +83,70 @@ const defaultProps: Partial<FeatureThresholdPanelProps> = {
...
+  const selectContainerRef = useRef<HTMLDivElement>(null);
+
+  /** Converts a threshold to a unique key that can be used to look up its information later. Matches on feature name and unit. */
+  const thresholdToKey = (threshold: FeatureThreshold): string => {
+    return `${encodeURIComponent(threshold.featureName)}:${threshold.unit ? encodeURIComponent(threshold.unit) : ""}`;
+  };
+  // Save the min/max values of each threshold in case the user switches to a dataset that no longer has the threshold's
+  // feature (no match for name + unit). This allows the sliders to be rendered with the same bounds as before until the user switches
+  // back to a dataset that has the thresholded feature.
+  const featureMinMax = useRef<Map<string, [number, number]>>(new Map());
```
toloudis, line 101: does this contain the whole list of what is visible in the panel, including the greyed-out ones?  (it seems like this info is already saved in featureThresholds)

ShrimpCryptid: Yes! This holds the absolute min/max range of the feature (essentially, the range shown on the slider) and not the min/max of the threshold. This prevents the slider bounds from jumping around as different datasets are loaded.

I'll update the comment, but let me know if it's still unclear.

toloudis: makes sense, thanks!

src/components/FeatureThresholdPanel.tsx
```diff
@@ -81,35 +158,55 @@ export default function FeatureThresholdPanel(inputProps: FeatureThresholdPanelP
...
+    // both in the current dataset to be valid.
     const featureData = props.dataset?.features[item.featureName];
-    if (!featureData) {
-      return <></>;
-    }
+    const disabled = featureData === undefined || featureData.units !== item.unit;
+    // If the feature is no longer in the dataset, use the saved min/max bounds.
+    const savedMinMax = featureMinMax.current.get(thresholdToKey(item)) || [0, 1];
+    const sliderMin = disabled ? savedMinMax[0] : featureData.min;
+    const sliderMax = disabled ? savedMinMax[1] : featureData!.max;
```
toloudis, line 195: should both of these lines use `featuredata!` ?

ShrimpCryptid: Oops, my mistake! Typescript is smart enough to know that `disabled` checks for the existence of `featureData`, so the `!` isn't needed anymore.

toloudis (approved):
👍

src/components/FeatureThresholdPanel.tsx
```diff
@@ -47,23 +84,70 @@ const defaultProps: Partial<FeatureThresholdPanelProps> = {
...
+  // slider.
+  // Doing this allows the sliders be rendered with the same bounds as before and prevents weird UI behavior.
+  const featureMinMaxBoundsFallback = useRef<Map<string, [number, number]>>(new Map());
+
+  useMemo(() => {
+    // Update the saved min/max bounds of any selected features.
+    // This is done as a useMemo and not when the feature is added, in case we switch to another dataset
+    // that has the same feature/unit but different min/max values. That way, our saved min/max bounds
+    // reflect the last known good values.
+    for (const threshold of props.featureThresholds) {
```
frasercl, line 111: These comments helped a lot with understanding what's going on here - thanks!

frasercl, line 111: These comments helped a lot with understanding what's going on here - thanks!

frasercl (approved):
:thumbsup:

PR #129, opened 2023-10-29
feat: Loading feature thresholds from URL

### LINKED ISSUE #124: Feature thresholding: UI improvements

ShrimpCryptid:
Problem
=======
Closes #124!

Solution
========
- Feature thresholds and current color ramp min+max are now saved to the URL if edited from the defaults.
- The app will now load the color ramp and thresholds from the URL on startup!
- Also refactors tests and other URL handling code to reduce assumptions about formatting in the utility scripts.

## Type of change

Steps to Verify:
----------------
1. Open the preview link.
2. Set the feature color map range. You should see the URL update.
3. Set feature thresholds in the filters pane. You should also see the URL update.
4. Copy the URL and paste it into a new tab or browser window.
5. The thresholds + color map range should persist.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,13 +40,7 @@ import LoadDatasetButton from "./components/LoadDatasetButton";
 import PlaybackSpeedControl from "./components/PlaybackSpeedControl";
 import PlotWrapper from "./components/PlotWrapper";
 import SpinBox from "./components/SpinBox";
-import {
-  DEFAULT_COLLECTION_FILENAME,
-  DEFAULT_COLLECTION_PATH,
-  DEFAULT_COLOR_RAMPS,
-  DEFAULT_COLOR_RAMP_ID,
-  DEFAULT_PLAYBACK_FPS,
-} from "./constants";
+import { DEFAULT_COLLECTION_PATH, DEFAULT_COLOR_RAMPS, DEFAULT_COLOR_RAMP_ID, DEFAULT_PLAYBACK_FPS } from "./constants";
 import FeatureThresholdPanel from "./components/FeatureThresholdPanel";
 
 function App(): ReactElement {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -134,8 +134,9 @@ function App(): ReactElement {
       feature: featureName,
       track: selectedTrack?.trackId,
       time: currentFrame,
+      thresholds: featureThresholds,
     });
-  }, [collection, datasetKey, dataset, featureName, selectedTrack, currentFrame]);
+  }, [collection, datasetKey, dataset, featureName, selectedTrack, currentFrame, featureThresholds]);
 
   // Update url whenever the viewer settings change
   // (but not while playing/recording for performance reasons)
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,13 +16,8 @@ import { Color } from "three";
 import styles from "./App.module.css";
 import { ColorizeCanvas, Dataset, Track } from "./colorizer";
 import Collection from "./colorizer/Collection";
-import {
-  BACKGROUND_ID,
-  DrawMode,
-  FeatureThreshold,
-  OUTLIER_COLOR_DEFAULT,
-  OUT_OF_RANGE_COLOR_DEFAULT,
-} from "./colorizer/ColorizeCanvas";
+import { BACKGROUND_ID, DrawMode, OUTLIER_COLOR_DEFAULT, OUT_OF_RANGE_COLOR_DEFAULT } from "./colorizer/ColorizeCanvas";
+import { FeatureThreshold } from "./constants/types";
 import TimeControls from "./colorizer/TimeControls";
 import { numberToStringDecimal } from "./colorizer/utils/math_utils";
 import { useConstructor, useDebounce } from "./colorizer/utils/react_utils";
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -137,14 +137,6 @@ function App(): ReactElement {
       datasetParam = dataset?.manifestUrl;
       collectionParam = undefined;
     }
-    // If the collection is the default collection path, don't include it in the URL.
-    if (
-      collectionParam &&
-      (collectionParam === DEFAULT_COLLECTION_PATH ||
-        collectionParam === DEFAULT_COLLECTION_PATH + "/" + DEFAULT_COLLECTION_FILENAME)
-    ) {
-      collectionParam = undefined;
-    }
     // check if current color map range matches the default feature range; if so, don't include
     // it in the URL parameter to reduce clutter.
     let rangeIsDefault = false;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -269,7 +269,7 @@ function App(): ReactElement {
       }
       if (initialUrlParams.track && initialUrlParams.track >= 0) {
         // Highlight the track. Seek to start of frame only if time is not defined.
-        findTrack(initialUrlParams.track, (initialUrlParams.time ?? 0) < 0);
+        findTrack(initialUrlParams.track, initialUrlParams.time !== undefined);
       }
       let newTime = currentFrame;
       if (initialUrlParams.time && initialUrlParams.time >= 0) {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -116,27 +116,47 @@ function App(): ReactElement {
   // UTILITY METHODS /////////////////////////////////////////////////////////////
 
   /**
+   * Formats the dataset and collection parameters for use in a URL.
+   */
+  const getDatasetAndCollectionParam = useCallback((): {
+    datasetParam?: string;
+    collectionParam?: string;
+  } => {
+    if (collection?.url === null) {
+      // A single dataset was loaded, so there's no collection URL. Use the dataset URL instead.
+      return { datasetParam: dataset?.manifestUrl, collectionParam: undefined };
+    } else {
+      return { datasetParam: datasetKey, collectionParam: collection?.url };
+    }
+  }, [datasetKey, dataset, collection]);
+
+  /**
+   * Get the optional color map feature range parameter for the URL. If the range
+   * is the full range of the feature's values (default), return undefined.
+   */
+  const getRangeParam = useCallback((): [number, number] | undefined => {
+    if (!dataset) {
+      return undefined;
+    }
+    // check if current selected feature range matches the default feature range; if so, don't provide
+    // a range parameter..
+    const featureData = dataset.getFeatureData(featureName);
+    if (featureData) {
+      if (featureData.min === colorRampMin && featureData.max === colorRampMax) {
+        return undefined;
+      }
+    }
+    return [colorRampMin, colorRampMax];
+  }, [colorRampMin, colorRampMax, featureName, dataset]);
+
+  /**
    * Get a URL query string representing the current collection, dataset, feature, track,
    * and frame information.
    */
   const getUrlParams = useCallback((): string => {
-    if (!dataset) {
-      return "";
-    }
-    let datasetParam: string | undefined = datasetKey;
-    let collectionParam = collection?.url;
-    // If there is no collection loaded (e.g. a single dataset URL was loaded),
-    // use the dataset URL instead.
-    if (collectionParam === null) {
-      datasetParam = dataset?.manifestUrl;
-      collectionParam = undefined;
-    }
-    // check if current selected feature range matches the default feature range; if so, don't include.
-    let rangeIsDefault = false;
-    const featureData = dataset.getFeatureData(featureName);
-    if (featureData) {
-      rangeIsDefault = featureData.min === colorRampMin && featureData.max === colorRampMax;
-    }
+    const { datasetParam, collectionParam } = getDatasetAndCollectionParam();
+    const rangeParam = getRangeParam();
+
     return urlUtils.paramsToUrlQueryString({
       collection: collectionParam,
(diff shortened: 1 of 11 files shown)


src/colorizer/utils/url_utils.ts
ShrimpCryptid, line 1: I think I wrote this code before I knew about the `Partial` and `Required` tags in Typescript, so I removed optional null values and used `Partial` instead to support undefined values.

src/colorizer/utils/url_utils.ts
```diff
@@ -53,40 +61,57 @@ export function fetchWithTimeout(
...
-    includedParameters.push(`${URL_PARAM_TRACK}=${track}`);
+  if (state.track !== undefined) {
+    includedParameters.push(`${URL_PARAM_TRACK}=${state.track}`);
   }
-  if (time && time > 0) {
-    // time = 0 is ignored because it's the default frame.
-    includedParameters.push(`${URL_PARAM_TIME}=${time}`);
+  if (state.time !== undefined) {
+    includedParameters.push(`${URL_PARAM_TIME}=${state.time}`);
+  }
```
ShrimpCryptid, line 98: Made the code less aware of requirements here; it makes no assumptions about what values are wanted/unwanted.

src/colorizer/utils/url_utils.ts
```diff
@@ -2,19 +2,27 @@
...
-  collection: string | null;
-  dataset: string | null;
-  feature: string | null;
+  collection: string;
+  dataset: string;
+  feature: string;
   track: number;
   time: number;
+  // TODO: bad code smell for url params to be aware of this type
+  thresholds: FeatureThreshold[];
```
ShrimpCryptid, line 24: Is it bad code smell for url_utils.ts to be aware of this type?
Maybe this script shouldn't live in utils?

src/colorizer/utils/url_utils.ts
```diff
@@ -1,20 +1,26 @@
...
+import { FeatureThreshold } from "../ColorizeCanvas";
+import { numberToStringDecimal } from "./math_utils";
 
 const URL_PARAM_TRACK = "track";
 const URL_PARAM_DATASET = "dataset";
 const URL_PARAM_FEATURE = "feature";
 const URL_PARAM_TIME = "t";
 const URL_PARAM_COLLECTION = "collection";
+// TODO: Make thresholds/filters language consistent. Requires talking with users/UX!
+const URL_PARAM_THRESHOLDS = "thresholds";
```
ShrimpCryptid, line 13: Outstanding question of whether thresholds of filters is more familiar for our users. Will ask Lyndsay and possibly update this PR.

src/colorizer/utils/url_utils.ts
```diff
@@ -53,40 +59,55 @@ export function fetchWithTimeout(
...
+    includedParameters.push(`${URL_PARAM_DATASET}=${encodeURIComponent(state.dataset)}`);
   }
-  if (feature) {
-    includedParameters.push(`${URL_PARAM_FEATURE}=${encodeURIComponent(feature)}`);
+  if (state.feature) {
+    includedParameters.push(`${URL_PARAM_FEATURE}=${encodeURIComponent(state.feature)}`);
   }
-  if (track && track >= 0) {
-    includedParameters.push(`${URL_PARAM_TRACK}=${track}`);
+  if (state.track !== undefined) {
```
meganrm, line 87: just curious, is this change because tracks can be negative?

src/colorizer/utils/url_utils.ts
meganrm: it doesn't seem terrible to me, but this also just feels like a subset of your overall state (guessing since I'm not looking at the rest of the code), so if you wanted to this whole thing could be defined where you're handling your overall app state and then imported here. Otherwise I think it can be nice to have a `types` file for shared types instead of having circular imports.

src/colorizer/utils/url_utils.ts
ShrimpCryptid: Tracks can't be negative... this change was more to remove some of the assumptions `url_utils` was making about the data format (making it a little simpler) and rely on the top-level App to handle that.

tests/url_utils.test.ts
```diff
@@ -60,20 +50,74 @@ describe("stateToUrlParamString", () => {
...
+    const originalParams: Required<UrlParams> = {
       collection: "collection",
       dataset: "dataset",
       feature: "feature",
       track: 25,
       time: 14,
-    });
-    expect(result).to.equal("?collection=collection&dataset=dataset&feature=feature&track=25&t=14");
+      thresholds: [
+        { featureName: "f1", min: 0, max: 0 },
```
meganrm, line 84: is there a check somewhere that `max `is larger than `min`?

ShrimpCryptid: The UI handles the case where the values are reversed, yes, but it's not directly handled in `url_utils`. I do think it makes sense for it to enforce min/max, though, so I'll make that change and add a test for it.

meganrm (approved):
Looks good

src/App.tsx
```diff
@@ -119,23 +125,54 @@ function App(): ReactElement {
...
-    if (!collection?.url) {
-      // The collection has no source file; use the dataset URL instead
+    let collectionParam = collection?.url;
+    // If there is no collection loaded (e.g. a single dataset URL was loaded),
+    // use the dataset URL instead.
+    if (collectionParam === null) {
       datasetParam = dataset?.manifestUrl;
+      collectionParam = undefined;
+    }
+    // If the collection is the default collection path, don't include it in the URL.
```
toloudis, line 140: I think this should just always be a full url.  There isn't really a default collection anymore, is there?

src/colorizer/utils/url_utils.ts
toloudis: since we use filter in the rest of the ui, it will be more consistent and understandable to use it in the url too.

src/colorizer/utils/url_utils.ts
```diff
@@ -53,40 +59,55 @@ export function fetchWithTimeout(
...
-  const { collection, dataset, feature, track, time } = state;
 
   // Don't include collection parameter in URL if it matches the default.
-  if (
-    collection &&
-    collection !== DEFAULT_COLLECTION_PATH &&
-    collection !== DEFAULT_COLLECTION_PATH + "/" + DEFAULT_COLLECTION_FILENAME
-  ) {
-    includedParameters.push(`${URL_PARAM_COLLECTION}=${encodeURIComponent(collection)}`);
+  if (state.collection) {
```
toloudis, line 78: The comment above here is obsolete now?

toloudis (approved):

src/colorizer/utils/url_utils.ts
frasercl: Agreed that it could be good to put this in the `types` file - it's used across lots of files, and it doesn't seem like any single one has a clear claim to being the "owner" of this type.

frasercl:
Looks good!

ShrimpCryptid:
Re-requesting @toloudis and @meganrm (optionally @frasercl ) because I added support for threshold units.

src/colorizer/utils/url_utils.ts
```diff
@@ -53,40 +60,58 @@ export function fetchWithTimeout(
...
+    // featureName is encoded in case it contains special characters (":" or ",")
+    // TODO: Are there better characters I can be using here? ":" and "," take up
+    // more space in the URL.
+    const thresholdsString = state.thresholds
+      .map((threshold) => {
+        const featureName = encodeURIComponent(threshold.featureName);
+        // Note that THRESHOLD_UNIT_UNDEFINED (="!") is not encoded here, so that it won't conflict
+        // with normal feature units names. (Users should not be using "!" as a unit anyway, but just in case.)
+        const featureUnit =
+          threshold.unit !== undefined ? encodeURIComponent(threshold.unit) : THRESHOLD_UNIT_UNDEFINED;
```
toloudis, line 104: Would empty string have worked for THRESHOLD_UNIT_UNDEFINED?  `name::min:max` instead of `name:!:min:max` ?  Just wondering.
Also what if threshold.unit === "" ? Then that is what you'd get anyway.

ShrimpCryptid: Currently empty strings are considered different from a feature unit being `undefined`? `threshold.unit` is either `undefined` or a string type.

If `THRESHOLD_UNIT_UNDEFINED = ""`, unit values of `undefined` and `""` would BOTH map to the same value, which would make the URL saving/parsing lossy. For now, `undefined` and `""` values mean different things in the interface, though I guess there could be a discussion about whether we should allow `""` as a valid unit type anyways.

Let me know if that makes sense?

toloudis: The real question is: what's the difference between undefined and ""?  Don't they both mean we have a unitless feature?

ShrimpCryptid: Hmm, okay, I started digging into the code for unit handling again. From what I see there isn't a situation where `undefined` is handled differently from `""` other than in tests. Does it make sense to just replace `undefined` values with an empty string on Dataset load, then?

ShrimpCryptid: I'm going to just make that change and update the relevant tests, I think.

toloudis: My hope was that it would be simpler to handle in that way.

src/colorizer/utils/data_utils.ts
```diff
@@ -13,7 +13,7 @@ import { FeatureThreshold } from "../ColorizeCanvas";
  */
 export function thresholdMatchFinder(
   featureName: string,
-  unit: string | undefined
+  units: string | undefined
```
toloudis, line 16: does this one still need to be `string | undefined` ?

toloudis (approved):
LGTM!

src/App.tsx
```diff
@@ -121,23 +121,46 @@ function App(): ReactElement {
...
+    // use the dataset URL instead.
+    if (collectionParam === null) {
       datasetParam = dataset?.manifestUrl;
+      collectionParam = undefined;
     }
-    return urlUtils.stateToUrlParamString({
-      collection: collection?.url,
+    // check if current color map range matches the default feature range; if so, don't include
+    // it in the URL parameter to reduce clutter.
+    let rangeIsDefault = false;
```
meganrm, line 138: seems like this could be a function on colorRamp class. Ie `rangeIsDefault = colorRamp.getRangeIsDefault(featureData, colorRampMin, colorRampMin`.

Also, could you just do a `if !dataset` return? at the top of this? hopefully to remove the multiple conditionals here.

src/App.tsx
```diff
@@ -121,23 +121,46 @@ function App(): ReactElement {
...
     let datasetParam: string | undefined = datasetKey;
-    if (!collection?.url) {
-      // The collection has no source file; use the dataset URL instead
+    let collectionParam = collection?.url;
+    // If there is no collection loaded (e.g. a single dataset URL was loaded),
+    // use the dataset URL instead.
+    if (collectionParam === null) {
       datasetParam = dataset?.manifestUrl;
+      collectionParam = undefined;
     }
```
meganrm, line 135: could this be moved into `urlUtils.stateToUrlQueryString` ?

meganrm: or if you want to keep that as something that cleanly only takes in a `state` parameter, a new function on `urlUtils` called `convertParamsToState` that returns the state object

src/App.tsx
ShrimpCryptid: Ah, I think the comment here I had was a bit misleading. The range here isn't governed by the ColorRamp class, but is a min and max feature value for where the endpoints of the gradient (held by ColorRamp) should be applied. I've updated the comment accordingly.

src/App.tsx
ShrimpCryptid: Hmm, I don't think there's a clean way to move this into `urlUtils` without introducing a lot of fragile boilerplate, since so much of the decisionmaking here is dependent on the way state is stored/represented in App...

ex: if I added a `convertParamsToState(rawParams: ParamType)` to `url_utils` I'd have to add the `ParamType` with an entry for every state variable in App.

ShrimpCryptid: Made some changes, let me know if that looks better?

src/App.tsx
```diff
@@ -121,23 +116,58 @@ function App(): ReactElement {
...
       feature: featureName,
       track: selectedTrack?.trackId,
-      time: currentFrame,
+      // Ignore time=0 to reduce clutter
+      time: currentFrame !== 0 ? currentFrame : undefined,
+      thresholds: featureThresholds,
+      range: rangeParam,
     });
-  }, [collection, datasetKey, dataset, featureName, selectedTrack, currentFrame]);
+  }, [getDatasetAndCollectionParam, getRangeParam, featureName, selectedTrack, currentFrame, featureThresholds]);
```
meganrm, line 170: yeah, thanks, this is more readable to me

meganrm (approved):
looks good, I appreciate all the tests too

Issue #135, opened 2023-11-03
Include UI overlays (scale bar, time, etc.) in exported frames and video.


ShrimpCryptid:
## Use Case

If scientists want to show off exported videos, scale bars and time stamps provide important scientific context for audiences.

## Acceptance Criteria

- Scale bar and time stamps appear in exported video.

## Details

- `ColorizeCanvas.ts` will most likely need to render the UI overlay to an OffscreenCanvas and then copy to the main canvas.

ShrimpCryptid:
Optional: Also add difference blending modes when combining the canvases, so that the scale bar still stands out against dark backgrounds. See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation

PR #136, opened 2023-11-03
feat: Scale Bar

### LINKED ISSUE #56: Scale Bar
## Use Case

Show a scale bar on the canvas. Requested by @mogres!

## Acceptance Criteria

TBD, may need help from UX in the future (@lynwilhelm)

## Details
- May be similar to volume-viewer's solution
- Should be done in the dataset, during data generation.
- Relative to pixels or whole image?
- Overlay on top of viewport?

ShrimpCryptid:
Problem
=======
Closes #56!

This update adds a scale bar to the main viewport and updates the dataset format to include metadata on the image frame size and units to be used. The scale bar resizes automatically and uses scientific notation to handle very large (and very small) numbers.

Solution
========
- Adds a new `CanvasOverlay.ts` for rendering UI overlays.
  - `ColorizeCanvas.ts` instantiates the overlay and renders it in the same location, and is responsible for updating it.
  - `CanvasOverlay` is a 2D Canvas-- I tried using a `CSS2DRenderer` and plain HTML, but there are various issues with each approach. The main limiting factor is the need to support including these overlays in video export in the future (see #135).
- Adds utilities for converting numbers to scientific notation, plus associated unit tests.

## Type of change

Videos (🔉 Enable Audio!)
-----------------------

**Scale Bar:**

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/40827a1a-5b44-4053-827e-48ba2cd2901c

**Visual Example of what the scale bar looks like at different sizes:**

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/104f6441-c1a2-4e68-975a-4a49946c6a0f


src/colorizer/ColorizeCanvas.ts
```diff
@@ -178,12 +185,36 @@ export default class ColorizeCanvas {
...
+      this.overlay.domElement.style.left = "0";
+      this.overlay.domElement.style.top = "0";
+    }
+
+    return this.canvasContainer;
+  }
+
+  get canvasElement(): HTMLCanvasElement {
     return this.renderer.domElement;
   }
```
frasercl, line 219: An odd distinction to make, though I see why you have to. Though is the plan to eventually render `CanvasOverlay` offscreen and draw it on top of the original canvas? If so, would these two properties remain?

src/colorizer/ColorizeCanvas.ts
```diff
@@ -178,12 +185,36 @@ export default class ColorizeCanvas {
...
+      this.canvasContainer.appendChild(this.renderer.domElement);
+      this.canvasContainer.appendChild(this.overlay.domElement);
+      this.canvasContainer.style.position = "relative";
+      this.overlay.domElement.style.position = "absolute";
+      this.overlay.domElement.style.left = "0";
+      this.overlay.domElement.style.top = "0";
+    }
+
+    return this.canvasContainer;
+  }
```
frasercl, line 215: Doing all this element-building in the getter could open up surprising results later. E.g. the client could add `canvasElement` to the document then later added `domElement` in a different place, thereby lifting the canvas out of its original context. I'd suggest putting this in the constructor? If it's important to have the option to use either the bare canvas or this wrapper, this could be a method, but I think it makes more sense to let the client treat the canvas as a single drop-in element that contains and abstracts everything.

frasercl (commented):
This is very interesting to look over after writing the scale bar for volume-viewer. Scale bars involve a surprising number of considerations! Yours is much more customizable, handles scientific notation and value increments better, and renders to a canvas rather than a DOM element (essential for including it in video exports later). The only features I can think of that are unique in volume-viewer is the option to render to different viewport corners (unused in practice, and likely to be unused here too) and drawing in white with `mix-blend-mode: difference` to ensure the scale bar stands out against many background colors, which might be worth considering as backgrounds become customizable here as well.

ShrimpCryptid:
> This is very interesting to look over after writing the scale bar for volume-viewer. Scale bars involve a surprising number of considerations! Yours is much more customizable, handles scientific notation and value increments better, and renders to a canvas rather than a DOM element (essential for including it in video exports later). The only features I can think of that are unique in volume-viewer is the option to render to different viewport corners (unused in practice, and likely to be unused here too) and drawing in white with `mix-blend-mode: difference` to ensure the scale bar stands out against many background colors, which might be worth considering as backgrounds become customizable here as well.

The difference blending mode is a really good idea, wow! I'm going to add a note about this to #135 when I start merging rendered canvases.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Makes sense, fixed!

ShrimpCryptid: Makes sense, fixed!

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Yeah, ideally `CanvasOverlay` can be rendered with `OffscreenCanvas` and the `domElement` getter could be removed.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -178,12 +185,34 @@ export default class ColorizeCanvas {
...
+
+  /**
+   * The DOM element containing the canvas and its overlay.
+   */
+  get domElement(): HTMLDivElement {
+    return this.canvasContainer;
   }
 
-  get domElement(): HTMLCanvasElement {
+  get canvasElement(): HTMLCanvasElement {
```
toloudis, line 215: nit: why use getters instead of a plain function?

src/colorizer/Dataset.ts
```diff
@@ -223,6 +234,7 @@ export default class Dataset {
     this.frameFiles = manifest.frames;
     this.featureFiles = manifest.features;
     this.outlierFile = manifest.outliers;
+    this.metadata = manifest.metadata;
```
toloudis, line 237: Nit/Idea: I wonder if it would be nicer to just not use Partials here and instead call a function that turns the necessarily partial `manifest.metadata` into a complete `this.metadata`.

src/colorizer/utils/math_utils.ts
```diff
@@ -24,3 +22,47 @@ export function numberToStringDecimal(
 export function setMaxDecimalPrecision(input: number, decimalPlaces: number): number {
   return Number.parseFloat(numberToStringDecimal(input, decimalPlaces, true));
 }
+
+// Adapted from https://gist.github.com/ArneS/2ecfbe4a9d7072ac56c0.
+function digitToUnicodeSupercript(n: number): string {
+  const subst = [0x2070, 185, 178, 179, 0x2074, 0x2075, 0x2076, 0x2077, 0x2078, 0x2079];
```
toloudis, line 28: This is how the pros do it.  I would have just used "1e-20" and be done with it.

src/colorizer/Dataset.ts
ShrimpCryptid: It's not a huge difference for code niceness I think, but I did make this change because it also captures a potential edge case! The new default is for frame dimensions to be 0 (width & height= 0), and the scale bar handling code will now hide the scale bar and skip the calculations so it doesn't divide by 0.

frasercl (approved):
Looks good!

src/colorizer/Dataset.ts
toloudis: I like it.  I feel like Partials make such a weak statement about types, that it's like one level removed from `any`.  This kind of change is generally beneficial.  Less `?`s too.

PR #140, opened 2023-11-07
refactor: Moved tab contents to their own components

ShrimpCryptid:
Problem
=======
Refactors the tab view in `App.tsx`.

Solution
========
- Moves the Plot and Settings tabs to their own components.
- Renames `FilterThresholdPanel.tsx` to `FilterThresholdsTab.tsx`.
- Adds styled utility components (`FlexDiv`, `FlexColumn`, `FlexRow`, `FlexRowCentered`) for boilerplate CSS styling.

Steps to Verify:
----------------
1. Open the preview link and click around on the panel view. Everything should behave as it did before.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -764,6 +764,7 @@ function App(): ReactElement {
                           dataset={dataset}
                           featureName={featureName}
                           selectedTrack={selectedTrack}
+                          disabled={disableUi}
                         />
                       </div>
                     ),
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -731,19 +731,16 @@ function App(): ReactElement {
                 disabled={disableTimeControlsUi}
                 wrapIncrement={true}
               />
-              <FlexRowCentered>
-                <span className={styles.verticalDivider} style={{ height: "24px", margin: "0 8px" }}></span>
-                <div style={{ width: "200px", maxWidth: "60vw" }}>
-                  <PlaybackSpeedControl
-                    fps={playbackFps}
-                    onChange={(fps) => {
-                      setPlaybackFps(fps);
-                      timeControls.setPlaybackFps(fps);
-                    }}
-                    disabled={disableTimeControlsUi}
-                  />
-                </div>
-              </FlexRowCentered>
+              <div style={{ display: "flex", flexDirection: "row", flexGrow: 1, justifyContent: "flex-end" }}>
+                <PlaybackSpeedControl
+                  fps={playbackFps}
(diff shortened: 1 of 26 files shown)


src/styles/utils.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const FlexRow = styled(FlexDiv)`
+  flex-direction: row;
+`;
+
+/**
+ * A flexbox container that lays out children in a row, and centers all items
+ * vertically.
+ * @param $gap: The gap, in pixels, between each child element. 0 by default.
+ */
+export const FlexRowCentered = styled(FlexRow)`
```
ShrimpCryptid, line 55: Does this need to be more descriptive? `FlexRowCenteredVertically` or something?

src/App.tsx
```diff
@@ -758,31 +758,11 @@ function App(): ReactElement {
                     key: "plot",
```
ShrimpCryptid, line 758: Should this whole Tabs section be a component? I'm debating on this because it feels like I'd be passing a lot of arguments down multiple levels in the hierarchy.

src/styles/utils.tsx
toloudis: I don't mind what it is, but maybe FlexRowAlignCenter? idk

toloudis (approved):
LGTM

frasercl (approved):
Looks good, always in favor of more code organization! Looking at the large number of imports at the top of App.tsx, there could be more work to be done along these lines.

src/App.tsx
meganrm: If you think it would clean things up, but it doesn't seem like a high priority to me. I like it when an App file shows at a high level everything that is going to be rendered in the App, so having these two tabs here makes sense to me, whereas there are other items above this that are dealing with user interaction that could be factored out first I think

src/components/tabs/PlotTab.tsx
```diff
@@ -0,0 +1,78 @@
...
+  disabled?: boolean;
+  findTrack: (trackId: number, seekToFrame?: boolean) => void;
+  currentFrame: number;
+  dataset: Dataset | null;
+  featureName: string;
+  selectedTrack: Track | null;
+};
+
+const defaultProps: Partial<PlotTabProps> = {
+  disabled: false,
```
meganrm, line 38: I feel like either all the props should have defaults set here or none of them should, and somewhere above all the defaults are set and sent in.

PR #144, opened 2023-11-10
feat: Support `/allen` file links

### LINKED ISSUE #133: Accept `/allen` resource links in the Load menu
## Use Case

Allow users to directly paste `/allen`-style filepaths into the Load dialog, and have it replace the prefix with the relevant URL to load it as an HTTPS path.

## Acceptance Criteria

- `/allen` paths from relevant directories are accepted over VPN.
- Also, should identify unsupported `/allen` links and throw up a custom error message ("This directory can't be loaded directly. Contact [insert contact info here] for more details."

ShrimpCryptid:
Problem
=======
Closes #133!

Solution
========
- Adds a lookup table for accepted `/allen` directories, with mappings to their HTTPS equivalents.
- Updates the error messages to give more info when loading unsupported allen directories.
- Compatible with the recent datasets dropdown!

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-144/
2. Load this `/allen` directory: `/allen/aics/assay-dev/computational/data/EMT_deliverable_processing/LH_Analysis/Version2_ColorizerTest`
3. The dataset should appear correctly.

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/colorizer/utils/url_utils.ts
+++ b/src/colorizer/utils/url_utils.ts
@@ -150,7 +150,7 @@ export function isAllenPath(input: string): boolean {
 }
 
 /**
- * Attempst to convert an Allen path to an HTTPS resource path.
+ * Attempts to convert an Allen path to an HTTPS resource path.
  * @returns Returns null if the path was not recognized or could not be converted,
  * otherwise, returns an HTTPS resource path.
  */
--- a/src/colorizer/utils/url_utils.ts
+++ b/src/colorizer/utils/url_utils.ts
@@ -144,10 +144,23 @@ export function isUrl(input: string | null): boolean {
 }
 
 /**
+ * Normalizes a file path to use only single forward slashes. Replaces
+ * backwards slashes with forward slashes, and removes double slashes.
+ */
+function normalizeFilePathSlashes(input: string): string {
+  // Replace all backslashes with forward slashes
+  input = input.replaceAll("\\", "/");
+  // Replace double slashes with single
+  input = input.replaceAll("//", "/");
+  return input;
+}
+
+/**
  * Returns whether the input string is a path to an Allen file server resource.
+ * Matches any path that starts with `/allen/`, normalizing for backwards and double slashes.
  */
 export function isAllenPath(input: string): boolean {
-  return input.startsWith(ALLEN_FILE_PREFIX);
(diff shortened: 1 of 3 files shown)


src/colorizer/utils/url_utils.ts
```diff
@@ -14,6 +14,12 @@ const URL_PARAM_RANGE = "range";
 const URL_PARAM_COLOR_RAMP = "color";
 const URL_COLOR_RAMP_REVERSED_SUFFIX = "!";
 
+const ALLEN_FILE_PREFIX = "/allen/";
```
toloudis, line 17: it seems like you don't really need this prefix if you just have the remapping table below.

src/colorizer/utils/url_utils.ts
```diff
@@ -14,6 +14,12 @@ const URL_PARAM_RANGE = "range";
...
+const ALLEN_PREFIX_TO_HTTPS: Record<string, string> = {
+  "/allen/aics/assay-dev": "https://dev-aics-dtp-001.int.allencell.org/assay-dev",
+  "/allen/aics/microscopy": "https://dev-aics-dtp-001.int.allencell.org/microscopy",
+};
```
toloudis, line 21: I think I would check for `\\allen` and `//allen` as well.  Sometimes people paste in windows paths with backslashes, and sometimes the leading slash is doubled.  So maybe normalize the slashes to forward slash first, if the first character is a slash?

toloudis: (in the convert function, that is)

toloudis (approved):
LGTM, just a couple minor notes!

frasercl (approved):
LGTM! Besides @toloudis's comments, which I think are worth a look

src/colorizer/utils/url_utils.ts
ShrimpCryptid: I'm going to keep this in because I use it to warn users if they are using an `/allen` file link but it's not one of the ones we support. (see LoadDatasetButton.tsx, line 130)

ShrimpCryptid: I'm going to keep this in because I use it to warn users if they are using an `/allen` file link but it's not one of the ones we support. (see LoadDatasetButton.tsx, line 130)

PR #145, opened 2023-11-22
allow more types of NaN data

toloudis:
Problem
=======
App was crashing (exception) on being unable to convert a null feature value to a string.

Solution
========
Handle this type of value a bit more gracefully.
Check for nulls.

### DIFF
--- a/src/colorizer/utils/math_utils.ts
+++ b/src/colorizer/utils/math_utils.ts
@@ -5,7 +5,7 @@
  * after the decimal place. Optionally ignores integers and returns them as-is.
  */
 export function numberToStringDecimal(
-  input: number | undefined,
+  input: number | undefined | null,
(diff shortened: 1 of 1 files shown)


src/colorizer/utils/math_utils.ts
```diff
@@ -9,7 +9,7 @@ export function numberToStringDecimal(
   decimalPlaces: number,
```
ShrimpCryptid, line 9: I can't make a suggestion because this line is a little too far up, but also should switch the type of `input` to explicitly accept `null` types here too.

ShrimpCryptid (approved):
Good catch, thank you! Added a quick typing change but otherwise good to go.

frasercl (approved):
Looks good!

PR #149, opened 2023-11-30
feat: Physical time

### LINKED ISSUE #61: physical time
## Use Case

As a user, I would like to see the wall clock time elapsed between frames or since the first frame.  This is useful for understanding how fast in real time the cells are actually changing.

## Acceptance Criteria

Currently the data is only represented by image number (aka frame number) in sequence. 
1. We need to obtain the wall clock times from the data source and incorporate it in the data.
2. Displaying the time elapsed from start as an overlay on the view (or elsewhere) would complete this request.

## Details

Bonus:   It could also be interesting to display this time value in the plot of a track, to help understand how long a track is.  This probably requires some nontrivial UX design.  We could promote this bonus item to a new issue.

ShrimpCryptid:
Problem
=======
Closes #61, physical time!

Solution
========
- Adds new dataset metadata parameters (start time + frame duration). The default nucmorph datasets have been updated to use these values.
- Plumbing code to pass parameters + config down to the `CanvasOverlay`.
- Rendering + formatting for the new timestamp.
- Changes the rendering pattern for `CanvasOverlay`.
  - A lot of these changes needed to happen to allow the overlay elements (scale bar + timestamp) to have a transparent background beneath them. The biggest change is delaying element rendering, so that the size of the elements can be summed and a backdrop can be drawn first beneath them.

## Type of change

Steps to Verify:
----------------
1. Open preview link (https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-149/)
2. Make sure the "Show timestamp" checkbox is enabled.
3. Play, and observe the timestamp updating.
4. Also, try disabling/enabling the scale bar and timestamp. The background will disappear.

Screenshots (optional):
-----------------------
image

### Video Talk-through (w/ Audio)

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/200feb1a-0554-4fe2-986d-7bfe4fcc222c

Addendum video about different timestamp formats:

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/28943b34-2e1f-45fd-9ae9-2c3d364a23da

Thanks for contributing!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -98,6 +98,7 @@ function App(): ReactElement {
   const [isColorRampRangeLocked, setIsColorRampRangeLocked] = useState(false);
   const [showTrackPath, setShowTrackPath] = useState(false);
   const [showScaleBar, setShowScaleBar] = useState(true);
+  const [showTimestamp, setShowTimestamp] = useState(true);
 
   // Provides a mounting point for Antd's notification component. Otherwise, the notifications
   // are mounted outside of App and don't receive CSS styling variables.
@@ -673,6 +674,7 @@ function App(): ReactElement {
                 }}
                 onMouseLeave={() => setShowHoveredId(false)}
                 showScaleBar={showScaleBar}
+                showTimestamp={showTimestamp}
               />
             </HoverTooltip>
 
@@ -832,6 +834,15 @@ function App(): ReactElement {
                             }}
                           >
                             Show scale bar
+                          </Checkbox>
+                          <Checkbox
+                            type="checkbox"
+                            checked={showTimestamp}
+                            onChange={() => {
+                              setShowTimestamp(!showTimestamp);
+                            }}
+                          >
+                            Show timestamp
                           </Checkbox>
                         </div>
                       </div>
--- a/src/colorizer/CanvasOverlay.ts
+++ b/src/colorizer/CanvasOverlay.ts
@@ -16,10 +16,10 @@ export type ScaleBarOptions = StyleOptions & {
 
 export type TimestampOptions = StyleOptions & {
   visible: boolean;
-  maxTimestampSeconds: number;
-  currentTimestampSeconds: number;
-  frameDurationSeconds: number;
-  startTimeSeconds: number;
+  maxTimeSec: number;
+  currTimeSec: number;
+  frameDurationSec: number;
+  startTimeSec: number;
 };
 
 const defaultStyleOptions: StyleOptions = {
@@ -39,10 +39,15 @@ const defaultScaleBarOptions: ScaleBarOptions = {
 const defaultTimestampOptions: TimestampOptions = {
   ...defaultStyleOptions,
   visible: false,
-  frameDurationSeconds: 1,
-  startTimeSeconds: 0,
-  maxTimestampSeconds: 1,
-  currentTimestampSeconds: 0,
+  frameDurationSec: 1,
+  startTimeSec: 0,
+  maxTimeSec: 1,
+  currTimeSec: 0,
+};
+
+type SizeAndRender = {
+  size: Vector2;
+  render: () => void;
 };
 
 /**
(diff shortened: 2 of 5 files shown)


src/colorizer/Dataset.ts
```diff
@@ -29,6 +29,9 @@ export type DatasetMetadata = {
     height: number;
     units: string;
   };
+  frameDurationSeconds: number;
+  /* Optional offset for the timestamp. */
+  startTimeSeconds: number;
```
ShrimpCryptid, line 34: Somewhat debatable how useful this is, but I figured it could be useful if we ever decide to show a slice of a dataset (or want to set this in terms of like, say, colony time.)

src/colorizer/CanvasOverlay.ts
```diff
@@ -1,38 +1,94 @@
...
 import { numberToSciNotation } from "./utils/math_utils";
 
-export type ScaleBarOptions = {
-  minWidthPx: number;
+export type StyleOptions = {
   fontSizePx: number;
   fontFamily: string;
   fontColor: string;
+  fontStyle: string;
+};
```
ShrimpCryptid, line 9: Styling options were identical across the scale bar and timestamp, so this now becomes its own type that the scale bar + timestamp options extend.

src/App.tsx
```diff
@@ -834,6 +836,15 @@ function App(): ReactElement {
...
                           </Checkbox>
+                          <Checkbox
+                            type="checkbox"
+                            checked={showTimestamp}
+                            onChange={() => {
+                              setShowTimestamp(!showTimestamp);
+                            }}
+                          >
+                            Show timestamp
+                          </Checkbox>
```
ShrimpCryptid, line 847: Should this checkbox be disabled if the dataset doesn't have the right metadata to show time?

src/colorizer/CanvasOverlay.ts
```diff
@@ -1,38 +1,96 @@
+import { Vector2 } from "three";
 import { numberToSciNotation } from "./utils/math_utils";
 
-export type ScaleBarOptions = {
-  minWidthPx: number;
+export type StyleOptions = {
```
toloudis, line 4: don't export? where else is this used?

src/colorizer/CanvasOverlay.ts
```diff
@@ -63,18 +159,23 @@ export default class CanvasOverlay {
     }
   }
 
-  private renderScaleBar(): void {
-    if (!this.scaleBarOptions.unitsPerScreenPixel || !this.scaleBarOptions.visible) {
-      return;
-    }
+  // NOTE: Rendering of the scale bar and timestamp is deferred here, since we need to draw a background behind them
+  // with an appropriate size. These methods return a callback that renders the scale bar or timestamp to the canvas,
```
toloudis, line 163: what methods return a callback?

toloudis: not sure if this comment is adding value... since RenderInfo is pretty clear as to what it does.  Things seem to be named in an understandable way.

src/colorizer/CanvasOverlay.ts
```diff
@@ -90,47 +191,202 @@ export default class CanvasOverlay {
...
+    };
+  }
+
+  /**
+   * Draws the background overlay in the bottom right corner of the canvas.
+   * @param ctx Canvas context to render to.
+   * @param size Size of the background overlay.
+   * @param options Configuration for the background overlay.
+   */
+  private static renderBackground(ctx: CanvasRenderingContext2D, size: Vector2, options: BackgroundOptions): void {
```
toloudis, line 338: minor concern for future: what if a user said they want the time overlay at bottom left and scale bar at bottom right?  Are they too tightly coupled in this code, just for the sake of having one single background rectangle?

src/colorizer/CanvasOverlay.ts
```diff
@@ -90,47 +191,202 @@ export default class CanvasOverlay {
...
     ctx.clearRect(0, 0, this.canvas.width, this.canvas.height);
 
-    // Draw scale bar and other elements
-    this.renderScaleBar();
+    // Get dimensions + render methods for the elements, but don't render yet so we can draw the background
+    // behind them.
+    const origin = this.backgroundOptions.marginPx.clone().add(this.backgroundOptions.paddingPx);
+    const { sizePx: scaleBarDimensions, render: renderScaleBar } = this.getScaleBarRenderer(ctx, origin);
+    origin.y += scaleBarDimensions.y;
+    const { sizePx: timestampDimensions, render: renderTimestamp } = this.renderTimestamp(ctx, origin);
```
toloudis, line 372: consider naming `getScaleBarRenderer` and `renderTimestamp` in a more consistent way, since they seem to return the same type of thing.

src/colorizer/Dataset.ts
toloudis: YAGNI 😉 but no need to remove it now.

src/App.tsx
toloudis: We *could* show frame numbers as the time overlay in that case, but it's probably not going to bother anyone if the checkbox is there either way.

toloudis:
The code all looks pretty reasonable to me.  I wonder if some unit tests are in order here.  (The time string formatting function made me think this, but it would be cool to do for the sizing code too if possible.)

src/colorizer/CanvasOverlay.ts
ShrimpCryptid: Super fair! One of the nice things about the current code is that, because each element reports its dimensions, they're not inherently dependent on each other for rendering, and don't make assumptions about other elements existing.

But yeah, all the current components assume they're in the same corner for now. The styling and options objects will make it easier to configure this in the future.

src/colorizer/CanvasOverlay.ts
ShrimpCryptid: `CanvasWrapper` does use `StyleOptions` for setting up shared properties. It technically doesn't have to, but it gives nicer DX through type hinting.

toloudis: I didn't see an import for it

ShrimpCryptid: hubris

src/colorizer/CanvasOverlay.ts
```diff
@@ -1,38 +1,96 @@
...
-  minWidthPx: 80,
+export type TimestampOptions = StyleOptions & {
+  visible: boolean;
+  maxTimeSec: number;
+  currTimeSec: number;
+  frameDurationSec: number;
+  startTimeSec: number;
+};
+
+export type BackgroundOptions = {
```
toloudis, line 26: do these other options types need to be exported?  ScaleBarOptions, TimestampOptions, BackgroundOptions?

ShrimpCryptid: I guess not, because they only ever define parameter types accepted by the CanvasOverlay. Updated!

toloudis (approved):

interim17 (approved):

interim17:
LGTM :) The style and behavior are all smooth on my machine.

Issue #148, opened 2023-11-30
UX Implementation for categorical features


ShrimpCryptid:
Implementation of design by @lynwilhelm for categorical features. See #73 for design discussion!

![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/d68739f8-87d5-4828-91d2-5edf155ba7ba)

https://www.figma.com/file/DcFEsrszbUOK0qU1E9wrH0/2023-Aug-NucMorph-Viz-App?type=design&node-id=667%3A3470&mode=design&t=3SHhzQgmYFjyMdGQ-1

ShrimpCryptid:
![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/b960c12e-a770-4fe4-9389-7f694bacd779)

PR #151, opened 2023-12-04
feat: Feature types in Dataset

### LINKED ISSUE #150: Categorical Features Pt. 1: Update Dataset to allow categorical data
Part 1 work ticket from #148 
  -  Add `type` property to feature metadata in `manifest.json`. `type` should accept a string from a list of arguments (`["continuous", "discrete", "categorical"]`).
    - This must live in manifest and not the feature JSON because the data needs to be top-level accessible so that the UI can reflect categories/feature types when filtering etc.
  - Add `categories` property to feature metadata, which should accept an array of strings, which are the ordered names of the categories. (`"categories": ["Category 0", "Category 1", "Category 2"]")
  - Note: categorical features will use integers between 0 and 12.
  - Update `Dataset.ts` to include this information.
    - Validation? Throw errors if data is not formatted as expected? Warnings?
    - If a feature does not include a `type`, assume it to be continuous.
    - Add unit tests with example JSON data.
  - Manually update the example dataset so it has categorical, continuous, and discrete features.

ShrimpCryptid:
Problem
=======
Part 1 of #148, closes #150!

Solution
========
Adds support for feature types in `Dataset.ts`!
- Additional validated fields added in `FeatureData` for `type` and `categories`.
- Added additional tests for loading feature metadata.

EDIT:
- Changes the Dataset spec so that all feature data is stored under the `features` key.
  - Adds backwards compatibility for the old dataset format; new documentation pending

## Type of change

Loading new dataset: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-151/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Ffeature%2Fcategorical-dataset%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=Continuous%20Feature&track=0&color=matplotlib-cool
Backwards compatibility with old datasets: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-151/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama%20Bear&feature=Height&color=matplotlib-cool


ShrimpCryptid:
Converting back to draft because I forgot to update the default dataset!

documentation/example_dataset/manifest.json
```diff
@@ -1,9 +1,27 @@
...
+      "type": "continuous"
+    },
+    "Discrete Feature": {
+      "units": "nodes",
+      "type": "discrete"
+    },
+    "Categorical Feature": {
+      "type": "categorical",
+      "categories": ["Spicy", "Sour", "Bitter", "Sweet", "Dry"]
+    }
```
toloudis, line 25: seeing this example, it looks weird to separate the metadata from the features, when instead it couldlook like this:
```
"features": { 
    "FeatureName1": {
        "data":"feature_0.json",
        "type":"categorical",
        "categories":[...]
    },
    etc
}
```

ShrimpCryptid: I was thinking that too... I'm not sure what the best way to handle that change would be in a way that doesn't break all of our existing datasets.

toloudis: one of 2 options:
1. fixup all existing datasets (not as hard as it sounds)
2. have the client check for the existence of `featuresMetadata` in the incoming json, for backward compatibility only

ShrimpCryptid: Leigh might have some datasets lying around, I'd have to ask. I don't think anyone else has any datasets we don't know about...?

ShrimpCryptid: I'll add backwards compatibility for now.

ShrimpCryptid: Dataset spec updated! I'll start working on a documentation update for `colorizer-data`.

src/colorizer/Dataset.ts
```diff
@@ -9,17 +9,41 @@ import Track from "./Track";
...
+ * was stored separately from the feature file path declaration.
+ */
+type DeprecatedFeatureMetadata = {
+  units?: string | null;
+  type?: string | null;
+  categories?: string[] | null;
+};
+
+/** JSON metadata for dataset features. */
+type FeatureMetadata = {
```
toloudis, line 42: Nit: regarding naming, it's slightly uncomfortable that `FeatureData` is the runtime datatype that the app uses but `FeatureMetadata` is the json file format datatype.  I'm not sure if you can namespace things or decorate the name with "FileFormat" or something.

src/colorizer/Dataset.ts
```diff
@@ -44,10 +68,16 @@ const defaultMetadata: DatasetMetadata = {
   startTimeSeconds: 0,
 };
 
+/** Maps from the feature label to its metadata, including relative filepath, type, and units. */
+type FeatureMap = Record<string, FeatureMetadata>;
+/** Maps from the feature label to its relative filepath only. */
+type DeprecatedFeatureMap = Record<string, string>;
```
toloudis, line 74: Nit: If you ever have to do this again, you will have DeprecatedDeprecatedFeatureMap.  This is a reasonable place to have FeatureMapV1, FeatureMapV2, and FeatureMap=FeatureMapV2.  That's how we did it in Simularium viewer.

Another way to deal with this is to convert extremely early, as soon as the json file is read, so that you only ever have to deal with FeatureMap and DeprecatedFeatureMap only need exist for the single conversion function.  It could be cleaner to move it out of here and into its own ts file

src/colorizer/Dataset.ts
```diff
@@ -246,21 +348,36 @@ export default class Dataset {
...
     this.tracksFile = manifest.tracks;
     this.timesFile = manifest.times;
     this.centroidsFile = manifest.centroids;
 
     this.frames = new FrameCache(this.frameFiles.length, MAX_CACHED_FRAMES);
-    const featuresPromises: Promise<void>[] = this.featureNames.map((name) =>
+
+    // Load feature data -> switch between deprecated and new feature type loading.
+    let featuresToMetadata: Record<string, FeatureMetadata> = {};
+    if (this.isFeatureDeprecated(manifest.features)) {
```
toloudis, line 362: based on my earlier comment, this could look something like `manifest.features = updateManifestVersion(manifest.features, manifest.featureMetadata)` .. roughly speaking

toloudis (approved):
LGTM, I offered some comments aimed toward making the deprecated format handling cleaner but nothing super serious

ShrimpCryptid:
Made a big refactor to shuffle types around; re-requesting your review @toloudis as a result. Thank you for your suggestions on how to make Dataset cleaner!

src/colorizer/Dataset.ts
```diff
@@ -9,32 +9,26 @@ import Track from "./Track";
...
-  /** Dimensions of the frame, in scale units. Default width and height are 0. */
-  frameDims: {
-    width: number;
-    height: number;
-    units: string;
-  };
-  frameDurationSeconds: number;
-  /* Optional offset for the timestamp. */
-  startTimeSeconds: number;
-};
```
ShrimpCryptid, line 35: Moved into `dataset_utils.ts`

src/colorizer/Dataset.ts
```diff
@@ -106,40 +88,77 @@ export default class Dataset {
...
+        tex: source.getTexture(FeatureDataType.F32),
+        data: source.getBuffer(FeatureDataType.F32),
+        min: source.getMin(),
+        max: source.getMax(),
+        units: metadata?.units || "",
+        type: featureType,
+        categories: featureCategories || null,
+      },
+    ];
   }
```
ShrimpCryptid, line 154: Rewrote `loadFeature` so it returns the result instead of directly inserting into `this.features[name]`. This created a race condition that led to features being sorted differently each time they were loaded.

src/colorizer/Dataset.ts
```diff
@@ -106,40 +88,77 @@ export default class Dataset {
...
+    const featureType = this.getFeatureTypeFromString(metadata?.type || "", FeatureType.CONTINUOUS);
+    const featureCategories = metadata?.categories;
+    // Validation
+    if (featureType === FeatureType.CATEGORICAL && !metadata?.categories) {
+      throw new Error(`Feature ${name} is categorical but no categories were provided.`);
+    }
+    if (featureCategories && featureCategories.length > MAX_FEATURE_CATEGORIES) {
+      throw new Error(
+        `Feature ${name} has too many categories (${featureCategories.length} > max ${MAX_FEATURE_CATEGORIES}).`
+      );
```
meganrm, line 139: was this just an arbitrary cutoff?

ShrimpCryptid: Yeah, we talked with scientists to find a value that seemed reasonably large and beyond what they expected we'd need. Part of the challenge is that every color palette needs to have at least `MAX_FEATURE_CATEGORIES (=12)` colors, so capping it at an arbitrary value was necessary.

src/colorizer/Dataset.ts
```diff
@@ -151,8 +170,43 @@ export default class Dataset {
...
+  /**
+   * Returns the FeatureType of the given feature, if it exists.
+   * @param name Feature name to retrieve
+   * @throws An error if the feature does not exist.
+   * @returns The FeatureType of the given feature (categorical, continuous, or discrete)
+   */
+  public getFeatureType(name: string): FeatureType {
+    const featureData = this.features.get(name);
+    if (!featureData) {
+      throw new Error(`Feature ${name} does not exist.`);
```
meganrm, line 189: I'm a little confused why you have this error check here but not in `getFeatureData`. it seems like `getFeatureData` should be checking for wether the feature exists and any function that is trying to unpack that data should call `getFeatureData` instead of `this.features.get(name)`

tests/Dataset.test.ts
```diff
@@ -54,14 +40,120 @@ describe("Dataset", () => {
...
+        const dataset = new Dataset(defaultPath, undefined, new MockArrayLoader());
+        const mockFetch = makeMockFetchMethod(defaultPath, manifest);
+        await dataset.open(mockFetch);
+
+        expect(dataset.isFeatureCategorical("feature1")).to.be.false;
+        expect(dataset.isFeatureCategorical("feature2")).to.be.false;
+        expect(dataset.isFeatureCategorical("feature3")).to.be.false;
+        expect(dataset.isFeatureCategorical("feature4")).to.be.false;
+        expect(dataset.isFeatureCategorical("feature5")).to.be.true;
+        expect(dataset.getFeatureCategories("feature5")).to.deep.equal(["small", "medium", "large"]);
```
meganrm, line 121: nit but I missed this function call on the first time reading it. It seems like a public function like this should be its own unit tests, and be checking for the cases you have below, but on the function instead of the whole class.

src/colorizer/Dataset.ts
```diff
@@ -272,14 +321,19 @@ export default class Dataset {
...
     this.outliers = outliers;
     this.trackIds = tracks;
     this.times = times;
     this.centroids = centroids;
     this.bounds = bounds;
 
+    // Keep original sorting order of features by inserting in promise order.
+    featureResults.forEach(([name, data]) => {
+      this.features.set(name, data);
+    });
```
frasercl, line 335: smart!

frasercl (approved):
Lots of comments but nothing worth blocking merging for. This looks good!

src/colorizer/utils/dataset_utils.ts
```diff
@@ -0,0 +1,92 @@
...
+export type AnyManifestFile = ManifestFileV1 | ManifestFileV2;
+
+///////////// Conversion functions /////////////////////
+
+/**
+ * Returns whether the dataset is using the older, deprecated manifest format, where feature metadata
+ * was stored in a separate object from the `feature` file path declaration.
+ */
+function isV1(manifest: AnyManifestFile): boolean {
+  return typeof Object.values(manifest.features)[0] === "string";
```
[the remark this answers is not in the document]
ShrimpCryptid: Added!

src/colorizer/utils/dataset_utils.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Converts potentially outdated manifests to the latest manifest format.
+ * @param manifest Manifest object, as parsed from a JSON file.
+ * @returns An object with fields reflecting the most recent ManifestFile spec.
+ */
+export const update_manifest_version = (manifest: AnyManifestFile): ManifestFile => {
```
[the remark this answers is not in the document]
ShrimpCryptid: My brain was in Python mode whoops

src/colorizer/utils/dataset_utils.ts
```diff
@@ -0,0 +1,92 @@
...
+// array of metadata objects.
+type ManifestFileV2 = Omit<ManifestFileV1, "features" | "featureMetadata"> & {
+  features: {
+    name: string;
+    data: string;
+    units?: string;
+    type?: string;
+    categories?: string[];
+  }[];
+};
```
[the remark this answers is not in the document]
ShrimpCryptid: I think I'm going to keep it as an inline type for now because I think it's more readable like this? But if another version gets added that continues to change `features`, it'll probably be worth moving to its own type. Thinking through this was a good exercise, thank you!

PR #152, opened 2023-12-05
feat: Categorical palettes

ShrimpCryptid:
Problem
=======
Part 2 of #148! Adds internal definitions for color palettes, and updates the shader to support them. (No UI changes yet!)

Solution
========
The changes are pretty simple, the constants files just take up a lot of lines :')

- Adds constants for the categorical color palettes, according to Lyndsay's designs.
  - Adds unit tests for color palette constants for basic validation.
- `ColorizeCanvas` will switch to a categorical palette when a categorical feature is chosen, and back to a regular ramp when a continuous or discrete value is chosen.

## Type of change

Validation
=========
This preview link shows the new categorical coloring, though it's not terribly exciting because there's no UI to change palettes yet. Try switching between features.

https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-152/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Ffeature%2Fcategorical-colors%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=Categorical%20Feature&track=0&t=4&color=matplotlib-cool

It's using this palette, though I've skipped some values so the "cells" skip some colors in the palette:
image


src/colorizer/ColorizeCanvas.ts
```diff
@@ -350,8 +355,14 @@ export default class ColorizeCanvas {
     this.pickMaterial.uniforms[name].value = value;
   }
 
+  /** Sets the current color ramp. Used when a continuous or discrete feature is selected. */
```
meganrm, line 358: just out of curiosity, why is discrete here instead of with categorical features? are discrete features usually always a ramp, ie numbers in order, just ints?

src/constants/categorical_palettes.ts
```diff
@@ -0,0 +1,235 @@
...
+      "#5D5D6A",
+      "#FFFFA4",
+      "#3A6CAC",
+      "#ED1B7D",
+      "#83C885",
+    ],
+  },
+  {
+    key: "matplotlib-tab10",
+    name: "Set 1",
```
meganrm, line 29: This may be a question for UI but it seems like these could have better names? I know you have the preview of the colors so it's not as critical

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Yes! Discrete features still use the ramps. They're essentially the same as continuous features but just fix the values at integer points.

I've been internally labeling integer features as `discrete`, and ones that are distinct categories `categorical`. I'm not sure if that breaks the convention we normally use since there's been some mixing over discrete/categorical values in conversation.

src/constants/categorical_palettes.ts
ShrimpCryptid: The names are designed to be flexible since everything is keyed with its own constant name. I'll check with Lyndsay on this one.

Were you thinking that they would have more colloquial names (like "Tropical" or "Marsh") or something else?

ShrimpCryptid: Gotcha!

toloudis (approved):
All LGTM.

I have a slight concern that we haven't chosen the best palettes. This could be addressed in a future PR.  I tend to trust the adobe and matplotlib more than iwanthue. My suggestion is that we pick a selection of palettes from a source like https://kwstat.github.io/pals/reference/discrete.html which has research paper references for most of the selections. It would be nice to be able to cite that the palettes have a basis in color theory and visualization science.   I also like the algorithm provided by https://medialab.github.io/iwanthue/ as a way to have a random generator, as we have discussed in Slack. But that's probably out of scope as long as we provide a decent small selection of presets to start with.

ShrimpCryptid:
> All LGTM.
> 
> I have a slight concern that we haven't chosen the best palettes. This could be addressed in a future PR. I tend to trust the adobe and matplotlib more than iwanthue. My suggestion is that we pick a selection of palettes from a source like https://kwstat.github.io/pals/reference/discrete.html which has research paper references for most of the selections. It would be nice to be able to cite that the palettes have a basis in color theory and visualization science. I also like the algorithm provided by https://medialab.github.io/iwanthue/ as a way to have a random generator, as we have discussed in Slack. But that's probably out of scope as long as we provide a decent small selection of presets to start with.

I made an issue for the random generator (#159) and I'll make another issue for cleaning out color palettes. Thank you!

PR #153, opened 2023-12-06
feat: Categorical color selection

ShrimpCryptid:
Problem
=======
Continues work on #148! Adds the ability to select from palettes for categorical colors and color individual categories.

Solution
========
- `ColorRampDropdown` can now switch to show categorical palettes, instead of only color ramps. Some custom styling was added to add visual separation between each palette.
  - Shows a number of colors based on the number of categories present in the selected feature.
- Adds a varying number of labeled color pickers through the new `CategoricalColorPicker` component, which serves as both the legend and a way to customize color ramps.
- Updates the canvas tooltip to show categories.
- Reorders palettes and replaces duplicate colors.
- Adds additional categorical features to the example dataset, to show different numbers of categories.

## Type of change

Steps to Verify:
----------------
1. Open this preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-153/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Ffeature%2Fcategorical-color-selection%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=Categorical%20Feature%20(5)&color=matplotlib-cool
1. Switch the feature between the three different categorical features. Different numbers of color pickers should be shown for each, and the number of colors in the displayed palette should also change.
1. Change the palette from the dropdown.
1. Click on a color picker and edit the color. You should see it change on the viewport and in the dropdown palette viewer.

Screenshots (optional):
-----------------------
image

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/6318b6aa-d77c-4cc1-9221-191f2e9a8614

-----------------------
1. `ColorRampDropdown.tsx`
3. `CategoricalColorPicker.tsx`


src/colorizer/ColorizeCanvas.ts
```diff
@@ -555,11 +562,9 @@ export default class ColorizeCanvas {
    */
   updateRamp(): void {
     if (this.featureName && this.dataset?.isFeatureCategorical(this.featureName)) {
-      const palette = this.categoricalPalette;
-      const colorRamp = new ColorRamp(palette.colorStops);
-      this.setUniform("colorRamp", colorRamp.texture);
+      this.setUniform("colorRamp", this.categoricalPalette.texture);
```
ShrimpCryptid, line 565: By saving `categoricalPalette` as a `ColorRamp`, we don't need to regenerate a `ColorPalette` on every render.

src/constants/categorical_palettes.ts
```diff
@@ -79,19 +133,19 @@ const rawPaletteData: RawColorData[] = [
...
-    name: "Bright",
+    key: "neon",
+    name: "Neon",
     colorStops: [
       "#33FCFE",
       "#F91EF8",
       "#49FE86",
       "#FAEC4C",
       "#AF26F3",
-      "#49FEFE",
```
ShrimpCryptid, line 90: Duplicated color

tests/ColorRampDropdown.test.tsx
ShrimpCryptid, line 1: Minimal changes in this file, just changed indentation and added missing properties

src/App.tsx
```diff
@@ -530,6 +542,17 @@ function App(): ReactElement {
...
+  if (lastHoveredId !== null && dataset) {
+    const featureVal = getFeatureValue(lastHoveredId);
+    const categories = dataset.getFeatureCategories(featureName);
+    if (categories !== null) {
+      hoveredFeatureValue = categories[Number.parseInt(featureVal, 10)];
+    } else {
+      hoveredFeatureValue = featureVal;
+    }
+  }
+
```
ShrimpCryptid, line 555: Moved from line 647 and expanded to handle categorical features.

src/constants/categorical_palettes.ts
ShrimpCryptid, line 1: Changed ordering of palettes, made ordering tweaks to the color stops.

tests/ColorRampDropdown.test.tsx
ShrimpCryptid, line 1: Spacing/indentation only change

src/components/ColorRampDropdown.module.css
```diff
@@ -42,20 +55,32 @@
...
   opacity: 0;
 
-  transition: all ease-in 0.2s 0.25s;
+  transition: all ease-in 0.2s 0.25s, gap 0s;
+}
+
+/** Overrides for when the categorical dropdown is being used, instead of
+ * the normal color ramp version. Apply an extra border to each element,
+ * making it thicker when hovered.
+ */
```
toloudis, line 79: did you try adding this extra outline to the continuous ramp dropdown so that it wouldn't have to be special-cased?

src/components/ColorRampDropdown.tsx
```diff
@@ -6,20 +6,75 @@ import styles from "./ColorRampDropdown.module.css";
...
+ * @returns A style object with a linear gradient background.
+ *
+ * @example
+ * ```
+ * const colors = [new Color(0xffffff), new Color(0xcccccc), new Color(0x000000)];
+ * const style = getColorPaletteStyling(colors);
+ * // style = {background : linear-gradient(to right, #ffffff 0.00%, #ffffff 33.33%, #cccccc 33.33%, #cccccc 66.67%, #000000 66.67%, #000000 100.00%}
+ * ```
+ */
+function getColorPaletteStyling(colors: Color[]): { background: string } {
```
toloudis, line 47: Nit: Something about this function name seems odd but I'm not sure if `createCSSGradientForDiscretePalette` would be any better.

src/components/ColorRampDropdown.tsx
```diff
@@ -106,22 +195,37 @@ const ColorRampSelector: React.FC<ColorRampSelectorProps> = (propsInput): ReactE
...
-            <img src={selectedRampColorUrl} />
+            {
+              // Switch between categorical palette and color map as needed
+              props.useCategoricalPalettes ? (
+                <div
+                  className={styles.categoricalColor}
+                  style={getColorPaletteStyling(props.selectedPalette.slice(0, props.numCategories))}
+                ></div>
+              ) : (
+                <img src={selectedRampColorUrl} />
```
toloudis, line 211: slightly surprised that you can't use the same visual representation for both!

src/components/CategoricalColorPicker.tsx
```diff
@@ -0,0 +1,89 @@
...
+    // Text label, hide overflow as ellipsis
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+`;
+
+export default function CategoricalColorPicker(inputProps: CategoricalColorPickerProps): ReactElement {
+  const props = { ...defaultProps, ...inputProps } as Required<CategoricalColorPickerProps>;
+  const categories = props.dataset?.getFeatureCategories(props.featureName);
```
toloudis, line 54: would there be any advantage to using categories as a prop instead of dataset+featurename?  The advantage I think of is that this component doesn't have to know anything about datasets but I suppose it doesn't matter too much.

ShrimpCryptid: Hmm, I can only think of future benefits if we add more tooltip data or something. Changing this to `categories` makes sense to me.

src/components/ColorRampDropdown.tsx
ShrimpCryptid: Hmm... I *could* make it work if I changed `ColorRamp` to support hard stop gradients, since currently it assumes all gradients have equally spaced color stops. Let me noodle on this one...

src/components/ColorRampDropdown.module.css
ShrimpCryptid: image

I did but I wasn't a huge fan? It feels like it adds a lot of visual noise.

toloudis (approved):
LGTM, I think all my comments are minor things

src/components/CategoricalColorPicker.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { ReactElement, useMemo } from "react";
+import { Dataset } from "../colorizer";
+import { ColorPicker, Tooltip } from "antd";
+import { Color, ColorRepresentation } from "three";
+
+import { Color as AntColor } from "antd/es/color-picker/color";
+import styled from "styled-components";
```
meganrm, line 7: nit: it seems like you've grouped these by 3rd party and local, but this and `Dataset` are in the wrong place

ShrimpCryptid: Thanks, good catch! I need to set up prettier to do this for me 😂

src/components/ColorRampDropdown.tsx
ShrimpCryptid: I was able to make this work, though I need to do some more cleanup in the meantime and I'll mark this as a draft until it's ready to look at again.

ShrimpCryptid: Done!

src/components/ColorRampDropdown.tsx
```diff
@@ -58,46 +84,89 @@ const ColorRampSelector: React.FC<ColorRampSelectorProps> = (propsInput): ReactE
...
+    // Generate color ramps from the palettes.
+    const paletteData = Array.from(props.categoricalPalettes.values());
+    // Append in the missing ColorRamp
+    const colorRampData = paletteData.map((data) => {
+      const visibleColors = data.colors.slice(0, Math.max(1, props.numCategories));
+      return { ...data, colorRamp: new ColorRamp(visibleColors, ColorRampType.HARD_STOP) };
+    });
+
+    return makeRampButtonList(colorRampData, onClick);
+  }, [props.categoricalPalettes, props.numCategories]);
```
toloudis, line 160: Nit: Did you consider combining paletteDropdownContents and rampDropdownContents into the same object (dropdownContents)? ...Since the component is always only in one mode or the other, based on props.

toloudis (approved):

src/components/ColorRampDropdown.tsx
ShrimpCryptid: Originally I was going to say calculating them separately saved some computation on each render, but it turns out that the way I was passing in categorical data made that untrue 😅 Did some cleanup on this file!

src/components/ColorRampDropdown.tsx
```diff
@@ -1,25 +1,51 @@
...
 const defaultProps: Partial<ColorRampSelectorProps> = {
   colorRamps: DEFAULT_COLOR_RAMPS,
   disabled: false,
+  useCategoricalPalettes: false,
+  categoricalPalettes: DEFAULT_CATEGORICAL_PALETTES,
 };
 
+/**
+ * Returns whether if the two arrays are equal, where arr1[i] === arr2[i] for all i.
+ */
```
frasercl, line 36: whether if? can't tell if this is an error or if I'm bad at parsing sentences

src/components/ColorRampDropdown.tsx
```diff
@@ -58,46 +84,93 @@ const ColorRampSelector: React.FC<ColorRampSelectorProps> = (propsInput): ReactE
...
+  // Swap between two image sources as needed
+  const dropdownButtonImgSrc = props.useCategoricalPalettes ? paletteImgSrc : rampImgSrc;
+
   const buttonDivClassName = styles.buttonContainer + (props.disabled ? ` ${styles.disabled}` : "");
-  const dropdownContainerClassName = styles.dropdownContainer + (forceOpen ? ` ${styles.forceOpen}` : "");
+
+  const dropdownClassNames = [styles.dropdownContainer];
+  props.useCategoricalPalettes && dropdownClassNames.push(styles.categorical);
+  forceOpen && dropdownClassNames.push(styles.forceOpen);
+  const dropdownContainerClassName = dropdownClassNames.join(" ");
```
frasercl, line 173: Idk how I feel about using lazy `&&` purely for a side effect like this, rather than to produce a value. I get why it looks a lot nicer in this context... but I'd encourage `if` here.

frasercl (approved):
All my comments are pretty minor - this looks good!

src/colorizer/ColorRamp.ts
```diff
@@ -1,15 +1,27 @@
-import { DataTexture, Color, ColorRepresentation, RGBAFormat, FloatType, LinearFilter } from "three";
+import { DataTexture, Color, ColorRepresentation, RGBAFormat, FloatType, LinearFilter, NearestFilter } from "three";
+
+export enum ColorRampType {
+  LINEAR,
+  HARD_STOP,
+}
```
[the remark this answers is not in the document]
ShrimpCryptid: I struggled constantly between these being enums or booleans 😂
I don't *think* we'll have more gradients types in the future. I did like the enum here because it's easier to read imo.

src/components/ColorRampDropdown.tsx
ShrimpCryptid: Nope that's absolutely a grammar error from me rewriting this method comment. Thanks for catching this! :X

PR #157, opened 2023-12-09
fix: Use placeholders for unknown bounds

### LINKED ISSUE #134: Incorrect range when loading thresholds not in dataset from URL
## Description

- If the user opens a URL that has thresholded features not in the current dataset, the range is shown as [0, 1] instead of the original range.

- There should be a fallback in the range slider if a min and max bound for the slider can't be found.
  - ex: `--`

ShrimpCryptid:
Problem
=======
Fixes #134, [Incorrect range when loading thresholds not in dataset from URL](https://github.com/allen-cell-animated/nucmorph-colorizer/issues/134).

Solution
========
- LabeledRangeSlider defaults to showing a `--` label when a bound is NaN.
- Changed default feature range to `[NaN, NaN]` for features from datasets that haven't been loaded yet.

## Type of change

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-157/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Fcollection.json&dataset=Test%203%20Custom%20Manifest&feature=feature&filters=feature%3Atest%2520unit%3A0%3A140%2Cfeature%3A%3A35%3A140&range=35%2C140&color=matplotlib-cool
1. Switch to the filters tab. Observe the new placeholders.
1. Change datasets to Test 1. The range values should update to their correct values.

Screenshots (optional):
-----------------------
Before: (note how bounds go from 0 to 1, even though this doesn't represent the selected range)
image

After:
image

### DIFF
--- a/src/components/LabeledRangeSlider.tsx
+++ b/src/components/LabeledRangeSlider.tsx
@@ -159,6 +159,14 @@ export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps):
     });
   }
 
+  // Use a placeholder ("---") if the min/max bounds are undefined
+  const minSliderLabel = Number.isNaN(props.minSliderBound)
+    ? "--"
+    : numberToStringDecimal(props.minSliderBound, props.maxDecimalsToDisplay);
+  const maxSliderLabel = Number.isNaN(props.maxSliderBound)
+    ? "--"
+    : numberToStringDecimal(props.maxSliderBound, props.maxDecimalsToDisplay);
+
   return (
     <ComponentContainer>
       <InputNumber
@@ -190,12 +198,8 @@ export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps):
             open: props.disabled ? false : undefined, // Hide tooltip when disabled
           }}
         />
-        <SliderLabel $disabled={props.disabled}>
-          {numberToStringDecimal(props.minSliderBound, props.maxDecimalsToDisplay)}
-        </SliderLabel>
-        <SliderLabel $disabled={props.disabled}>
-          {numberToStringDecimal(props.maxSliderBound, props.maxDecimalsToDisplay)}
-        </SliderLabel>
+        <SliderLabel $disabled={props.disabled}>{minSliderLabel}</SliderLabel>
+        <SliderLabel $disabled={props.disabled}>{maxSliderLabel}</SliderLabel>
       </SliderContainer>
       <InputNumber
         ref={maxInput}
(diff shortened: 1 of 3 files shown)


src/components/LabeledRangeSlider.tsx
```diff
@@ -159,6 +159,14 @@ export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps):
...
   }
 
+  // Use a placeholder if the min/max bounds are undefined
+  const minSliderLabel = Number.isNaN(props.minSliderBound)
+    ? "--"
+    : numberToStringDecimal(props.minSliderBound, props.maxDecimalsToDisplay);
+  const maxSliderLabel = Number.isNaN(props.maxSliderBound)
+    ? "--"
+    : numberToStringDecimal(props.maxSliderBound, props.maxDecimalsToDisplay);
+
```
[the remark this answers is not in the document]
ShrimpCryptid: No strong reason, so I made a fix that uses `Number.NaN` as the default max and min values of the range sliders.

frasercl (approved):

src/components/LabeledRangeSlider.tsx
```diff
@@ -105,7 +106,7 @@ const SliderLabel = styled.p<{ $disabled?: boolean }>`
  * separately from the min and max value bounds and acts as a suggested range.
  */
 export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps): ReactElement {
-  const props = { ...defaultProps, ...inputProps } as Required<LabeledRangeSliderProps>;
+  const props = { ...defaultProps, ...excludeUndefinedValues(inputProps) } as Required<LabeledRangeSliderProps>;
```
toloudis, line 109: question: does this imply that you can't force a value to be explicitly undefined through inputProps?

toloudis (approved):
LGTM!  just one q about the excludeUndefinedValues thing

src/components/LabeledRangeSlider.tsx
ShrimpCryptid: Good question! Yes, at least for this instance. I was having an issue where the default values for `minSliderBounds` and `maxSliderBounds` were getting overridden by `undefined` input props.

ShrimpCryptid: I could probably add in a manual check for those two properties, but it seemed like that might cause problems elsewhere too.

PR #158, opened 2023-12-10
fix: Adds support for ordered feature arrays

### LINKED ISSUE #155: Features are out of order
## Description

Introduced by changes in #151...
When feature dictionaries JSON objects are parsed, their order is not guaranteed.

To fix, change `features` in manifest to be an array of objects.

ShrimpCryptid:
Problem
=======
Fixes #155, "Features are out of order." 

Solution
========
- Updates the manifest definition to accept `features` as an array, rather than a map with no guarantee of ordering.
- Changes `Dataset.ts` to use a `Map` internally, which enforces key ordering.
  - (technically JS objects _should_ enforce key ordering but it's not guaranteed.)
- Adds tests to accept both the outdated and new datasets.
  - Note: I don't think any datasets have actually been created with the last format (V2). Should I just remove the definitions for it altogether?

## Type of change

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -11,7 +11,6 @@ import { FeatureArrayType, FeatureDataType } from "./types";
 import * as urlUtils from "./utils/url_utils";
 import { MAX_FEATURE_CATEGORIES } from "../constants";
 import { AnyManifestFile, ManifestFile, ManifestFileMetadata, update_manifest_version } from "./utils/dataset_utils";
-import { RecordValue } from "./utils/type_utils";
 
 export enum FeatureType {
   CONTINUOUS = "continuous",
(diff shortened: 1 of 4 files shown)


src/colorizer/Dataset.ts
```diff
@@ -137,27 +139,26 @@ export default class Dataset {
...
+      {
+        tex: source.getTexture(FeatureDataType.F32),
+        data: source.getBuffer(FeatureDataType.F32),
+        min: source.getMin(),
+        max: source.getMax(),
+        units: metadata?.units || "",
+        type: featureType,
+        categories: featureCategories || null,
+      },
+    ];
```
ShrimpCryptid, line 153: This was the cause of the feature sorting order bug. It was originally inserting them whenever the async load operation finished, which meant that each completed at different times. Now, each promise returns the resulting array, allowing them to be inserted in the correct order.

PR #162, opened 2023-12-14
refactor: Sort Imports

ShrimpCryptid:
Problem
=======
Imports are out of order in some files, and don't follow JS/TS conventions

Solution
========
- Adds a plugin to prettier (`.prettierrc`) that automatically formats file imports.
- Adds a rule to eslint `(.eslintrc`) that warns if imports are not ordered correctly.
- Adds a new `npm run lint-fix` command that auto-fixes all file imports
- Sorted imports. So many sorted imports.

Keyfiles:
-----------------------
1. `package.json`
2. `.eslintrc`
3. `.prettierrc`


meganrm:
I would love to have this *somewhere* to have it auto added to other projects. But our [template](https://github.com/allen-cell-animated/github-boilerplate/tree/master/.github) is just GitHub stuff so I'm not sure where

meganrm (approved):

src/colorizer/types.ts
ShrimpCryptid, line 1: A bugfix that I slipped in, where I moved these imports from `ColorizeCanvas` to `types`. This ensures `types.ts` doesn't depend on any other internal modules, and can be safely loaded by other modules in `colorizer`.

.prettierrc
```diff
@@ -3,5 +3,8 @@
   "jsxBracketSameLine": false,
   "printWidth": 120,
   "trailingComma": "es5",
-  "tabWidth": 2
+  "tabWidth": 2,
+  "importOrder": ["^@core/(.*)$", "^@server/(.*)$", "^@ui/(.*)$", "^[./]"],
```
toloudis, line 7: can remove core, server, and ui ?

package.json
```diff
@@ -13,11 +13,13 @@
     "preview": "vite preview",
     "deploy": "vite build --base=/nucmorph-colorizer/ && gh-pages -d dist",
     "lint": "eslint --config ./.eslintrc --ext .ts --ext .tsx --ext .js --ext .jsx .",
+    "lint-fix": "eslint --rule \"simple-import-sort/imports: warn\" --rule \"simple-import-sort/exports: warn\" --fix --config ./.eslintrc --ext .ts --ext .tsx --ext .js --ext .jsx .",
```
toloudis, line 16: does this mean that lint-fix will ONLY do the import resorting?

ShrimpCryptid: I think I had the rules disabled in linting for a sec there, because prettier and the eslint plugin were fighting. BUT I finally figured out the setting to get them to play nice! Now this does the same thing as regular linting, just adding the `--fix` flag.

.prettierrc
```diff
@@ -4,7 +4,8 @@
   "printWidth": 120,
   "trailingComma": "es5",
   "tabWidth": 2,
-  "importOrder": ["^@core/(.*)$", "^@server/(.*)$", "^@ui/(.*)$", "^[./]"],
+  "importOrder": ["^[./]"],
   "importOrderSeparation": true,
-  "importOrderSortSpecifiers": true
+  "importOrderSortSpecifiers": true,
+  "importOrderCaseInsensitive": true
```
toloudis, line 10: one question about case insensitive, maybe a nit: If we have a file naming convention designed to separate react components from utility / business logic files based on casing, are you sure you want to make the sorting case insensitive?

toloudis (approved):
LGTM!

.prettierrc
ShrimpCryptid: So I tried implementing this by changing the sorting to separate lowercase and uppercase filenames, but it breaks down in a few places based on our naming conventions.


Good lowercase + uppercase sorting:
```
// utilities
import { getColorMap, getInRangeLUT, thresholdMatchFinder, validateThresholds } from "./colorizer/utils/data_utils";
import { numberToStringDecimal } from "./colorizer/utils/math_utils";

// classes
import Collection from "./colorizer/Collection";
import Export from "./components/Export";
```

Bad Lowercase sorting:
```
// utilities
import { ColorizeCanvas, Dataset, Track } from "./colorizer";
import { FeatureThresholdsTab, PlotTab, ScatterPlotTab, SettingsTab, TabType } from "./components/tabs";
```

Basically, this starts to break down anytime modules have been introduced. It raises the question of whether these module names should be capitalized if they mostly include business logic.

ShrimpCryptid: (Oh and just to add, afaik the case insensitive sorting is only applied within imports per module.)

ShrimpCryptid: Okay, added this and refactored a few submodule names. It's not perfect, but I do like the new separation between components and logic.

Issue #165, opened 2023-12-15
Customize outline and path colors


ShrimpCryptid:
## Use Case
- Default pink outline color can be hard to make out against certain backgrounds or color ramps.

## Acceptance Criteria
- Color pickers for outlines and path in UI
- Reasonable defaults for path + outline, a few bright/fluorescent defaults.

## Details
- Also allow users to set background color?

ShrimpCryptid:
Also, fix inconsistently thin outlines on high resolution dataset images.

ShrimpCryptid:
Adding that @chantelleleveille also encountered issues with this. She also suggested adding a default color for each color map that contrasts with it.

Issue #164, opened 2023-12-15
Export menu and viewer settings alignment


ShrimpCryptid:
![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/a216ec8c-4b11-4ebb-a3f0-e10680380351)

- Use CSS grid/subgrid to align text and value fields in the export menu and the settings menu.
- Text should be right-aligned, values centered or left-aligned.

References: 
![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/2a9f1425-c513-44ad-933e-7a90bbdc46f4)
![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/9d990a8d-809a-47a2-b60b-ef94d37962d5)
![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/e8193d08-b7df-4aa4-8b81-121c93fef88e)

PR #168, opened 2023-12-18
feat: Background images

### LINKED ISSUE #66: Provide background images for compositing
## Use Case

From user:

If I made the track overlays into gifs for each position could they be set as background channel for the plot instead of the gray?  Would be great if background channel that is displayed was configurable so these plots could be used for validation of nuc and cell segmentation.
For nuc segmentation validation I would use the nuc channel alone.  For CAAX seg validation I would use the nuc & caax channel as a 2 color RGB image.
Making it configurable would increase the use case tremendously

## Acceptance Criteria

Prototype Idea: provide a set of background frames as pngs and composite the main view on top of them.

## Details

The user may have multiple backdrop sets they want to try out.  Probably need a blend-type parameter too.

ShrimpCryptid:
Problem
=======
Adds support for background images; closes #66.

Solution
========
*Changes are grouped into these four buckets. See the Key files section for suggested review grouping of files.*
- Updates the dataset format to accept an array of backdrop image paths.
- Adds uniforms to the shader for passing in the backdrop image texture, a saturation value, and an opacity slider.
- Adds new controls in `SettingsTab.tsx`.
- Fixes to `TimeControls.tsx` to prevent jumps and skipping during playback.

Future TODOs:
- Add caching for backdrop images (see #169)
- Adjust settings tab layout to use subgrid (see #164 and https://github.com/allen-cell-animated/nucmorph-colorizer/pull/172)
- Change overlay from dark/light mode depending on background color

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-168/?dataset=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2F3500005828_25%2Fmanifest.json&feature=Adjacent%20neighbors&color=matplotlib-cool
  - Also this one: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-168/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Ffeature%2Fbackground-images%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=Continuous%20Feature&color=matplotlib-cool 
3. Go to the settings tab and the Backdrop section.
4. Click and change to a different backdrop image.
5. Change the brightness slider on the backdrop.
6. Change the opacity slider on objects.
7. Change the saturation slider on the backdrop.
8. Switch to a different backdrop image.

Screenshots (optional):
-----------------------
image

**Example exported video:**

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/501cdb26-fb70-4a8a-9bfd-07f368b3d75b

**Talkthrough: 🔊**
(Since this video I've made some changes to how brightness works-- the behavior is a little more extreme, but it makes it behave as expected on backdrops that are very dark.)

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/e767af47-9d0a-4240-bb8a-d7f74f1eb8bf

-----------------------
Shader Changes:
  1. `colorize_RGBA8U.frag`
  1. `ColorizeCanvas.tsx` 
 
Dataset:
  1. `dataset_utils.ts`
  3. `Dataset.ts` 

UI Changes:
  1. `SettingsTab.tsx`
  1. `App.tsx` 
  1. `CanvasWrapper.tsx`
 
Time Playback
  1. `TimeControls.tsx`


src/colorizer/CanvasUIOverlay.ts
ShrimpCryptid, line 1: Renamed `BackgroundOptions` to `OverlayBackgroundOptions` to avoid confusion with the background images feature. There's probably a better name for this, it's the transparent background fill used to add contrast with the background.

image

src/colorizer/ColorizeCanvas.ts
```diff
@@ -522,22 +530,65 @@ export default class ColorizeCanvas {
...
+   * @param forceUpdate Force a reload of the frame data, even if the frame
+   * is already loaded.
    */
-  async setFrame(index: number): Promise<void> {
+  async setFrame(index: number, forceUpdate: boolean = false): Promise<void> {
     // Ignore same or bad frame indices
-    if (this.currentFrame === index || !this.isValidFrame(index)) {
+    if ((!forceUpdate && this.currentFrame === index) || !this.isValidFrame(index)) {
       return;
     }
```
ShrimpCryptid, line 566: Add an option to force reload. This is used when switching backdrops, where we need to reload the current frame to get the backdrop image.

src/colorizer/TimeControls.ts
ShrimpCryptid, line 1: Changed TimeControls to stop using `setInterval`. Instead, it waits for frames to be loaded and optionally adds a delay so it reaches a minimum of the set FPS between frames. This prevents frames from being skipped and reduces frame jitter.

ShrimpCryptid:
Open question for @toloudis about the best way to handle brightness adjustments...

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -64,7 +112,9 @@ bool isEdge(vec2 uv, ivec2 frameDims) {
   float thickness = 2.0;
   float wStep = 1.0 / float(frameDims.x);
   float hStep = 1.0 / float(frameDims.y);        
-    // sample around the pixel to see if we are on an edge
+  // sample around the pixel to see if we are on an edge
+  // TODO: Fix this so it samples using canvas pixel offsets instead of frame pixel offsets.
+  // Currently, the edge detection is sparser when loading high-resolution frames.
```
ShrimpCryptid, line 117: Will likely address this when I fix issue #165.

src/colorizer/loaders/ImageFrameLoader.ts
ShrimpCryptid, line 1: Updated to allow loading of float and unsigned integer textures. See `Dataset.ts` for usage.

interim17:
Probably not related to this PR, but I noticed that the slider behavior is a bit funny when clicked during playback.

The selected frame will flash briefly, the number will show on the input, and the slider icon moves, but then playback resumes wherever it was playing from before. If its intentional to prevent hopping around via slider during playback maybe it should be disabled?

ShrimpCryptid:
> Probably not related to this PR, but I noticed that the slider behavior is a bit funny when clicked during playback.
> 
> The selected frame will flash briefly, the number will show on the input, and the slider icon moves, but then playback resumes wherever it was playing from before. If its intentional to prevent hopping around via slider during playback maybe it should be disabled?

Hmm, thanks for noting this. I think the slider has always (mis)behaved during playback. Disabling it is a good solution, but now I'm starting to wonder if I should have playback stop temporarily while the user is interacting with the slider and then resume once released...

interim17:
> > Probably not related to this PR, but I noticed that the slider behavior is a bit funny when clicked during playback.
> > The selected frame will flash briefly, the number will show on the input, and the slider icon moves, but then playback resumes wherever it was playing from before. If its intentional to prevent hopping around via slider during playback maybe it should be disabled?
> 
> Hmm, thanks for noting this. I think the slider has always (mis)behaved during playback. Disabling it is a good solution, but now I'm starting to wonder if I should have playback stop temporarily while the user is interacting with the slider and then resume once released...

I believe that's what Simularium does, call pause during click or mouse down event, then resume at new frame.

src/App.tsx
```diff
@@ -62,6 +56,11 @@ function App(): ReactElement {
   const [selectedTrack, setSelectedTrack] = useState<Track | null>(null);
   const [currentFrame, setCurrentFrame] = useState<number>(0);
 
+  const [backdropKey, setBackdropKey] = useState<string | null>(null);
```
toloudis, line 59: at first glance when reading the code it's hard to tell what this "key" is.  I wonder if `selectedBackdropKey` or just put a comment up here where it's set up, like "the key for the current choice of backdrop images".   It's not necessarily intuitive that there would be multiple possible backdrops.

src/colorizer/CanvasUIOverlay.ts
```diff
@@ -23,7 +23,7 @@ type TimestampOptions = StyleOptions & {
   startTimeSec: number;
 };
 
-type BackgroundOptions = {
+type OverlayBackgroundOptions = {
```
toloudis, line 26: Nit: "overlay background" seems like a contradiction

src/colorizer/ColorizeCanvas.ts
```diff
@@ -85,7 +89,8 @@ const getDefaultUniforms = (): ColorizeUniforms => {
   const emptyOutliers = packDataTexture([0], FeatureDataType.U8);
   const emptyInRangeIds = packDataTexture([0], FeatureDataType.U8);
   const emptyColorRamp = new ColorRamp(["black"]).texture;
-  const emptyOverlay = new DataTexture(new Uint8Array([0, 0, 0, 0]), 1, 1, RGBAIntegerFormat, UnsignedByteType);
+  const emptyOverlay = new DataTexture(new Uint8Array([0, 0, 0, 0]), 1, 1, RGBAFormat, UnsignedByteType);
+  const emptyBackdrop = new DataTexture(new Uint8Array([1, 0, 0, 0]), 1, 1, RGBAFormat, UnsignedByteType);
```
toloudis, line 93: again kind of hard to tell what "overlay" means here. does the overlay mean the main colorized image?

toloudis: I know what would help, is if `emptyFrame` were also right next to the lines for emptyOverlay and emptyBackdrop. :)

src/colorizer/ColorizeCanvas.ts
```diff
@@ -503,6 +510,7 @@ export default class ColorizeCanvas {
     }
     // Save the array to a texture and pass it into the shader
     this.setUniform("inRangeIds", packDataTexture(Array.from(inRangeIds), FeatureDataType.U8));
+    this.render();
```
toloudis, line 513: is this fixing a bug or is it really related to the current changes?

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: The overlay is the extra UI overlaid on top of the main view, like the timestamp and scalebar.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -529,22 +537,63 @@ export default class ColorizeCanvas {
...
     }
     // New frame, so load the frame data.
     this.currentFrame = index;
-    const frame = await this.dataset?.loadFrame(index);
+    let backdropPromise = undefined;
+    if (this.backdropKey && this.dataset?.hasBackdrop(this.backdropKey)) {
+      backdropPromise = await this.dataset?.loadBackdrop(this.backdropKey, index);
+    }
+    const framePromise = await this.dataset?.loadFrame(index);
+    const result = await Promise.all([framePromise, backdropPromise]);
```
toloudis, line 581: It looks like this result might be unnecessary because you already awaited both calls?

src/colorizer/ColorizeCanvas.ts
```diff
@@ -529,22 +537,63 @@ export default class ColorizeCanvas {
...
     if (!frame) {
       return;
     }
+    if (this.currentFrame !== index) {
+      // This load request has been superceded by a request for another frame, which has already loaded in image data.
+      // Drop this request.
+      return;
+    }
+    if (overlay) {
+      this.setUniform("backdrop", overlay);
```
toloudis, line 593: semantically this looks wrong.  a backdrop is the opposite of an overlay :)

ShrimpCryptid:
> > > Probably not related to this PR, but I noticed that the slider behavior is a bit funny when clicked during playback.
> > > The selected frame will flash briefly, the number will show on the input, and the slider icon moves, but then playback resumes wherever it was playing from before. If its intentional to prevent hopping around via slider during playback maybe it should be disabled?
> > 
> > 
> > Hmm, thanks for noting this. I think the slider has always (mis)behaved during playback. Disabling it is a good solution, but now I'm starting to wonder if I should have playback stop temporarily while the user is interacting with the slider and then resume once released...
> 
> I believe that's what Simularium does, call pause during click or mouse down event, then resume at new frame.

I have a semi-working fix, but there's a few edge cases I'm running into that will take some extra logic to address. I've disabled the time slider for now when playback is happening, and I'm going to spin this into a separate PR.

src/colorizer/Dataset.ts
```diff
@@ -298,6 +334,17 @@ export default class Dataset {
     this.timesFile = manifest.times;
     this.centroidsFile = manifest.centroids;
 
+    if (manifest.backdrops) {
+      for (const { name, key, frames } of manifest.backdrops) {
+        this.backdropData.set(key, { name, frames });
+        if (frames.length !== this.frameFiles.length || 0) {
+          console.warn(
```
toloudis, line 341: wondering if this could be promoted to error.  if they are not matched then playback will have rejected promises when trying to load.

src/colorizer/TimeControls.ts
```diff
@@ -44,24 +38,46 @@ export default class TimeControls {
...
+      // do the update
       if (this.setFrameFn) {
-        this.setFrameFn(nextFrame);
+        await this.setFrameFn(nextFrame);
       }
+      const endTime = Date.now();
+      const timeElapsed = endTime - startTime;
       onNewFrameCallback();
+
+      // TODO: Could add some sort of smoothing here to make the playback more consistent.
```
toloudis, line 74: there is definitely something sketchy about having awaits in the middle of the redraw/rendering call.  might merit some thinking in a future ticket.

src/colorizer/TimeControls.ts
```diff
@@ -72,28 +88,22 @@ export default class TimeControls {
   }
 
   public handlePauseButtonClick(): void {
-    clearInterval(this.timerId);
-    this.timerId = DEFAULT_TIMER_ID;
+    if (this.timerId !== DEFAULT_TIMER_ID) {
+      clearTimeout(this.timerId);
+      this.timerId = DEFAULT_TIMER_ID;
```
toloudis, line 93: Nit: `DEFAULT_TIMER_ID` might be more properly named `NO_TIMER_ID` !

src/colorizer/TimeControls.ts
```diff
@@ -72,28 +88,22 @@ export default class TimeControls {
...
-    if (disabled) {
-      clearInterval(this.timerId);
-      this.timerId = DEFAULT_TIMER_ID;
+      await this.setFrameFn(this.wrapFrame(this.canvas.getCurrentFrame() + delta));
     }
   }
 
   public isPlaying(): boolean {
-    return this.timerId !== -1;
+    return this.currentlyPlaying;
```
toloudis, line 106: is currentlyPlaying always in sync with (timerId !== -1) ?

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -87,25 +156,55 @@ vec4 alphaBlend(vec4 a, vec4 b) {
...
+
+vec4 getBackdropColor(vec2 sUv) {
+  if (isOutsideBounds(sUv)) {
+    return TRANSPARENT;
+  }
+  vec4 backdropColor = texture(backdrop, sUv).rgba;
+  // Switch between HSV and HSL color adjustments based on whether we are increasing or
+  // decreasing brightness. HSL is better for increasing brightness, since it does not increase
+  // perceptual saturation when lightness increases (the way HSV does when value increases).
+  // HSV is better than HSL for decreasing brightness for similar reasons.
```
toloudis, line 171: Interesting. I'm surprised that HSV is not good enough

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -135,9 +234,17 @@ vec4 getMainPixelColor() {
...
+  vec4 mainColor = getObjectColor(sUv);
+  mainColor.a *= objectOpacity;
+
   vec4 overlayColor = texture(overlay, vUv).rgba;  // Unscaled UVs, because it is sized to the canvas
 
-  gOutputColor = alphaBlend(overlayColor, mainColor);
+  gOutputColor = vec4(backgroundColor, 1.0);
+  gOutputColor = alphaBlend(backdropColor, gOutputColor);
+  gOutputColor = alphaBlend(mainColor, gOutputColor);
+  gOutputColor = alphaBlend(overlayColor, gOutputColor);
```
toloudis, line 249: this is nice, I like how it reads.

src/colorizer/shaders/colorize_RGBA8U.frag
ShrimpCryptid: This may be a symptom of me using value/lightness as a proxy for brightness and not looking for an actual algorithmic approach 😅

image

If we're increasing lightness/value, then HSV causes colors to become unexpectedly more saturated.

image

When decreasing lightness/value, HSL can sometimes cause colors to become more saturated, which is most evident for very light, off-white colors.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Oops! Good catch, thanks. Removed the two `awaits` for each promise so they can run in parallel.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: Embarrassed that I didn't catch this! I was thrashing a bit on whether to name the background images an `overlay` or `backdrop`, since I had briefly considered letting them sort both in front of and behind the main view.

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: It's a bugfix! Before, the canvas thresholding was one update behind thresholding on the UI, but it was less noticeable because we were dealing with continuous numeric values and not categories.

src/colorizer/shaders/colorize_RGBA8U.frag
ShrimpCryptid: I did also find this algorithm, which is what GIMP uses: https://en.wikipedia.org/wiki/Image_editing#Contrast_change_and_brightening

src/colorizer/TimeControls.ts
ShrimpCryptid: It wasn't guaranteed before, but I made some changes and removed `currentlyPlaying`.

src/colorizer/shaders/colorize_RGBA8U.frag
toloudis: that algorithm looks so much simpler, I kinda like that. just per-rgb without changing color space.  if it works for gimp.... ?  From the pictures you show, it seems clear that L and V are not quite right and I can see why you do the conditional.  From those pictures, L looks much more appropriate than V as a brightness control.  But since the gimp code is so simple, you could try it and see if it feels right.

src/colorizer/TimeControls.ts
```diff
@@ -71,12 +67,18 @@ export default class TimeControls {
       const timeElapsed = endTime - startTime;
       onNewFrameCallback();
 
-      // TODO: Could add some sort of smoothing here to make the playback more consistent.
+      if (this.timerId === NO_TIMER_ID) {
```
toloudis, line 70: nit / DRY: all these conditional checks could now use isPlaying() :shrug:

src/colorizer/TimeControls.ts
toloudis: looks good

src/colorizer/TimeControls.ts
ShrimpCryptid: True! I sort of like having this here as a reminder that `isPlaying` relies on `this.timerId`... but it's probably cleaner to use `isPlaying` directly

toloudis (approved):
LGTM! can't wait to see this in use

src/colorizer/shaders/colorize_RGBA8U.frag
ShrimpCryptid: It works!

interim17:
Works well on my end and seems like a really useful feature.

interim17 (approved):

PR #170, opened 2023-12-20
feat: Categorical thresholding

ShrimpCryptid:
Problem
=======
Continues work on #148, implementing thresholding on categorical features.

Solution
========
This change is large enough that I'm planning to do a code walkthrough with any interested parties in the new year!

- Updates `ColorizeCanvas` to filter using both categorical and numeric thresholds.
- Updates the `FeatureThreshold` type to handle both categorical and numeric (continuous/discrete number) features.

URL Changes:
- `url_utils.ts`: Adds support for serializing/deserializing categorical thresholds to the URL. Selected categorical thresholds are saved as hex strings.

UI Changes:
- Rendering for categorical features in the thresholds panel.
- Adds scroll shadows to the thresholds panel, based on designs from Lyndsay.
- Fixes an inconsistent layout bug with the side panel when browser width changes.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-170/?dataset=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Fnucmorph-colorizer%2Ffeature%2Fcategorical-thresholding%2Fdocumentation%2Fexample_dataset%2Fmanifest.json&feature=Continuous%20Feature&t=4&color=matplotlib-cool
2. Go to the filters tab and select one or more categorical features.
3. Change the Color map at the top. The checkboxes should update.
4. Enable/disable checkboxes and watch the matching objects be enabled/disabled in the main viewport.
5. Copy the URL and open a new tab. Your filters should remain constant.

Screenshots (optional):
-----------------------

image

Video Talkthrough (🔊)

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/5f9e9ebf-8cbc-4f73-acdf-1b5883a9b05b

-----------------------
1. `src/colorizer/types.ts` -> changes to `FeatureThreshold` type
6. `ColorizeCanvas.ts` -> filtering with the new `FeatureThresold` categorical type
7. `Dataset.ts` -> minor API change, `tryGetFeatureData`
8. `App.tsx`
9. `FeatureThresholdsTab.tsx` -> rendering of thresholds

Thanks for contributing!


src/App.tsx
```diff
@@ -288,7 +288,11 @@ function App(): ReactElement {
     }
     const setupInitialParameters = async (): Promise<void> => {
       if (initialUrlParams.thresholds) {
-        setFeatureThresholds(initialUrlParams.thresholds);
+        if (dataset) {
+          setFeatureThresholds(validateThresholds(dataset, initialUrlParams.thresholds));
```
ShrimpCryptid, line 292: `validateThresholds` is defined in `data_utils`, and fixes any thresholds that were labeled as categorical that should have been numeric and vice-versa. It's run both on the initial parameter load, which is happening here, and whenever the dataset is changed, which is below.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -486,21 +495,25 @@ export default class ColorizeCanvas {
...
+      for (const threshold of validThresholds) {
+        const featureData = this.dataset?.tryGetFeatureData(threshold.featureName);
+        if (featureData && !this.isValueWithinThreshold(featureData.data[id], threshold)) {
+          arr[id] = 0;
+          return;
         }
       }
-    }
+      arr[id] = 1;
+    });
```
ShrimpCryptid, line 515: This loop gives better performance in the best-case scenario (we can iterate over each `id` once instead of minimum `id * num_thresholds`) and is a little easier to read IMO.

src/App.module.css
ShrimpCryptid, line 1: Fixes an inconsistent layout issue in Chrome browsers, suggested by @frasercl!

src/colorizer/Dataset.ts
ShrimpCryptid, line 1: Fixes a loading issue when swapping datasets, caused by a change to the return types of `getFeatureData` in a previous PR.

src/components/CanvasWrapper.tsx
```diff
@@ -116,7 +116,10 @@ export default function CanvasWrapper(inputProps: CanvasWrapperProps): ReactElem
     // YAGNI: Debouncing for this is possible but no performance issues encountered yet.
     // Add only if needed.
     // Timeout in case of slowdowns to prevent this from halting the UI.
-    setTimeout(() => canv.setFeatureThresholds(props.featureThresholds), 0);
+    setTimeout(() => {
+      canv.setFeatureThresholds(props.featureThresholds);
+      canv.render();
```
ShrimpCryptid, line 121: Fixes a bug where changing the thresholds would not update the canvas view.

ShrimpCryptid:
Adding a note in case anyone notices the difference between the video and the preview, I removed colored checkboxes as requested by Lyndsay to reduce visual noise!

src/colorizer/ColorizeCanvas.ts
```diff
@@ -487,20 +496,24 @@ export default class ColorizeCanvas {
...
-      for (let i = 0, n = inRangeIds.length; i < n; i++) {
-        if (inRangeIds[i] === 1 && (featureData.data[i] < threshold.min || featureData.data[i] > threshold.max)) {
-          inRangeIds[i] = 0;
+    // Ignore thresholds with features that don't exist in this dataset or whose units don't match
+    const validThresholds = thresholds.filter((threshold) => {
+      const featureData = this.dataset?.tryGetFeatureData(threshold.featureName);
+      return featureData && featureData.units === threshold.units;
+    });
+
+    inRangeIds.forEach((_value, id, arr) => {
```
ShrimpCryptid, line 506: Replace with integer for loop

src/App.module.css
```diff
@@ -174,11 +183,8 @@
 .sidePanels {
   gap: 10px;
   flex-grow: 1;
-
-  /* TODO: This is copied from CanvasWrapper and should be shared sizing behavior. */
-  max-width: calc(min(100vw - 75px, 730px));
-  width: fit-content;
   min-width: 300px;
+  max-width: var(--content-width);
```
toloudis, line 187: naive css question: how does .sidePanels see the var that is scoped to .contentPanels?

ShrimpCryptid: Where it's used in the JSX, the `.sidePanels` div is a child of the `.contentPanels` div. CSS variables flow down from parents to children!

src/colorizer/ColorizeCanvas.ts
toloudis: If you consider this a performance sensitive part of the code, we have measured something like at least 30% perf improvement by using the old style integer for loops.
Consider changing the innermost loop to do the same if possible.

src/colorizer/Dataset.ts
```diff
@@ -147,7 +147,15 @@ export default class Dataset {
...
-  public getFeatureData(name: string): FeatureData {
+  /**
+   * Attempts to get the feature data from this dataset for the given feature name.
+   * Returns `undefined` if feature is not in the dataset.
+   */
+  public tryGetFeatureData(name: string): FeatureData | undefined {
+    return this.features.get(name);
+  }
+
+  private getFeatureData(name: string): FeatureData {
```
toloudis, line 158: Nit: something about this makes me think we should get rid of the old one and have the error check always be outside of the function?  I'm assuming this is here for cases where it's ok for the requested feature to be missing. 

I'm not sure about the naming - maybe something like `getFeatureDataOptional` is slightly better than putting `try` in the name of the function.  In reality, the old function is more like a "try" because it can throw something that would be caught.

A minor concern really.

src/colorizer/types.ts
```diff
@@ -74,9 +74,39 @@ export const featureTypeSpecs: { [T in FeatureDataType]: FeatureTypeSpec<T> } =
   },
 };
 
-export type FeatureThreshold = {
+// Similar to `FeatureType`, but indicates that thresholds are lossy when it comes
+// to numeric data. Thresholds do not track if their source feature is integer
```
toloudis, line 78:
```suggestion
// to numeric data. Numeric thresholds do not track if their source feature is integer
```

src/colorizer/utils/data_utils.ts
```diff
@@ -27,3 +28,45 @@ export function getColorMap(colorRampData: Map<string, ColorRampData>, key: stri
...
+  // Validate feature data for each threshold. If the threshold is the wrong type, update it.
+  const newThresholds: FeatureThreshold[] = [];
+
+  for (const threshold of thresholds) {
+    const featureData = dataset.tryGetFeatureData(threshold.featureName);
+    const isInDataset = featureData && featureData.units === threshold.units;
+
+    if (isInDataset && featureData.type === FeatureType.CATEGORICAL && isThresholdNumeric(threshold)) {
+      // Threshold is not categorical but the feature is.
+      // Convert the threshold to categorical.
```
toloudis, line 49: what is the situation when these conditions can happen?  maybe give one example of why we need to do this in the jsdoc comment above?  I assume it's basically when switching datasets and trying to keep same filter settings?

src/colorizer/utils/react_utils.ts
```diff
@@ -71,3 +71,91 @@ export function excludeUndefinedValues<T extends Object>(obj: T): Partial<T> {
...
+ * @returns an object with three properties:
+ * - scrollShadowStyle: a CSSProperties object that can be applied to the element to add a shadow. NOTE:
+ * This does not have to be the scrolling element; see the example for an overlay shadow.
+ * - onScrollHandler: an event handler to attach to the scrolling element's `onScroll` event.
+ * - scrollRef: a ref to attach to the scrolling element.
+ *
+ * @example
+ * ```
+ * function MyComponent() {
+ *   const { scrollShadowStyle, onScrollHandler, scrollRef } = useScrollWithShadow();
```
toloudis, line 91:
```suggestion
 *   const { scrollShadowStyle, onScrollHandler, scrollRef } = useScrollShadow();
```

src/colorizer/utils/url_utils.ts
```diff
@@ -50,6 +51,123 @@ export function fetchWithTimeout(
...
+ * The binary representation is `00101`, which is `0x05` in hex.
+ */
+function serializeThreshold(threshold: FeatureThreshold): string {
+  // featureName + units are encoded in case it contains special characters (":" or ",").
+  // TODO: remove once feature keys are implemented.
+  const featureName = encodeURIComponent(threshold.featureName);
+  const featureUnit = encodeURIComponent(threshold.units);
+
+  // TODO: Are there better characters I can be using here? ":" and "," take up
+  // more space in the URL. -> once features are converted to use keys, use "-" as a separator here? "|"?
```
toloudis, line 76: agree that using keys with restricted character ranges is the way to go. but it will be a breaking change for compatibility with old existing urls!

src/colorizer/utils/data_utils.ts
ShrimpCryptid: It's most likely to happen if users are opening links from older versions of Timelapse Colorizer (where categorical features are treated like numeric thresholds) or the case you mentioned, where one dataset labels a feature as categorical that another does not. I'll add an additional comment here.

tests/url_utils.test.ts
```diff
@@ -135,26 +161,26 @@ describe("Loading + saving from URL query strings", () => {
...
     const queryString = paramsToUrlQueryString(originalParams);
     const parsedParams = loadParamsFromUrlQueryString(queryString);
 
     expect(parsedParams.thresholds).deep.equals([
-      { featureName: "feature1", units: "m", min: 0, max: 1 },
-      { featureName: "feature2", units: "m", min: -34, max: 12 },
-      { featureName: "feature3", units: "m", min: 0.25, max: 0.5 },
+      { featureName: "feature1", units: "m", type: ThresholdType.NUMERIC, min: 0, max: 1 },
+      { featureName: "feature2", units: "m", type: ThresholdType.NUMERIC, min: -34, max: 12 },
+      { featureName: "feature3", units: "m", type: ThresholdType.NUMERIC, min: 0.25, max: 0.5 },
```
toloudis, line 176: Minor: I think it would be good if these tests that go to and from urlquerystring also tested against the expected generated query string. ( expect queryString to equal whatever )

ShrimpCryptid: Made this change! These are ALL going to break when I update how filter URLs are stored later 😅

toloudis (approved):
✅ ✅ 🔲  🔲  ✅

src/components/tabs/FeatureThresholdsTab.tsx
```diff
@@ -151,9 +210,23 @@ export default function FeatureThresholdsTab(inputProps: FeatureThresholdsTabPro
...
+    const threshold = newThresholds[index];
+    if (isThresholdCategorical(threshold)) {
+      threshold.enabledCategories = enabled_categories;
+    }
+    props.onChange(newThresholds);
+  };
+
+  /** Handle the threshold for a feature changing. */
+  const onNumericThresholdChanged = (index: number, min: number, max: number): void => {
     const newThresholds = [...props.featureThresholds];
```
frasercl, line 224: nit: two copies of the same doc comment

src/colorizer/utils/react_utils.ts
```diff
@@ -71,3 +71,92 @@ export function excludeUndefinedValues<T extends Object>(obj: T): Partial<T> {
...
+  useEffect(() => {
+    // Duplicates above handler to hide shadows even if the
+    // element stops being scrollable, or before first scroll
+    // interaction.
+    if (scrollRef.current) {
+      setScrollTop(scrollRef.current.scrollTop);
+      setScrollHeight(scrollRef.current.scrollHeight);
+      setClientHeight(scrollRef.current.clientHeight);
+    }
+  });
```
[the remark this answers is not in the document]
ShrimpCryptid: I was originally going to do this with `addEventListener`, but apparently ResizeObserver is more performant? I also found this hook, which I've added to the project, and includes a bunch of nice optimizations: https://www.npmjs.com/package/@react-hook/resize-observer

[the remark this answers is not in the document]
ShrimpCryptid: Nevermind, this actually doesn't work because resize doesn't check for changes to scroll height. I added in a MutationObserver instead!

frasercl (approved):
LGTM!

PR #172, opened 2023-12-22
feature: Settings layouts

### LINKED ISSUE #164: Export menu and viewer settings alignment

ShrimpCryptid:
Problem
=======
Closes #164, Export menu and viewer settings alignment.

Solution
========
- Adds alignment for settings panels via `SettingsItem` and `SettingsContainer`
  - Any settings object is a `label` with two children, a `span` and another element. The `span` is formatted on a grid column which allows the settings object to line up with other objects vertically. (see below for visual example).
- Adds collapsible headers for the main settings tab.
- Edits labels and label formatting for some settings for consistency across the app.

Note: I highly recommend **hiding whitespace** because I changed nesting depth in a few places.

## Type of change

Screenshots (optional):
-----------------------
Before:
image
After:
image

Example with columns:
image

Steps to test:
---------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-172
2. Look at the settings tab. Try collapsing and uncollapsing menus.
3. Open the Export modal and switch to a custom range and MP4 video export.
(Compare against https://allen-cell-animated.github.io/nucmorph-colorizer/main)


src/components/CustomCollapse.tsx
ShrimpCryptid, line 1: Convenience + styled class, lets you create a single collapsible header with the collapsed content as children.

toloudis:
@lynwilhelm can you help review this? I suspect there are some design tweaks that could make the ui more readable visually

ShrimpCryptid:
Moving this to draft; Lyndsay requested that I try to make the settings areas hug the left text label, rather than being a fixed width. I think it's possible with CSS grid but it'll take some big changes.

src/components/tabs/SettingsTab.tsx
```diff
@@ -69,90 +67,104 @@ export default function SettingsTab(props: SettingsTabProps): ReactElement {
...
+            <HiddenMarksSlider
+              style={{ maxWidth: "200px", width: "100%" }}
+              min={0}
+              max={100}
+              value={props.objectOpacity}
+              onChange={props.setObjectOpacity}
+            />
+          </label>
+          <label>
+            <span></span>
```
ShrimpCryptid, line 143: Dummy span allows checkbox to be aligned properly

ShrimpCryptid:
Changes were not as bad as I thought! Reopening this PR!

src/components/Export.tsx
```diff
@@ -497,67 +500,77 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
...
-                    options={videoQualityOptions}
-                    optionType="button"
-                    value={videoBitsPerSecond}
-                    onChange={(e) => setVideoBitsPerSecond(e.target.value)}
-                  />
-                  <p style={{ color: theme.color.text.hint }}>(~{getApproximateVideoFilesizeMb()})</p>
-                </HorizontalDiv>
-              </VerticalDiv>
+              <SettingsContainer>
+                <label>
```
toloudis, line 553: is there a particular reason why `label` is the main setting container element?

toloudis: ah, I see that I needed to read ahead to utils.tsx see how it's set up

src/components/tabs/SettingsTab.tsx
toloudis: is there a way to have a SettingsContainerCheckbox that automatically does this for you?  Maybe not super necessary for this PR but if you keep making more and more checkboxes...

toloudis: it's like you almost need some kind of component wrapper that abstracts the whole "Label plus Span plus input element" thing.  probably something to think about for the future.

toloudis (approved):
I like the changes here!  
Part of me thinks the `label` and `span` structure could be hidden in a component that takes the label string and input component as props (roughly speaking).  But maybe that would be extra layers of code that dont really help much.

ShrimpCryptid:
> I like the changes here! Part of me thinks the `label` and `span` structure could be hidden in a component that takes the label string and input component as props (roughly speaking). But maybe that would be extra layers of code that dont really help much.

Hmm, I do like this because it makes the system a bit more resilient against future changes if I ever need to change how settings layouts are organized. Thanks for the suggestion!

toloudis (approved):

interim17:
LGTM, useful for me to review these styled-components because I think I need to start using them in NBSV where we use antd 5.

Sim website still uses antd 4 but we may want to migrate to 5 or later at some point and will need to ditch LESS for these CSS-in-JS approaches.

interim17 (approved):

PR #174, opened 2023-12-22
refactor: Viewer config objects

### LINKED ISSUE #166: Refactor viewer settings into an object
## Use Case
image

- The list of props for the `SettingsTab` is getting VERY long. 
- Turn list of props into a settings object, with a setter for updating specific fields.

## Details

- May also be a good opportunity to think about local saving...

ShrimpCryptid:
NOTE: In draft until #172 is merged!

Problem
=======
Closes #166 ! Tackles the LARGE number of props being passed into the `SettingsTab` and `CanvasWrapper` componenents.

Solution
========
- Greatly simplifies passing viewer props by making a single `ViewerConfig` object type.
  - This is intended to only capture high-level viewer settings that likely don't need to change between datasets.
- Moves shared types out of `ColorizeCanvas`. 
- Updates imports and props where necessary.

Note: There are probably additional settings objects that could be created! A useful one would be a per-dataset settings object, but I'll leave this for a future refactor.

-----------------------
1. `types.ts`


src/colorizer/types.ts
```diff
@@ -110,3 +121,39 @@ export const isThresholdCategorical = (threshold: FeatureThreshold): threshold i
...
+  showTrackPath: true,
+  showScaleBar: true,
+  showTimestamp: true,
+  keepRangeBetweenDatasets: true,
+  /** Opacity, as a number percentage. */
+  backdropBrightness: 100,
+  /** Saturation, as a number percentage. */
+  backdropSaturation: 100,
+  /** Opacity, as a number percentage. */
+  objectOpacity: 100,
```
ShrimpCryptid, line 156: Arguably these can be per-dataset settings, but that's a change for another PR :')

src/App.tsx
```diff
@@ -869,24 +856,12 @@ function App(): ReactElement {
                     children: (
                       <div className={styles.tabContent}>
                         <SettingsTab
-                          // TODO: Refactor all of this into a settings or configuration object
-                          outOfRangeDrawSettings={outOfRangeDrawSettings}
-                          outlierDrawSettings={outlierDrawSettings}
-                          showScaleBar={showScaleBar}
-                          showTimestamp={showTimestamp}
+                          config={config}
+                          updateConfig={updateConfig}
```
toloudis, line 860: Is this a very standard practice to pass both the whole top-level state var and its updater function down to child components?  This looks like some kind of pattern or antipattern but I don't know which it is :)

toloudis (approved):
LGTM.  One question: I was looking for the code that serializes these to url, assuming it would have had to change.  Did I miss something?

ShrimpCryptid:
> LGTM. One question: I was looking for the code that serializes these to url, assuming it would have had to change. Did I miss something?

Good question! Currently, none of these settings are being serialized to the URL. There's the issue #182 for the scatterplot settings and I've opened #183 for the backdrop.

src/App.tsx
```diff
@@ -62,11 +61,14 @@ function App(): ReactElement {
...
-  const [backdropBrightness, setBackdropBrightness] = useState<number>(100);
-  const [backdropSaturation, setBackdropSaturation] = useState<number>(100);
-  const [objectOpacity, setObjectOpacity] = useState(100);
+
+  // TODO: Save these settings in local storage
+  // Use reducer here in case multiple updates happen simultaneously
+  const [config, updateConfig] = useReducer(
+    (current: ViewerConfig, newProperties: Partial<ViewerConfig>) => ({ ...current, ...newProperties }),
+    defaultViewerConfig
+  );
```
frasercl, line 71: `useReducer`! Heck yeah!

src/App.tsx
frasercl: My guess is the only guidance React people would give here boils down to "do your best to encapsulate," i.e. keep state as far up the component tree as it can be and don't give child components access to data they don't need. This state needs to be at the top-level so this component can also pass it to `CanvasWrapper`, and all the settings are relevant to `SettingsTab` because that's where settings get set. So I think we're good!

frasercl (approved):
Looks great, really cleans things up! And sets you up well to easily add settings (to the `ViewerConfig` type) without too much setup (like a `useState`).

PR #175, opened 2024-01-02
feature: GitHub pages build

ShrimpCryptid:
Adds an action that triggers on pushes to main that will deploy a public build to GitHub pages.

New builds should be accessible via this link: 
https://allen-cell-animated.github.io/nucmorph-colorizer/

### DIFF
--- a/.github/workflows/deploy-public-build.yml
+++ b/.github/workflows/deploy-public-build.yml
@@ -19,7 +19,7 @@ jobs:
           node-version: "18"
 
       - name: Install and Build
-        run: npm ci && npx vite build --base=https://allen-cell-animated.github.io/nucmorph-colorizer/
+        run: npm ci && npx vite build
 
       - name: Run Linter
         run: npm run lint
--- a/.github/workflows/deploy-public-build.yml
+++ b/.github/workflows/deploy-public-build.yml
@@ -18,9 +18,8 @@ jobs:
         with:
           node-version: "18"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-public-build.yml
```diff
@@ -0,0 +1,37 @@
+# Adapted from https://github.com/marketplace/actions/deploy-pr-preview.
+name: Deploy Public Build
+
+on:
+  push:
+    branches:
+      - main
```
toloudis, line 7: one thing to consider is doing versioned releases where we only do this build based on whether the push to main has a tag starting with `v`.  Maybe a separate PR since it adds a bit more infrastructure.

.github/workflows/deploy-public-build.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "18"
+
+      - name: Install and Build
+        run: npm ci && npx vite build --base=https://allen-cell-animated.github.io/nucmorph-colorizer/
```
toloudis, line 22: is it necessary to hardcode the url here? oh well

toloudis (approved):
LGTM

.github/workflows/deploy-public-build.yml
ShrimpCryptid: I can try it without, it was only necessary for the preview subfolders.

frasercl (approved):

PR #176, opened 2024-01-04
fix: Allowed time slider interaction during playback

ShrimpCryptid:
Problem
=======
Fixes a bug where the time slider would behave erratically when dragged during playback.

Solution
========
Now, the time playback will pause temporarily when the time slider changes, and resume only once the user releases the mouse button.

## Type of change

Change summary:
---------------
* Added temporary pause when user interacts with time slider during playback. 

Steps to Verify:
----------------

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -116,6 +116,18 @@ function App(): ReactElement {
 
   const [isRecording, setIsRecording] = useState(false);
   const timeControls = useConstructor(() => new TimeControls(canv!, playbackFps));
+  // TODO: Move all logic for the time slider into its own component.
+  // Flag used to indicate that playback was occurring before the slider was moved,
+  // and that it should resume after the slider is released.
+  const [isPlaybackPausedTemporarily, setIsPlaybackPausedTemporarily] = useState(false);
+  // Flag used to indicate that the slider is currently being dragged.
+  const [clickedOnSlider, setClickedOnSlider] = useState(false);
+
+  useEffect(() => {
+    if (timeControls.isPlaying()) {
+      setIsPlaybackPausedTemporarily(false);
+    }
+  }, [timeControls.isPlaying()]);
 
   /** The frame selected by the time UI. Changes to frameInput are reflected in
    * canvas after a short delay.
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -116,6 +116,18 @@ function App(): ReactElement {
 
   const [isRecording, setIsRecording] = useState(false);
   const timeControls = useConstructor(() => new TimeControls(canv!, playbackFps));
+  // TODO: Move all logic for the time slider into its own component.
```
ShrimpCryptid, line 119: I was tempted to do this as part of this PR, but I think the changes are much more readable if I don't touch App just yet.

src/colorizer/TimeControls.ts
```diff
@@ -81,22 +86,34 @@ export default class TimeControls {
     this.timerId = window.setTimeout(() => loadNextFrame(this.canvas.getCurrentFrame()), 0);
   }
 
-  public async handlePlayButtonClick(): Promise<void> {
+  /**
+   * Begins playback of the time series. If the time series is already playing, this function does
+   * nothing.
+   * @param onNewFrameCallback An optional callback that will be called whenever a new frame is loaded.
+   */
+  public async play(onNewFrameCallback: () => void = () => {}): Promise<void> {
```
toloudis, line 94: I don't see when this callback is passed in anywhere

src/colorizer/TimeControls.ts
```diff
@@ -81,22 +86,34 @@ export default class TimeControls {
...
     }
     this.timerId = NO_TIMER_ID;
     this.pauseCallbacks.forEach((callback) => callback());
   }
 
-  public async handleFrameAdvance(delta: number = 1): Promise<void> {
+  /**
+   * Increment or decrement the time series by the given number of frames.
+   */
+  public async offsetFrame(delta: number = 1): Promise<void> {
```
toloudis, line 116: Nit: I think advanceFrame or incrementFrame might be slightly better.  The word offset for me can have other array, indexing, or memory implications.

src/App.tsx
```diff
@@ -120,6 +120,18 @@ function App(): ReactElement {
 
   const [isRecording, setIsRecording] = useState(false);
   const timeControls = useConstructor(() => new TimeControls(canv!, playbackFps));
+  // TODO: Move all logic for the time slider into its own component!
```
toloudis, line 123: could timeControls encapsulate this logic?

toloudis (approved):
This change seems like a very good thing.  
Not sure if you need both of isPlaybackPaused and isDraggingSlider, it seems like you could have only one boolean but I haven't thought that hard about it.

src/App.tsx
ShrimpCryptid: Maybe, but I think the flags I have here are all more interaction-focused, and `TimeControls` is more of a pure utility class. Having a UI component handle these sorts of interaction minutia makes more sense to me.

src/colorizer/TimeControls.ts
ShrimpCryptid: The callback isn't currently being used, yeah. I figured I should do the refactor to make it make it accessible in case that changes, but it's a maybe YAGNI situation :) Remove?

ShrimpCryptid:
> This change seems like a very good thing. Not sure if you need both of isPlaybackPaused and isDraggingSlider, it seems like you could have only one boolean but I haven't thought that hard about it.

Made this change! Though the new flag is called `isTimeSliderDraggedDuringPlayback` which is not the most succinct.

src/App.tsx
toloudis: Agree.  And I like the idea of having "business logic" classes that are independent of ui framework, such that the ui framework makes calls into that api. Feels more maintainable when we try to have that separation.

toloudis (approved):

interim17:
LGTM, nice!

interim17 (approved):

PR #178, opened 2024-01-05
fix: Build and deploy GitHub action no longer deletes existing files

ShrimpCryptid:
Problem
=======
Fixes an issue where the GitHub deployment action would delete the existing PR preview files. Now, the build files are placed in a subdirectory `/main`. Also updates the resource links and adds artifacts so the build files can be inspected.


toloudis (approved):

toloudis (commented):
just thought of something: want to update the README to add the public url and say that it is always the current build of `main` branch

frasercl (approved):

Issue #189, opened 2024-01-22
Use color ramp to colorize scatterplot


ShrimpCryptid:
## Use Case

- Use the current color ramp (and/or color palette) to colorize the scatterplot.

## Acceptance Criteria

- Scatterplot should have an option use the currently selected color ramp and the feature/min max.
- Scatterplot should also copy the out-of-range value options, as set in Settings.

## Details

- While this feature is useful, priority is not to slow down the rest of the UI and to keep the colorizer performant.
  - Having a manual sync button is OK.
  - May need to split coloring into multiple traces of solid colors to reduce the rendering time. See TODOs in ScatterPlotTab.
- See #185.

PR #193, opened 2024-01-26
refactor: Changed Dataset outliers to load as a buffer instead of texture

ShrimpCryptid:
Changes the Dataset so it loads outliers to a buffer/array that can then be read from directly.

This will let the scatterplot colorizer read which values are outliers as part of the changes for #189.

### DIFF
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -329,7 +329,7 @@ export default class ColorizeCanvas {
     }
     this.dataset = dataset;
     if (this.dataset.outliers) {
-      this.setUniform("outlierData", this.dataset.outliers);
+      this.setUniform("outlierData", packDataTexture(Array.from(this.dataset.outliers), FeatureDataType.U8));
     } else {
       this.setUniform("outlierData", packDataTexture([0], FeatureDataType.U8));
     }
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -61,7 +61,7 @@ export default class Dataset {
   private features: Map<string, FeatureData>;
 
   private outlierFile?: string;
-  public outliers?: Texture | null;
+  public outliers?: Uint8Array | null;
 
   private tracksFile?: string;
   private timesFile?: string;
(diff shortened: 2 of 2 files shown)


src/colorizer/Dataset.ts
```diff
@@ -227,18 +227,18 @@ export default class Dataset {
...
+  //     return null;
+  //   }
+  //   try {
+  //     const url = this.resolveUrl(fileUrl);
+  //     const source = await this.arrayLoader.load(url);
+  //     return source.getTexture(dataType);
+  //   } catch (e) {
+  //     return null;
+  //   }
+  // }
```
frasercl, line 241: Probably safe to fully remove this?

src/colorizer/ColorizeCanvas.ts
```diff
@@ -329,7 +329,7 @@ export default class ColorizeCanvas {
     }
     this.dataset = dataset;
     if (this.dataset.outliers) {
-      this.setUniform("outlierData", this.dataset.outliers);
+      this.setUniform("outlierData", packDataTexture(Array.from(this.dataset.outliers), FeatureDataType.U8));
```
frasercl, line 332: Going back over `packDataTexture` and realizing that it probably ought to be accepting `TypedArray`s rather than regular JS arrays... not on you to fix for this PR, just an observation

frasercl (approved):

src/colorizer/Dataset.ts
ShrimpCryptid: Deleted!

src/colorizer/Dataset.ts
```diff
@@ -61,7 +61,7 @@ export default class Dataset {
   private features: Map<string, FeatureData>;
 
   private outlierFile?: string;
-  public outliers?: Texture | null;
+  public outliers?: Uint8Array | null;
```
interim17, line 65: I think I understand this. The threejs `Texture` is opaque to the scatterplot colorizer so that's why you need the `Uint8Array`?

interim17 (approved):

src/colorizer/Dataset.ts
ShrimpCryptid: Yes, exactly! The scatterplot needs access to the outlier data array. Sorry, could have explained this a bit more.

PR #194, opened 2024-01-26
feature: Colorize scatterplot

### LINKED ISSUE #189: Use color ramp to colorize scatterplot

ShrimpCryptid:
Problem
=======
Closes #189, adding the color ramp coloring to the scatter plot view!

Solution
========
Plotly's existing color ramp rendering is very, very slow. To work around it, we split the color ramp gradient into many smaller colors, then take our data and sort them into many different buckets corresponding to each color. Each of these buckets can then be rendered with a single color marker, which Plotly is able to handle much faster.

This PR adds the new method `colorizeScatterPlotPoints` in `ScatterPlotTab`, which does the following steps:
1. It subsamples the current color ramp (or uses the current categorical palette).
1. It uses the feature data to sort each object into a bucket with its matching color.
1. It splits large buckets into smaller buckets (enforcing a max point size of 1024). 
1. Transforms the buckets into individual traces (think groups of data + formatting) that can be rendered by Plotly.

These changes led to a BIG performance increase on top of the existing coloring! I suspect that Plotly dislikes huge data arrays, and splitting up the changes allowed it to run much faster.

New helper methods have been added to `scatter_plot_data_utils.ts`, and I've included unit tests for the new functionality.

## Type of change

Change summary:
---------------
* Adds color ramp + palette coloring to the `ScatterPlotTab`.
* Refactors how traces are handled by `ScatterPlotTab`.
* Adds helper methods for Plotly data and associated unit tests.

Steps to Verify:
----------------
Experiment with these two preview links: 
  - https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-194/
  -  https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-194/?dataset=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2F3500005828_25%2Fmanifest.json

Screenshots (optional):
-----------------------

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/56e5488f-6449-4b5f-ac41-fe99557dddda

image

image

image

Suggested review order:
-----------------------
1. `scatter_plot_data_utils.ts`
3. `ScatterPlotTab`

### DIFF
--- a/src/colorizer/utils/math_utils.ts
+++ b/src/colorizer/utils/math_utils.ts
@@ -51,6 +51,9 @@ export function remap(
   if (clamp) {
     input = Math.min(Math.max(input, inMin), inMax);
   }
+  if (inMin === inMax) {
+    return outMin;
+  }
   return ((input - inMin) / (inMax - inMin)) * (outMax - outMin) + outMin;
 }
 
--- a/src/colorizer/utils/math_utils.ts
+++ b/src/colorizer/utils/math_utils.ts
@@ -49,7 +49,9 @@ export function remap(
   clamp: boolean = true
 ): number {
   if (clamp) {
-    input = Math.min(Math.max(input, inMin), inMax);
+    const min = Math.min(inMin, inMax);
+    const max = Math.max(inMin, inMax);
+    input = Math.min(Math.max(input, min), max);
   }
   if (inMin === inMax) {
     return outMin;
--- a/src/colorizer/utils/math_utils.ts
+++ b/src/colorizer/utils/math_utils.ts
@@ -36,7 +36,9 @@ function numberToUnicodeSuperscript(input: number): string {
 }
 
 /**
- * Remaps a value from one range to another.
+ * Remaps a value from one range to another, optionally clamping the input value to the input range.
+ *
+ * Handles reversed ranges (e.g. `inMin > inMax` or `outMin > outMax`). If `inMin === inMax`, returns `outMin`.
  *
  * @param clamp If true (default), clamps the input to the input range.
  */
--- a/src/colorizer/utils/react_utils.ts
+++ b/src/colorizer/utils/react_utils.ts
@@ -41,34 +41,6 @@ export function useDebounce<T>(value: T, delayMs?: number): T {
 }
 
 /**
- * Combines useTransition and useDebounce. Debounces value and initiates a state transition once the value has come
- * to rest.
- * @param inputValue The value to debounce and transition.
- * @param startTransition The startTransition function returned by useTransition.
- * @param onChange An optional callback that triggers when the debounced value changes, but before the transition.
- * @param delayMs The delay, in milliseconds. 500 ms by default.
- */
-export function useTransitionedDebounce<T>(
-  inputValue: T,
-  startTransition: React.TransitionStartFunction,
-  onChange: () => void = () => {},
-  delayMs: number = 500
-): T {
-  const [value, setValue] = useState<T>(inputValue);
-  const debouncedValue = useDebounce(inputValue, delayMs);
-
-  // Only update value when the debounced value has changed.
-  useEffect(() => {
-    onChange();
-    startTransition(() => {
-      setValue(debouncedValue);
-    });
-  }, [debouncedValue]);
-
-  return value;
-}
-
-/**
  * Returns a reference to a constructed value that will not be re-computed between renders.
  *
  * Functionally, this is a wrapper around useRef and allows it to be used in a type-safe way.
--- a/src/components/tabs/ScatterPlotTab.tsx
+++ b/src/components/tabs/ScatterPlotTab.tsx
@@ -25,7 +25,6 @@ import {
   splitTraceData,
   subsampleColorRamp,
 } from "./scatter_plot_data_utils";
-import { clamp } from "three/src/math/MathUtils";
 
 /** Extra feature that's added to the dropdowns representing the frame number. */
 const TIME_FEATURE = { key: "scatterplot_time", name: "Time" };
@@ -597,7 +596,7 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
       if (isUsingTime) {
         traces.push(makeLineTrace(trackData.xData, trackData.yData));
       }
-      // Render track only
+      // Render track points
       const trackTraces = colorizeScatterplotPoints(
         trackData.xData,
         trackData.yData,
--- a/src/components/tabs/ScatterPlotTab.tsx
+++ b/src/components/tabs/ScatterPlotTab.tsx
@@ -35,6 +35,7 @@ const PLOTLY_CONFIG: Partial<Plotly.Config> = {
 };
 
 const MAX_POINTS_PER_TRACE = 1024;
+const COLOR_RAMP_SUBSAMPLES = 100;
 
 enum RangeType {
   ALL_TIME = "All time",
@@ -452,7 +453,7 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
   /**
    * Applies coloring to point traces in a scatterplot. Does this by splitting the data into multiple traces each with a solid
    * color, which is much faster than using Plotly's native color ramping. Also enforces a maximum number of points
-   * per trace, which also seems to speed up Plotly renders.
+   * per trace, which significantly speeds up Plotly renders.
    *
    * @param xData
    * @param yData
@@ -489,12 +490,11 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
       });
     } else {
       // Generate colors
-      const NUM_RANGE_COLORS = 100; // TBD if this is a good number?
       const categories = dataset.getFeatureCategories(selectedFeatureName);
       const colors =
         categories !== null
           ? categoricalPalette.slice(0, categories.length)
-          : subsampleColorRamp(colorRamp, NUM_RANGE_COLORS);
+          : subsampleColorRamp(colorRamp, COLOR_RAMP_SUBSAMPLES);
 
       // Make a bucket group for each ramp/palette color and for the out-of-range and outliers.
       // index 0 = out of filter range, index 1 = outliers.
--- a/src/components/tabs/ScatterPlotTab.tsx
+++ b/src/components/tabs/ScatterPlotTab.tsx
@@ -15,7 +15,15 @@ import { FlexRow, FlexRowAlignCenter } from "../../styles/utils";
 import { SwitchIconSVG } from "../../assets";
 import { Color } from "three";
 import { DrawMode, ViewerConfig } from "../../colorizer/types";
-import { TraceData, getBucketIndex, splitTraceData, subsampleColorRamp } from "./scatter_plot_data_utils";
+import {
+  DataArray,
+  TraceData,
+  drawCrosshair,
+  getBucketIndex,
+  makeLineTrace,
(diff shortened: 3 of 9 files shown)


src/colorizer/utils/math_utils.ts
ShrimpCryptid, line 1: Allows flipping the min/max bounds and fixes a divide-by-zero error when the input min/max were the same.

src/colorizer/utils/math_utils.ts
```diff
@@ -49,7 +51,12 @@ export function remap(
   clamp: boolean = true
 ): number {
   if (clamp) {
-    input = Math.min(Math.max(input, inMin), inMax);
+    const min = Math.min(inMin, inMax);
+    const max = Math.max(inMin, inMax);
+    input = Math.min(Math.max(input, min), max);
+  }
+  if (inMin === inMax) {
+    return outMin;
```
toloudis, line 59: Nit: a couple of tiny simple unit tests might be nice here, not sure if they already exist. (regular in-bounds case, out of bounds case, reversed range in and out of bounds cases, and input===inMin or inMax... clamp/no clamp for all).

src/colorizer/utils/react_utils.ts
```diff
@@ -40,34 +40,6 @@ export function useDebounce<T>(value: T, delayMs?: number): T {
...
-  useEffect(() => {
-    onChange();
-    startTransition(() => {
-      setValue(debouncedValue);
-    });
-  }, [debouncedValue]);
-
-  return value;
-}
-
```
toloudis, line 70: 🚀

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -386,6 +392,131 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
     );
   };
 
+  /**
+   * Applies coloring to point traces in a scatterplot. Does this by splitting the data into multiple traces each with a solid
+   * color, which is much faster than using Plotly's native color ramping. Also enforces a maximum number of points
+   * per trace, which significantly speeds up Plotly renders.
```
toloudis, line 398: Nit: minor concern that a scientist looking very closely at the points will notice the binning and therefore inaccurate visualization. Just something to anticipate, maybe unlikely that it comes up.

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -386,6 +392,131 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+   * @param objectIds
+   * @param trackIds
+   * @param markerConfig Additional marker configuration to apply to all points. By default,
+   * markers are size 4.
+   * @param {Color | undefined} overrideColor When defined, uses a base color for all points, instead of
+   * calculating based on the color ramp or palette.
+   */
+  const colorizeScatterplotPoints = (
+    xData: DataArray,
+    yData: DataArray,
```
toloudis, line 411: are these the complete pre-binned data points?

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -386,6 +392,131 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+  ): Partial<PlotData>[] => {
+    if (selectedFeatureName === null || dataset === null || !xAxisFeatureName || !yAxisFeatureName) {
+      return [];
+    }
+
+    const featureData = dataset.getFeatureData(selectedFeatureName);
+
+    const traceDataBuckets: TraceData[] = [];
+    if (!featureData || markerConfig.color || overrideColor) {
+      // Do no coloring! Keep all points in the same bucket.
```
toloudis, line 425: Would we get a speedup if we binned anyway?? just wondering.

ShrimpCryptid: These points will get binned below in an extra step! I'll update this comment.

src/components/tabs/ScatterPlotTab.tsx
ShrimpCryptid: That's fair. I imagine if you spread the color ramp ends VERY far apart (like, 100X past the ends of the color ramp) it might be visible? At some point we'd be running into the limitations of RGBA itself anyways, since there's a max of 256 values (and therefore a max of 256 relevant subsampled colors) per channel.

src/colorizer/utils/math_utils.ts
ShrimpCryptid: They do! ✨✨

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -386,6 +392,131 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+   * @param xData
+   * @param yData
+   * @param objectIds
+   * @param trackIds
+   * @param markerConfig Additional marker configuration to apply to all points. By default,
+   * markers are size 4.
+   * @param {Color | undefined} overrideColor When defined, uses a base color for all points, instead of
+   * calculating based on the color ramp or palette.
+   */
+  const colorizeScatterplotPoints = (
```
toloudis, line 409: I wonder if most or all of this function could move into scatter_plot_data_utils.

src/components/tabs/scatter_plot_data_utils.ts
toloudis, line 1: not sure why we have snake case file naming here. is that to distinguish it from react components?

ShrimpCryptid: It was a project convention I'd been following that I hadn't thought too hard about 😅
Yes, currently any utility files that aren't classes or components use snake case.

src/components/tabs/ScatterPlotTab.tsx
ShrimpCryptid: It could, it currently accesses a lot of state variables that I'd need to pass into it though. Let me think on this one.

toloudis (approved):

src/components/tabs/scatter_plot_data_utils.ts
frasercl: I believe I started this with texture_utils.ts, and yeah the idea was for files that didn't contain a single class or component

tests/scatter_plot_data_utils.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    expect(sorter(10)).to.equal(9);
+    expect(sorter(100)).to.equal(9);
+    expect(sorter(Number.POSITIVE_INFINITY)).to.equal(9);
+  });
+
+  it("Uses rounding to determine bucket index", () => {
+    const sorter = bucketSorterFactory(0, 1, 2);
+    expect(sorter(0.499999)).to.equal(0);
+    expect(sorter(0.5)).to.equal(1);
+  });
```
frasercl, line 63: This test doesn't capture the above comment since it's sampling right in the middle of the range, but e.g.
```typescript
const sorter = bucketSorterFactory(0, 1, 4);
expect(sorter(0.2499999)).to.equal(0);
expect(sorter(0.25)).to.equal(1);
```
will fail.

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -111,11 +107,21 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+
+  const isDebouncePending =
+    dataset !== props.dataset || colorRampMin !== props.colorRampMin || colorRampMax !== props.colorRampMax;
+
+  // Signal that the plot is stalled when playback starts.
+  useEffect(() => {
+    if (isPlaying) {
+      setIsRendering(true);
+    }
+  }, [isPlaying]);
```
frasercl, line 124: Would it work to add `isPlaying` to the expression below that determines whether the loading spinner is shown, rather than updating state here?

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -386,6 +392,131 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+
+      // Make a bucket group for each ramp/palette color and for the out-of-range and outliers.
+      // index 0 = out of filter range, index 1 = outliers.
+      for (let i = 0; i < colors.length + 2; i++) {
+        let color;
+        if (i === 0) {
+          color = viewerConfig.outOfRangeDrawSettings.color;
+        } else if (i === 1) {
+          color = viewerConfig.outlierDrawSettings.color;
+        } else {
```
frasercl, line 451: Believe I commented this in a child PR (whoops), but could you move these into two initial `push` calls and get rid of the `if` inside this loop?

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -386,6 +392,131 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+    }
+
+    // Transform buckets into traces
+    const traces: Partial<PlotData>[] = traceDataBuckets
+      .filter((bucket) => bucket.x.length > 0) // Remove empty buckets
+      .reduce((acc: TraceData[], bucket: TraceData) => {
+        // Split the traces into smaller chunks to prevent plotly from freezing.
+        acc.push(...splitTraceData(bucket, MAX_POINTS_PER_TRACE));
+        return acc;
+      }, [])
```
frasercl, line 495: would love it if I could recommend `flatMap` here... would look so much nicer... would (I think) be less efficient...

frasercl (commented):
This looks quite nice overall! As performance hacks go, binning colors is a clever and satisfying one. And just the right amount of complexity is offloaded into  utility methods.

Waiting to approve since one of my comments describes a possible behavior change, but didn't want to lose that this is real clean.

frasercl (commented):
> Maybe a tiny thing given the number of buckets, but I think this mapping will cause the buckets at the very edges to be half-sized? Might recommend `Math.floor` with `remap`'s output range from `0` to `numBuckets` (rather than `numBuckets - 1`), though you'd have to special-case `value >= maxValue`, which would get mapped to `numBuckets` rather than `numBuckets - 1`. 

You know what - this matches with how the colors are sampled out of the ramp. So we're good after all

frasercl (approved):

src/components/tabs/ScatterPlotTab.tsx
ShrimpCryptid: This made me cackle out loud, thank you

src/components/tabs/ScatterPlotTab.tsx
ShrimpCryptid: Yup, should be fixed in the child PR!

tests/scatter_plot_data_utils.test.ts
ShrimpCryptid: Ended up making the behavior explicit through a method comment 👍

PR #195, opened 2024-01-29
feature: Thresholding in Scatterplot

### LINKED ISSUE #185: Apply current threshold in scatterplot
## Use Case

Extension of #94. 

## Acceptance Criteria
- Move threshold calculation code out of `ColorizeCanvas` so that it only needs to be calculated in one location on changes to threshold.
  - May also be a good opportunity to evaluate whether it's slow enough to move into a worker thread.
- Apply out-of-range coloration to the scatterplot.
  - Filter scatterplot data values by the current threshold, and move filtered points into their own trace.
  - Update whenever the threshold/thresholded values change.

## Details

- Threshold and scatterplot should stay in sync automatically. (Debouncing thresholds is OK)

ShrimpCryptid:
Problem
=======
Closes #185!
This took WAY less time than I thought it would!

Solution
========
- Moved the threshold calculation code into `data_utils.ts`, and performed the calculation step in `App.tsx`.
- Adjusted `ColorizeCanvas` to take the threshold LUT in as a prop.
- Updated `ScatterPlotTab` to colorize using the threshold LUT.
- Fixes to scatter plot for interactions between color ramp and thresholded values.

## Type of change

## Suggested review order:
(Please turn off "Show Whitespace"-- some of the changes I made were an indentation change from a removed `if` check, which makes it harder to review.)
1. `data_utils.ts` and `ColorizeCanvas.ts`: Moved logic for determining what values are in range out of `ColorizeCanvas`.
1. `App.tsx` and `CanvasWrapper.tsx`: Now calculates the feature threshold LUT and passes it down to `ColorizeCanvas`.
2. `ScatterPlotTab.tsx`: Add support for thresholding! And additional edits to coloring and interaction with tracks.

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-195/
2. Set a filter on the dataset from the Filters tab.
3. Open the Scatterplot tab and try selecting different tracks and switching through different time steps.
4. Try hiding or changing the custom color for filtered values from the Settings tab.

Screenshots (optional):
-----------------------

Video Talkthrough (🔊):
https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/8ec98986-e5f9-4477-998f-efeb7aa091d4

Threshold on two features:
image

Custom coloring:
image

Hide filtered values:
image

### DIFF
--- a/src/colorizer/utils/data_utils.ts
+++ b/src/colorizer/utils/data_utils.ts
@@ -85,3 +85,37 @@ export function isValueWithinThreshold(value: number, threshold: FeatureThreshol
     return threshold.enabledCategories[value];
   }
 }
+
+/**
+ * Returns an array of Uint8 values representing whether the object ID at that index is in range of
+ * the thresholds.
+ * @param {Dataset} dataset A valid Dataset object.
+ * @param {FeatureThreshold[]} thresholds Array of feature thresholds, which must define the feature name, min, and max.
+ * If a feature name cannot be found in the dataset, it will be ignored.
+ * @returns A Uint8Array, with a length equal to the number of objects in the dataset.
+ * For each object ID `i`, `inRangeIds[i]` will be 1 if the object is in range and 0 if it is not.
+ */
+export function getInRangeIds(dataset: Dataset, thresholds: FeatureThreshold[]): Uint8Array {
+  // TODO: Optimize memory by using a true boolean array?
+  // TODO: If optimizing, use fuse operation via shader.
+  const inRangeIds = new Uint8Array(dataset.numObjects);
+
+  // Ignore thresholds with features that don't exist in this dataset or whose units don't match
+  const validThresholds = thresholds.filter((threshold) => {
+    const featureData = dataset.getFeatureData(threshold.featureName);
+    return featureData && featureData.units === threshold.units;
+  });
+
+  for (let id = 0; id < dataset.numObjects; id++) {
+    inRangeIds[id] = 1;
+    for (let thresholdIdx = 0; thresholdIdx < validThresholds.length; thresholdIdx++) {
+      const threshold = validThresholds[thresholdIdx];
+      const featureData = dataset.getFeatureData(threshold.featureName);
+      if (featureData && !isValueWithinThreshold(featureData.data[id], threshold)) {
+        inRangeIds[id] = 0;
+        break;
+      }
+    }
+  }
+  return inRangeIds;
+}
--- a/src/colorizer/utils/data_utils.ts
+++ b/src/colorizer/utils/data_utils.ts
@@ -87,13 +87,14 @@ export function isValueWithinThreshold(value: number, threshold: FeatureThreshol
 }
 
 /**
- * Returns a Uint8 look-up table of whether the object ID at that index is in range of
- * the thresholds (=1) or not (=0).
+ * Returns a Uint8 array look-up table indexed by object ID, storing whether that object ID is in range of
+ * the given thresholds (=1) or not (=0).
  * @param {Dataset} dataset A valid Dataset object.
- * @param {FeatureThreshold[]} thresholds Array of feature thresholds, which must define the feature name, min, and max.
+ * @param {FeatureThreshold[]} thresholds Array of feature thresholds, which match agaisnt the feature name and unit.
  * If a feature name cannot be found in the dataset, it will be ignored.
  * @returns A Uint8Array, with a length equal to the number of objects in the dataset.
- * For each object ID `i`, `inRangeIds[i]` will be 1 if the object is in range and 0 if it is not.
+ * For each object ID `i`, `inRangeIds[i]` will be 1 if the object is in range of the thresholds
+ * and 0 if it is not.
  */
 export function getInRangeLUT(dataset: Dataset, thresholds: FeatureThreshold[]): Uint8Array {
   // TODO: Optimize memory by using a true boolean array?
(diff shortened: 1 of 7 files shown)


src/colorizer/utils/data_utils.ts
ShrimpCryptid, line 1: Copied from ColorizeCanvas

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -420,69 +426,94 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+    const colorMaxValue = isCategorical ? categories.length - 1 : colorRampMax;
 
-        const bucket = traceDataBuckets[bucketIndex];
-        bucket.x.push(xData[i]);
-        bucket.y.push(yData[i]);
-        bucket.objectIds.push(objectIds[i]);
-        bucket.trackIds.push(trackIds[i]);
+    // Make a bucket group for each ramp/palette color and for the out-of-range and outliers.
+    // index 0 = out of filter range, index 1 = outliers.
+    const traceDataBuckets: TraceData[] = [];
```
ShrimpCryptid, line 450: `TraceData.color` is now a hex string instead of a Three.js `Color`, since `Color` can't include transparency information.

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -411,78 +417,105 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+    const usingOverrideColor = markerConfig.color || overrideColor;
+    overrideColor = overrideColor || new Color(markerConfig.color as ColorRepresentation);
+
+    let colors: Color[];
+    if (usingOverrideColor) {
+      // Use a single color bucket
+      colors = [overrideColor];
+    } else if (isCategorical) {
+      colors = categoricalPalette.slice(0, categories.length);
     } else {
```
ShrimpCryptid, line 433: This caused a bug where hidden outliers/out-of-range values would be mistakenly revealed on the plot, since the hiding/revealing of values is handled by what used to be the `else` branch. I merged everything into one code pathway here.

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -411,78 +417,105 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+        marker = { ...markerConfig, ...markerConfig.outliers };
+      } else {
+        color = `#${colors[i - 2].getHexString()}`;
+        marker = markerConfig;
       }
+      if (usingOverrideColor) {
+        color = `#${overrideColor.getHexString()}`;
+      }
+      traceDataBuckets.push({ x: [], y: [], objectIds: [], trackIds: [], color, marker });
+    }
```
frasercl, line 470: Might suggest unrolling "out of range" and "outlier" categories from this loop, rather than hitting an `if` for every iteration?

frasercl (approved):

src/colorizer/utils/data_utils.ts
```diff
@@ -85,3 +85,38 @@ export function isValueWithinThreshold(value: number, threshold: FeatureThreshol
...
+ * the given thresholds (=1) or not (=0).
+ * @param {Dataset} dataset A valid Dataset object.
+ * @param {FeatureThreshold[]} thresholds Array of feature thresholds, which match agaisnt the feature name and unit.
+ * If a feature name cannot be found in the dataset, it will be ignored.
+ * @returns A Uint8Array, with a length equal to the number of objects in the dataset.
+ * For each object ID `i`, `inRangeIds[i]` will be 1 if the object is in range of the thresholds
+ * and 0 if it is not.
+ */
+export function getInRangeLUT(dataset: Dataset, thresholds: FeatureThreshold[]): Uint8Array {
+  // TODO: Optimize memory by using a true boolean array?
```
toloudis, line 100: Probably not needed but if this was ever a concern, could also pack 8 boolean bits into each uint8!

src/components/tabs/ScatterPlotTab.tsx
toloudis: Agree with the above.. also could preallocate the size of the traceDataBuckets array but I suppose this is not a performance bottleneck yet.  It's sufficiently readable for me the way it is too.

src/components/tabs/ScatterPlotTab.tsx
```diff
@@ -411,78 +417,105 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+
+      let bucketIndex;
+      if (isOutOfRange) {
+        bucketIndex = 0;
+      } else if (isOutlier) {
+        bucketIndex = 1;
+      } else if (usingOverrideColor) {
+        bucketIndex = 2;
+      } else {
+        bucketIndex = getBucketIndex(featureData.data[objectId], colorMinValue, colorMaxValue, colors.length) + 2;
```
toloudis, line 490: this +2 offset starts to look like a bookkeeping problem.  Maybe at least call it NUM_PRESET_COLOR_BUCKETS or something?

toloudis: Also could name 0=COLOR_INDEX_OUTOFRANGE and 1=COLOR_INDEX_OUTLIERS etc. to keep it consistent across a couple of places

toloudis (approved):
just one minor comment about magic numbers that show up in more than one place, but not a showstopper - overall LGTM!

PR #196, opened 2024-01-29
fix: Fixed bad aspect ratio when loading datasets

### LINKED ISSUE #186: Occasional incorrect aspect ratio when loading new datasets
## Description

Sometimes, when switching between two datasets, the loaded image in the main canvas view will suddenly have an incorrect aspect ratio. It only goes away when refreshed, or if switching to another dataset.

ShrimpCryptid:
Problem
=======
Closes #186, fixing a bug where aspect ratios would occasionally appear squashed when loading a dataset.

Solution
========
It turns out the bug was caused when the viewer was at a timestep that didn't exist in the loaded dataset. This caused the viewer to skip loading that frame. There's a frame dimension variable saved inside of the Dataset that is only updated when frames are loaded, so it would be `null` and would use an (incorrect) fallback instead.

To fix it, I forced the first frame in the dataset to load when the Dataset is open, guaranteeing that the frame dimensions will be set correctly before its data can be accessed.

(I also added unit tests to validate the new fix and test it with a few different resolutions.)

## Type of change

Steps to Verify:
----------------
Repro steps:
1. Open the main viewer build here: https://allen-cell-animated.github.io/nucmorph-colorizer/main/
2. Slide the time slider all the way to the right.
3. Open the load dialog and paste in this dataset: https://raw.githubusercontent.com/allen-cell-animated/nucmorph-colorizer/main/documentation/example_dataset/manifest.json
4. The boxes should appear squished instead of square.

To test the fix:
1. Repeat all the above steps on the preview build: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-196/
1. The boxes should appear square.

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -373,7 +373,7 @@ export default class Dataset {
       this.loadFrame(0), // load first frame to set frame dimensions
       ...featuresPromises,
     ]);
-    const [outliers, tracks, times, centroids, bounds, _loadFrame, ...featureResults] = result;
+    const [outliers, tracks, times, centroids, bounds, , ...featureResults] = result;
 
     this.outliers = outliers;
     this.trackIds = tracks;
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -370,10 +370,11 @@ export default class Dataset {
       this.loadToBuffer(FeatureDataType.U32, this.timesFile),
       this.loadToBuffer(FeatureDataType.U16, this.centroidsFile),
       this.loadToBuffer(FeatureDataType.U16, this.boundsFile),
-      this.loadFrame(0), // load first frame to set frame dimensions
+      this.loadFrame(0), // load first frame to set frame dimensions. Load result will be discarded
       ...featuresPromises,
     ]);
-    const [outliers, tracks, times, centroids, bounds, , ...featureResults] = result;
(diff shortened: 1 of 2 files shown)


interim17 (approved):

frasercl (approved):

frasercl (approved):

src/colorizer/Dataset.ts
```diff
@@ -370,9 +370,10 @@ export default class Dataset {
       this.loadToBuffer(FeatureDataType.U32, this.timesFile),
       this.loadToBuffer(FeatureDataType.U16, this.centroidsFile),
       this.loadToBuffer(FeatureDataType.U16, this.boundsFile),
+      this.loadFrame(0), // load first frame to set frame dimensions
       ...featuresPromises,
     ]);
-    const [outliers, tracks, times, centroids, bounds, ...featureResults] = result;
+    const [outliers, tracks, times, centroids, bounds, , ...featureResults] = result;
```
toloudis, line 376: did this extra comma sneak in?

toloudis (approved):

src/colorizer/Dataset.ts
ShrimpCryptid: The extra comma is me discarding the results of the frame load. I've gotten a few comments about it so I'll make it explicit that I'm throwing away the results...

PR #202, opened 2024-02-02
fix: Incorrect range when loading non-default feature from URL without range parameter

### LINKED ISSUE #197: Bounds don't reset when loading new dataset
## Description

The color ramp slider's min and max values do not reset when swapping between datasets, even if the checkbox is disabled.

## Expected Behavior

- If "Keep range between datasets" is disabled, the ramp's min and max should reset to the min and max value of the newly-selected feature.

ShrimpCryptid:
Fixed a bug where loading any feature besides the first (default) feature in a dataset via the URL would cause it to use the default feature's color map range instead. This caused odd, unexpected coloring on initial load.

Also fixes #197!

This fixes the bug by using a local variable (`newFeatureName`) to get the feature's color ramp range, rather than the `featureName` state (which may not be updated).

To test:
- Broken build (should appear in bright cyan): https://dev-aics-dtp-001.int.allencell.org/nucmorph-colorizer/dist/index.html?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama%20Bear&feature=Height&color=matplotlib-cool
- Fixed build: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-202/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama%20Bear&feature=Height&color=matplotlib-cool

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -243,6 +243,52 @@ function App(): ReactElement {
     [canv, dataset, featureName, currentFrame]
   );
 
+  /**
+   * Attempts to replace the current feature with a new feature from a dataset.
+   * If the feature cannot be loaded, returns the old feature name and does nothing.
+   * @param newDataset the dataset to pull feature data from.
+   * @param newFeatureName the name of the new feature to select.
+   * @returns the new feature name if it was successfully found and loaded. Otherwise, returns the old feature name.
+   */
+  const replaceFeature = useCallback(
+    (featureDataset: Dataset, newFeatureName: string): string => {
+      if (!featureDataset?.hasFeature(newFeatureName)) {
+        console.warn("Dataset does not have feature '" + newFeatureName + "'.");
+        return featureName;
+      }
+      setFeatureName(newFeatureName);
+      canv.setFeature(newFeatureName);
+      return newFeatureName;
+    },
+    [canv]
+  );
+
+  /**
+   * Updates the color ramp to default min and max values based on a feature and dataset.
+   *
+   * Does nothing if the viewer is configured to keep the range between datasets.
+   *
+   * If the feature is thresholded, the color ramp will be set to the threshold's min and max.
+   * Otherwise, the color ramp will be set to the feature's min and max.
+   */
+  const resetColorRampToDefaults = useCallback(
+    (featureDataset: Dataset, featureName: string): void => {
+      const featureData = featureDataset.getFeatureData(featureName);
+      if (!config.keepRangeBetweenDatasets && featureData) {
+        // Use min/max from threshold if there is a matching one, otherwise use feature min/max
+        const threshold = featureThresholds.find(thresholdMatchFinder(featureName, featureData.units));
+        if (threshold && isThresholdNumeric(threshold)) {
+          setColorRampMin(threshold.min);
+          setColorRampMax(threshold.max);
+        } else {
+          setColorRampMin(featureData.min);
+          setColorRampMax(featureData.max);
+        }
+      }
+    },
+    [featureThresholds, config.keepRangeBetweenDatasets]
+  );
+
   // INITIAL SETUP  ////////////////////////////////////////////////////////////////
 
   // Only retrieve parameters once, because the URL can be updated by state updates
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -459,10 +460,10 @@ function App(): ReactElement {
   );
 
   const updateFeature = useCallback(
-    async (newDataset: Dataset, newFeatureName: string): Promise<void> => {
+    async (newDataset: Dataset, newFeatureName: string): Promise<string> => {
```
frasercl, line 463: Does this need to be `async` at all? I don't see any `await` points

src/App.tsx
```diff
@@ -480,6 +481,7 @@ function App(): ReactElement {
       }
 
       canv.setFeature(newFeatureName);
+      return newFeatureName;
```
frasercl, line 484: `newFeatureName` is an argument to this function and is never modified within its body. Could the caller just use the value they passed in rather than getting it back out of this function?

src/App.tsx
ShrimpCryptid: Good catch, it doesn't.

src/App.tsx
ShrimpCryptid: There's an early return statement that returns the existing `featureName` state if the new feature could not be found.

frasercl: Makes sense!

frasercl (approved):

src/App.tsx
```diff
@@ -458,11 +459,16 @@ function App(): ReactElement {
     [replaceDataset]
   );
 
+  /** Attempts to update the current feature. Also updates the color ramp min/max if the feature has changed.
```
toloudis, line 462: Readability:  this comment.  What does "update" mean.  The function is already called updateFeature. Maybe it should be called attemptUpdateFeatureAndColorRamp ?  Info I would be looking for here:  why does it need an update? Update from what to what?  Like for example, it could say "this is called when we need to update an already loaded feature when a new dataset is loaded that contains the same feature."

toloudis (approved):
LGTM no showstoppers but I would consider clarifying one of the comments

src/App.tsx
ShrimpCryptid: Ended up moving this into two separate methods, `replaceFeature` and `resetColorRampToDefaults`. I also had to move them higher up in `App.tsx`, because they were used before their declaration and it was causing another bug (!) in `replaceDataset`.

ShrimpCryptid:
Rerequesting @toloudis and @frasercl because I renamed and moved a few methods.

toloudis (approved):
awesome!

src/App.tsx
```diff
@@ -399,7 +446,16 @@ function App(): ReactElement {
...
+    [
+      dataset,
+      featureName,
+      canv,
+      currentFrame,
+      getUrlParams,
+      replaceFeature,
+      resetColorRampRangeToDefaults,
+      featureThresholds,
+    ]
```
frasercl, line 458: This makes me wonder how common a pattern it is to have hooks depend on upstream `useCallbacks` within the same component, especially since this dependency array has all of `replaceFeature`'s `useCallback` dependencies alongside `replaceFeature` itself. It's not something I've encountered (or perhaps it's an error I've been leaving around in my own React code?), though it makes perfect sense for it to be there rather than risk unexpected behavior in the future.

frasercl (approved):

src/App.tsx
ShrimpCryptid: Thanks for the note! I've started putting `useCallbacks` into dependency arrays because, based on prior experience, not doing so can lead to unexpected consequences and bugs. As you said here, `replaceFeature` has shared dependencies with `replacedDataset` so it's strictly necessary to include it as a dependency, but this protects the behavior if `replaceFeature` ever adds additional dependencies.

PR #205, opened 2024-02-05
refactor: Enforce naming

ShrimpCryptid:
Enforces common naming conventions during linting.


src/colorizer/Dataset.ts
```diff
@@ -350,11 +350,10 @@ export default class Dataset {
       this.loadToBuffer(FeatureDataType.U32, this.timesFile),
       this.loadToBuffer(FeatureDataType.U16, this.centroidsFile),
       this.loadToBuffer(FeatureDataType.U16, this.boundsFile),
-      this.loadFrame(0), // load first frame to set frame dimensions. Load result will be discarded
+      this.loadFrame(0), // load first frame to set frame dimensions. Load result will be discarded (see `, ,` below)
       ...featuresPromises,
     ]);
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const [outliers, tracks, times, centroids, bounds, _loaded_frame, ...featureResults] = result;
+    const [outliers, tracks, times, centroids, bounds, , ...featureResults] = result;
```
toloudis, line 356: single underscore doesn't work?

src/colorizer/utils/url_utils.ts
```diff
@@ -24,10 +24,10 @@ const URL_PARAM_PALETTE = "palette";
 const URL_PARAM_PALETTE_KEY = "palette-key";
 
 const ALLEN_FILE_PREFIX = "/allen/";
-const ALLEN_PREFIX_TO_HTTPS: Record<string, string> = {
-  "/allen/aics/assay-dev": "https://dev-aics-dtp-001.int.allencell.org/assay-dev",
-  "/allen/aics/microscopy": "https://dev-aics-dtp-001.int.allencell.org/microscopy",
-};
+const ALLEN_PREFIX_TO_HTTPS: Map<string, string> = new Map([
```
toloudis, line 27: was this change required by the new lint rules?

ShrimpCryptid: Yeah, it unfortunately disallows object properties that start in non-alphabetic or "_" characters. Record is really just a type constraint around a Javascript object...

src/colorizer/Dataset.ts
ShrimpCryptid: Fixed this! I added a special exception for underscores.

src/colorizer/utils/url_utils.ts
toloudis: I see. Ok that makes sense!

toloudis (approved):

src/colorizer/utils/url_utils.ts
frasercl: Not at all a blocking concern, but it does bug me a bit that this is compelling a (subtle, unimportant) behavior change in the code, since I believe `Map`s are represented differently and afford the interpreter fewer opportunities for optimization of constant properties. (That "I believe" makes this an extra unimportant criticism, since, like any of us, I have no way of knowing what V8 actually does with any of our code!)

frasercl (approved):

src/colorizer/utils/url_utils.ts
toloudis: I agree - my sense is that it would be better to have a linter ignore comment than to be forced to use a different type that has potentially different behavior.  In this particular case I don't think it really matters, since this is such a simple scenario 😀

PR #209, opened 2024-02-07
fix: Fixed bug where tracks with one time could not be selected

### LINKED ISSUE #188: Cannot select tracks with only one object/time
## Description

- Otherwise valid tracks with only one object in them cannot be selected, causing unexpected behavior with the search bar and the scatter plot.

## Expected Behavior

- Either the track should be selected, or a warning message should pop up.

ShrimpCryptid:
Fixes #188.

`Track.length()` returns 0 for tracks with only a single timepoint, so I substituted a check for the number of object IDs instead.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -239,7 +239,7 @@ function App(): ReactElement {
 
       const newTrack = dataset!.buildTrack(trackId);
 
-      if (newTrack.length() < 1) {
+      if (newTrack.ids.length < 1) {
         // Check track validity
         return;
       }


frasercl (approved):

toloudis (approved):
perfect!  can we have track.length() return the right thing too?  is it being used elsewhere?

ShrimpCryptid:
> perfect! can we have track.length() return the right thing too? is it being used elsewhere?

Yup! The next PR changes it to a `duration` and edits all the places where it's used.

PR #210, opened 2024-02-07
fix: Fixes track path desync for tracks that skipped times

ShrimpCryptid:
Fixes a bug where the calculation of points for the track paths didn't take into account that some tracks could be missing segmentations for certain frames. This would cause paths to advance one or more frames ahead of where they should be.

I fixed this by using fallbacks on frames where a track did not exist, during the calculation of points in the path.
This update also adds better documentation and adjusts a few property names (`length` vs. `duration`).

### Testing
Tracks with missing frames would not update their paths at the very end of the tracks due to the desync. Open the following links and advance a few frames to the very end of the selected track. In the fixed build, you should be able to see the track path update until the very last frame. (it also won't update the track path on the missing frame like in the broken build)

Main (broken) build: https://dev-aics-dtp-001.int.allencell.org/nucmorph-colorizer/dist/index.html?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama%20Bear&feature=Volume&track=279&t=97&color=esri-green_4&palette-key=adobe
Fixed build: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-210/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama%20Bear&feature=Volume&track=279&t=97&color=esri-green_4&palette-key=adobe

### DIFF
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -387,12 +387,24 @@ export default class ColorizeCanvas {
     }
     // Make a new array of the centroid positions in pixel coordinates.
     // Points are in 3D while centroids are pairs of 2D coordinates in a 1D array
-    this.points = new Float32Array(track.length() * 3);
+    this.points = new Float32Array(track.duration() * 3);
 
-    for (let i = 0; i < track.length(); i++) {
+    // Tracks may be missing objects for some timepoints, so use the last known good value as a fallback
+    let lastTrackIndex = 0;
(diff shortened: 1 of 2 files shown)


frasercl (approved):
Looks good!

toloudis (approved):

PR #218, opened 2024-02-14
refactor: Changed url parameters to an enum

ShrimpCryptid:
Grouped all of the URL parameters together instead of having many loose constants.


ascibisz (approved):
Looks good to me! I like grouping constants like this to make things more intuitive

interim17 (approved):

Issue #182, opened 2024-01-10
Save Scatter Plot settings to URL (+current open tab)


ShrimpCryptid:
## Use Case
- Save the x and y features and the current range type of the scatter plot to the URL so that it can be shared.
- Also save the currently open tab, for scientists sharing scatter plot views.

ShrimpCryptid:
`scatter-x`: 
`scatter-y`:
`tab`: 
^ add enum for tabs

PR #232, opened 2024-02-23
feature: ScatterPlot config

ShrimpCryptid:
Problem
=======
First step in #182, "Save Scatter Plot settings to URL (+current open tab)"

*Expected review time: small, 15-20 min*

Creates a new config object for the Scatter Plot tab, containing its current selected feature axes and the range type. This is passed into the `ScatterPlotTab`, and stored as state in the `App`.

The config object will eventually be serialized to the URL for sharing!

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-232/
2. Open the scatter plot tab and click around, changing various settings. Everything should still load in and work as expected.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,7 +13,14 @@ import React, { ReactElement, useCallback, useContext, useEffect, useMemo, useRe
 import { ColorizeCanvas, Dataset, Track } from "./colorizer";
 import { DEFAULT_CATEGORICAL_PALETTE_ID, DEFAULT_CATEGORICAL_PALETTES } from "./colorizer/colors/categorical_palettes";
 import { DEFAULT_COLOR_RAMP_ID, DEFAULT_COLOR_RAMPS } from "./colorizer/colors/color_ramps";
-import { defaultViewerConfig, FeatureThreshold, isThresholdNumeric, ViewerConfig } from "./colorizer/types";
+import {
+  defaultViewerConfig,
+  FeatureThreshold,
+  getDefaultScatterPlotConfig,
+  isThresholdNumeric,
+  ScatterPlotConfig,
+  ViewerConfig,
+} from "./colorizer/types";
 import { getColorMap, getInRangeLUT, thresholdMatchFinder, validateThresholds } from "./colorizer/utils/data_utils";
 import { numberToStringDecimal } from "./colorizer/utils/math_utils";
 import { useConstructor, useDebounce } from "./colorizer/utils/react_utils";
@@ -61,6 +68,10 @@ function App(): ReactElement {
   const [config, updateConfig] = useReducer(
     (current: ViewerConfig, newProperties: Partial<ViewerConfig>) => ({ ...current, ...newProperties }),
     defaultViewerConfig
+  );
+  const [scatterPlotConfig, updateScatterPlotConfig] = useReducer(
+    (current: ScatterPlotConfig, newProperties: Partial<ScatterPlotConfig>) => ({ ...current, ...newProperties }),
+    getDefaultScatterPlotConfig()
   );
 
   const [isInitialDatasetLoaded, setIsInitialDatasetLoaded] = useState(false);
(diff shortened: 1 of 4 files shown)


src/components/Tabs/ScatterPlotTab.tsx
```diff
@@ -142,26 +139,56 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+    useEffect(() => {
+      if (propValue !== value) {
+        onChange && onChange();
+        startTransition(() => {
+          _setValue(propValue);
+        });
+      }
+    }, [propValue]);
     return [value, setState];
   };
```
ShrimpCryptid, line 175: Would like a gut check here on if I made this too complicated, or if I could make the comments clearer.

src/components/LoadingSpinner.tsx
ShrimpCryptid, line 1: Increased delay before the spinner appears from 0.1 to 0.2 seconds, to reduce flickering

src/App.tsx
```diff
@@ -62,6 +69,10 @@ function App(): ReactElement {
     (current: ViewerConfig, newProperties: Partial<ViewerConfig>) => ({ ...current, ...newProperties }),
     defaultViewerConfig
   );
+  const [scatterPlotConfig, updateScatterPlotConfig] = useReducer(
+    (current: ScatterPlotConfig, newProperties: Partial<ScatterPlotConfig>) => ({ ...current, ...newProperties }),
+    getDefaultScatterPlotConfig()
+  );
```
ShrimpCryptid, line 75: The `ViewerConfig` (just above the fold here) uses the same pattern. Using `useReducer` allows `scatterPlotConfig` to be called multiple times in the same state update and reflect all of those changes in the next.

toloudis: do you mean to say "updated multiple times in the same state update"?

ShrimpCryptid: Yes :X

src/colorizer/types.ts
```diff
@@ -167,3 +167,22 @@ export const defaultViewerConfig: ViewerConfig = {
   outOfRangeDrawSettings: { mode: DrawMode.USE_COLOR, color: new Color(OUT_OF_RANGE_COLOR_DEFAULT) },
   outlierDrawSettings: { mode: DrawMode.USE_COLOR, color: new Color(OUTLIER_COLOR_DEFAULT) },
 };
+
+export enum RangeType {
```
toloudis, line 171: Is this going to be another "permanent ids can never be changed" once you serialize this to url?

toloudis (approved):

src/colorizer/types.ts
ShrimpCryptid: We'll need to always ensure backwards compatibility for these specific values in the URL, but we can always add a layer of indirection here or remap them as needed.

ShrimpCryptid: See here: https://github.com/allen-cell-animated/nucmorph-colorizer/pull/233/files#diff-49ae85ec948625f41a63acb8324ff556e601766969a1c26ac5c44c28c35ab61aR318-R328

src/components/Tabs/ScatterPlotTab.tsx
```diff
@@ -765,13 +754,15 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
           label={"X"}
           selected={xAxisFeatureName || ""}
           items={menuItems}
-          onChange={setXAxisFeatureName}
+          onChange={(key) => props.updateScatterPlotConfig({ xAxis: key })}
```
rugeli, line 757: I like the refactoring of this handler to directly call `props.updateScatterPlotConfig` with the new key!

rugeli (approved):
Looks good to me! First time reviewing frontend work, it's great to learn how you handled state synchronization and prop updates.

PR #233, opened 2024-02-24
feature: Saving current scatterplot to the URL

### LINKED ISSUE #182: Save Scatter Plot settings to URL (+current open tab)

ShrimpCryptid:
Problem
=======
Closes #182! Saves the currently open tab and the scatterplot configuration to the URL.

*Expected review size: medium, ~20-30 minutes*

Solution
========
- Adds serializing and deserializing for the `ScatterPlotConfig`, allowing it to be saved and retrieved from the URL.
- Adds the currently open tab to the `ViewerConfig`, and saves it to and from the URL.

## Type of change

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-233/?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama%20Bear&feature=Volume&track=602&t=123&color=matplotlib-cool&palette-key=adobe&bg-sat=100&bg-brightness=100&fg-alpha=100&outlier-color=c0c0c0&outlier-mode=1&filter-color=dddddd&filter-mode=1&tab=scatter_plot&scalebar=1&timestamp=1&path=1&keep-range=0&scatter-range=frame&scatter-x=Height&scatter-y=Volume
2. The scatterplot tab should already be open, with "Height" and "Volume" as the two axes. The range should be set to "Current frame."

image

Thanks for contributing!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -120,7 +120,6 @@ function App(): ReactElement {
   }, [dataset, featureThresholds]);
 
   const [playbackFps, setPlaybackFps] = useState(DEFAULT_PLAYBACK_FPS);
-  const [openTab, setOpenTab] = useState(TabType.TRACK_PLOT);
 
   // Provides a mounting point for Antd's notification component. Otherwise, the notifications
   // are mounted outside of App and don't receive CSS styling variables.
@@ -917,8 +916,8 @@ function App(): ReactElement {
                 type="card"
                 style={{ marginBottom: 0, width: "100%" }}
                 size="large"
-                activeKey={openTab}
-                onChange={(key) => setOpenTab(key as TabType)}
+                activeKey={config.openTab}
+                onChange={(key) => updateConfig({ openTab: key as TabType })}
                 items={[
                   {
                     label: "Track Plot",
@@ -949,7 +948,7 @@ function App(): ReactElement {
                           selectedTrack={selectedTrack}
                           findTrack={findTrack}
                           setFrame={setFrameAndRender}
-                          isVisible={openTab === TabType.SCATTER_PLOT}
+                          isVisible={config.openTab === TabType.SCATTER_PLOT}
                           isPlaying={timeControls.isPlaying() || isRecording}
                           selectedFeatureName={featureName}
                           colorRampMin={colorRampMin}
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -119,7 +120,6 @@ function App(): ReactElement {
   }, [dataset, featureThresholds]);
 
   const [playbackFps, setPlaybackFps] = useState(DEFAULT_PLAYBACK_FPS);
-  const [openTab, setOpenTab] = useState(TabType.TRACK_PLOT);
```
ShrimpCryptid, line 122: `openTab` is now part of `ViewerConfig` (`config`)

src/colorizer/types.ts
```diff
@@ -134,6 +134,18 @@ export type DrawSettings = {
   color: Color;
 };
 
+// CHANGING THESE VALUES CAN POTENTIALLY BREAK URLs. See `url_utils.parseDrawSettings` for parsing logic.
+export enum TabType {
+  FILTERS = "filters",
+  TRACK_PLOT = "track_plot",
+  SCATTER_PLOT = "scatter_plot",
+  SETTINGS = "settings",
+}
```
ShrimpCryptid, line 143: There is a matching unit test that checks for these explicit string values, to make sure they don't accidentally get changed in the future.

kmitcham (approved):
Looks good to me.  What is the correct/expected behavior if I go to a plot based on a URL, and update the plot values on the page? Should the URL update, or just not be in sync with the display until serialize is called?

ShrimpCryptid:
> Looks good to me. What is the correct/expected behavior if I go to a plot based on a URL, and update the plot values on the page? Should the URL update, or just not be in sync with the display until serialize is called?

The URL should always update in sync with changes on the page. The serialization of the current config is triggered [here](https://github.com/allen-cell-animated/nucmorph-colorizer/pull/233/files#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2R224-R225), where any changes to `scatterPlotConfig` causes the URL updating code to fire.

ascibisz (approved):
Code looks good to me! I'm glad you wrote tests, I also tested manually and everything worked as expected for me

PR #224, opened 2024-02-20
feat: Help menu

ShrimpCryptid:
Problem
=======
First part of implementation for #184.

*Expected review time: small, <15 min*

Solution
========
- Adds the `HelpDropdown` component. Currently only includes an issue reporting link.
- Adds the `VisuallyHidden` accessibility component, which adds text that can be read by screen readers.

## Type of change

* New feature (non-breaking change which adds functionality)

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-224/ and try clicking on the help button.

Screenshots (optional):
-----------------------
image

PR #237, opened 2024-02-27
refactor: Made new shared `AccessibleDropdown` component. 

ShrimpCryptid:
Problem
=======
The newly-added Help menu (see #224) had some interactivity issues, where clicking on it would cause the menu to immediately close due to a conflict with hovering behavior. It was enough of a push to make me completely refactor the Dropdowns into a much nicer `AccessibleDropdown` component.

*Estimated review time: large, 60 minutes (faster if comparing with original files)*

Solution
========
This PR is pretty large because Git didn't recognize some of the file renames, despite my best attempts. When I post this, I'm going to talk through this with the requested reviewers because the size is a bit absurd.

This change takes apart the old `LabeledDropdown`, which had three responsibilities:
1. Allowing for selection of a single element from a dropdown -> now `SelectionDropdown`
2. Allowing for keyboard navigation and clicking -> now `AccessibleDropdown`
3. Styling of Dropdown elements -> now `DropdownItem`

The `AccessibleDropdown` is now the root of most of the dropdowns used in the viewer, with `SelectionDropdown` being a nice wrapper around it for rendering selectable dropdown lists.

## Type of change

* Internal refactor

Change summary:
---------------
* Something else I/we did

Screenshots (optional):
-----------------------

-----------------------
1. AccessibleDropdown
2. DropdownItem
3. SelectionDropdown


src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid, line 1: I recommend comparing against the deleted `LabeledDropdown.tsx` file. Most of the core pieces are the same (the clicking behavior and styling), with a few new props to make `AccessibleDropdown` easier to override in areas of styling.

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,229 @@
...
+  /** Text label to include with the dropdown. If null or undefined, hides the label. */
+  label?: string | null;
+
+  /** Contents to be rendered inside the dropdown. */
+  dropdownContent: ReactElement;
+  renderDropdownContent?: ((content: ReactElement) => ReactElement) | null;
+  disabled?: boolean;
+  /** The type of button to render for the dropdown. See Antd's button types:
+   * https://ant.design/components/button#components-button-demo-basic */
+  buttonType?: "primary" | "default" | "outlined" | "dashed" | "text" | "link";
```
ShrimpCryptid, line 19: `outlined` is not a recognized Ant type-- it's similar to `default` but with grey outlines instead of Ant's default purple.

ShrimpCryptid: *Default:*
image

*Outlined:*
image

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,233 @@
+import { Button, Dropdown, Tooltip } from "antd";
+import useToken from "antd/es/theme/useToken";
```
meganrm, line 2: why not import `import { theme } from 'and'; const { useToken} = theme;`?

src/components/Dropdowns/HelpDropdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+    return () => {
+      window.open(link, "_blank", "noopener noreferrer");
+    };
+  };
+
+  const dropdownContent = (
+    <DropdownItemList>
+      <DropdownItem
+        key={"issues"}
+        onClick={makeOnButtonClick("https://github.com/allen-cell-animated/nucmorph-colorizer/issues")}
```
meganrm, line 21: why isn't this just wrapping a link element?

src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid: Fixed, thank you!

src/components/Dropdowns/HelpDropdown.tsx
ShrimpCryptid: I made these use an `onClick` because originally the help links didn't respond to keyboard-based click interactions. It was also easier to style them like the other dropdowns by making them into buttons.

src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid: Yes, fixed!

src/components/Dropdowns/HelpDropdown.tsx
ShrimpCryptid: Mm, correction, I think I want this to be a link element too now. I'd like for the user to be able to see what link they'll be sent to.

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,242 @@
...
+          <div>
+            {props.buttonText}
+            <VisuallyHidden>(click to open menu)</VisuallyHidden>
+          </div>
+          <DropdownSVG />
+        </MainButtonContents>
+      </MainButton>
+    );
+  };
+  const renderButton = props.renderButton || defaultRenderButton;
```
interim17, line 202: Given that none of the current implementations are providing a `renderButton` prop could you reduce the complexity in this file by just having the default, or do you anticipate instances that will need custom button rendering functions?

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const defaultRenderDropdownContent = (content: ReactElement): ReactElement => {
+    // Fake the menu background styling
+    const dropdownStyle: React.CSSProperties = {
+      backgroundColor: token.colorBgElevated,
+      borderRadius: token.borderRadiusLG,
+      boxShadow: token.boxShadowSecondary,
+    };
+    return <DropdownContentWrapper style={dropdownStyle}>{content}</DropdownContentWrapper>;
+  };
+  const renderDropdownContent = props.renderDropdownContent || defaultRenderDropdownContent;
```
interim17, line 215: Same question as above here, and I acknowledge this isn't part of the change set, but as it is a big PR I'm struck by the fact that there some of these optional props aren't used anywhere currently.

On the one hand I dig that thoroughness, and it has me thinking about building a shared set of components for use in different AICS projects, it could be really useful when trying to make Simularium more accessible if we could port accessible components from NucMorph over there and their generality is an asset in that case. But could also be a YAGNI situation?

interim17:
These menus are working for me and look good. The accessibility is a great feature, nice work. I approve. Non-blocking comments and questions above.

interim17 (approved):

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,242 @@
...
+  renderButton: null,
+  onButtonClicked: () => {},
+};
+
+/** Contains the main dropdown button and label text */
+const MainContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  width: max-content;
```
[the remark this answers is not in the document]
ShrimpCryptid: I think this was a hold-over from old CSS I had written-- I've tested this and agree that it doesn't seem to be doing anything. I've removed it!

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,242 @@
...
+        }
+      `;
+    }
+    return;
+  }}
+`;
+
+/** Container for the text label and dropdown arrow SVG inside the main dropdown-triggering button. */
+const MainButtonContents = styled.div`
+  width: 100%;
```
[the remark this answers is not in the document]
ShrimpCryptid: I originally thought this would help with the `justify-content: space-between;` attribute by allowing the contents to take up the full width, but agreed, it looks like this isn't doing anything. The `width: 100%` in the div selector below does prevent some strange formatting bugs though.

src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid: Hmm, I think because both of the optional props I've included don't increase the complexity by a *huge* amount, I'm willing to leave them as-is. It was useful to think about making the construction steps of the dropdown more modular.

Admittedly, I didn't use either of the optional props in #244, which is where I expected to need them, though.

src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid: Definitely up for debate. I'm comfortable with leaving them in I think, because I do anticipate situations where custom components may want to adjust the rendering style.

src/components/Dropdowns/AccessibleDropdown.tsx
frasercl: Continuing from my comment above, this could become a separate component outside the body of this one, with a type that it shares with the type of that field in `AccessibleDropdownProps`. 100% optional suggestion, just a thought.

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+  /** Contents to be rendered inside the dropdown. */
+  dropdownContent: ReactElement;
+  renderDropdownContent?: ((content: ReactElement) => ReactElement) | null;
+  disabled?: boolean;
+  /**
+   * The type of button to render for the dropdown. See Antd's button types:
+   * https://ant.design/components/button#components-button-demo-basic
+   */
+  buttonType?: ButtonProps["type"] | "outlined";
```
frasercl, line 20: Nit: looks like you could `import { ButtonType } from "antd/es/button"` to replace the `ButtonProps` import and save an index. 100% a matter of your preference though.

frasercl (approved):
Asked some questions and made some suggestions, but all are ignorable at your discretion.

(No version of "looks good!" that I can think of fits the rhyme above)

src/components/Dropdowns/SelectionDropdown.tsx
```diff
@@ -0,0 +1,119 @@
...
+  /** An array of ItemType that describes the item properties (`{key, label}`),
+   * or an array of strings. Dropdown items will be presented in the provided array order.
+   *
+   * If a string array is provided, ItemType objects will be
+   * auto-generated with `key` and `label` values set to the string.*/
+  items: ItemType[] | string[];
+  disabled?: boolean;
+  /** The type of button to render for the dropdown. See Antd's button types:
+   * https://ant.design/components/button#components-button-demo-basic */
+  buttonType?: "primary" | "default" | "outlined" | "dashed" | "text" | "link";
```
frasercl, line 22: Nit: could this be `ButtonProps["type"]` (or `ButtonType`) as well?

src/components/Dropdowns/DropdownItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+/** Styled Antd Button for use with Dropdown inputs. */
+export default function DropdownItem(inputProps: PropsWithChildren<DropdownItemProps>): ReactElement {
+  const props = { ...defaultProps, ...inputProps } as PropsWithChildren<Required<DropdownItemProps>>;
+  return (
+    <DropdownItemButton $selected={props.selected} key={props.key} type="text" onClick={() => props.onClick(props.key)}>
+      {props.children}
+    </DropdownItemButton>
+  );
+}
```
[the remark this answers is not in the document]
ShrimpCryptid: Yup, this export was primarily motivated by wanting to enforce `type="text"` in the props. I also split this out into its own file because I anticipated it would be used in a few other dropdowns (later updates use this for `HelpDropdown` and `SelectionDropdown` iirc).

src/components/Dropdowns/SelectionDropdown.tsx
ShrimpCryptid: Yes, thanks!

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -0,0 +1,240 @@
...
+   * @param onClick: The callback to open the dropdown.
+   *
+   * Overrides must call the provided `onClick` callback during click or interaction events for the
+   * dropdown to work correctly.
+   *
+   * Tooltips may not work as expected if the root element is not a button or does not take callbacks
+   *  as props for the following events: `onMouseEnter/Leave`, `onPointerEnter/Leave`,
+   * `onFocus`, or `onBlur`. See implementation of `IconButton.tsx` for an example.
+   */
+  renderButton?: ((props: AccessibleDropdownProps, isOpen: boolean, onClick: () => void) => ReactElement) | null;
```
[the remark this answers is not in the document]
ShrimpCryptid: It could... I think I slightly prefer leaving it as a function and not as a component, because then you don't need to create a component every time you want to override this callback.

src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid: I'm going to leave it as-is for now, since I slightly prefer not reaching into Ant's internals. Thank you for the comment though!

PR #251, opened 2024-03-05
fix: Fixed OpenGL warning from zero-sized texture

ShrimpCryptid:
Problem
=======
Fixes a warning @toloudis and I noticed yesterday, where OpenGL was warning about a zero-sized texture.

image

*Estimated review size: tiny, ~2 minutes*

Solution
========
- Adds a check for zero-sized textures for the `inRangeIds` data texture and skips the update if needed.

Steps to Verify:
----------------
1. Open the preview link and confirm from the console that there are no OpenGL warnings: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-251/

### DIFF
--- a/src/colorizer/ColorizeCanvas.ts
+++ b/src/colorizer/ColorizeCanvas.ts
@@ -80,7 +80,7 @@ const getDefaultUniforms = (): ColorizeUniforms => {
   const emptyFeature = packDataTexture([0], FeatureDataType.F32);
   const emptyOutliers = packDataTexture([0], FeatureDataType.U8);
   const emptyInRangeIds = packDataTexture([0], FeatureDataType.U8);
-  const emptyColorRamp = new ColorRamp(["black"]).texture;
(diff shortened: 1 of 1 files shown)


src/colorizer/ColorizeCanvas.ts
```diff
@@ -80,7 +80,7 @@ const getDefaultUniforms = (): ColorizeUniforms => {
   const emptyFeature = packDataTexture([0], FeatureDataType.F32);
   const emptyOutliers = packDataTexture([0], FeatureDataType.U8);
   const emptyInRangeIds = packDataTexture([0], FeatureDataType.U8);
-  const emptyColorRamp = new ColorRamp(["black"]).texture;
+  const emptyColorRamp = new ColorRamp(["#aaa", "#fff"]).texture;
```
ShrimpCryptid, line 83: This should never be visualized but added a nice neutral grey texture instead of a harsh black.

meganrm (approved):

ascibisz (approved):

PR #253, opened 2024-03-05
fix: Change video codec version

ShrimpCryptid:
*Estimated review size: Tiny, <5 minutes*

Updates the AVC1 video codec to actually support up to 100 MB/s, which is what the high quality video export mode is set to. (The previous version, `avc1.420032`, maxed out around 20 MB/s.)

Thanks to @interim17 for pointing this out!

## Type of change


rugeli (approved):
Nice!

toloudis (approved):
just curious, is there any noticeable performance difference with this change?

ShrimpCryptid:
> just curious, is there any noticeable performance difference with this change?

Nothing I've noticed yet. I think it's a constraint with our frame resolution but TBD if we get the canvas to be a larger size for video export.

Issue #250, opened 2024-03-05
Initial landing page implementation for colorizer


ShrimpCryptid:
Implement Lyndsay's design for the colorizer's landing page.

- [x] Implement page layout
- [x] Decide if the page should be the root URL (or `/landing`) -> leaning towards root URL for initial introduction, but will break all existing internal links...
- [x] Plumbing for routing existing app vs. landing page

toloudis:
opengraph tags
small screen size warning (separate ticket)

Issue #241, opened 2024-02-28
Add version info to help menu


ShrimpCryptid:
## Use Case

We would like to display version info like what Simularium does.

## Acceptance Criteria

Implement a simple dialog following the simularium model.  Subject to UX design.

Issue #239, opened 2024-02-28
Help menu option - About this tool


ShrimpCryptid:
## Use Case

_Please provide a use case to help us understand your request in context_

## Acceptance Criteria

_Please describe how you know this is done_

## Details

_Please provide any helpful specifications_

PR #257, opened 2024-03-08
refactor: Header component

ShrimpCryptid:
Problem
=======
Refactors the header into a reusable series of components as part of #250.

*Estimated review size: small, <15 minutes. (Recommend turning off whitespace)*

Solution
========
- Moves the header bar into its own `Header.tsx` component file, which pages can reuse.
  - We can't mount the header as part of the root component, because many of the buttons that go in the header (export, copy, etc.) are context-specific and require access to page state.
  - This component will eventually be used by the landing page.

## Type of change

Steps to Verify:
----------------
1. Open the preview link and confirm that the inputs and interactions in the header work: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-257/
2. Open the error page. The header should also appear: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-257/#/bad

Screenshots (optional):
-----------------------
image


src/assets/AICS-logo-and-name.svg
ShrimpCryptid, line 1: Edited the SVG's title attribute to say "Allen Institute for Cell Science"

src/Viewer.module.css
ShrimpCryptid, line 1: Moved to `Header.tsx`

meganrm:
Design feedback: it seems weird to have the help be a dropdown that only has one item. is that because you plan to have more in that dropdown in the future?

meganrm (approved):

ShrimpCryptid:
> the preview doesn't load the dataset (I get an error) is that expected?

It should be loading... Are you on VPN?

ShrimpCryptid:
> Design feedback: it seems weird to have the help be a dropdown that only has one item. is that because you plan to have more in that dropdown in the future?

Yup! There are a few issues in the backlog for adding additional options to the help dropdown. See #241  and #239.

frasercl (approved):
Looks good!

PR #260, opened 2024-03-11
feat: Initial landing page implementation

ShrimpCryptid:
Problem
=======
Initial implementation of #250. Implements most of the dataset-related content for the landing page, with placeholder examples.

*Estimated review size: medium, ~20-30 minutes (it's mostly example boilerplate; please turn off whitespace!)*

See Lyndsay's original designs here: https://www.figma.com/file/DcFEsrszbUOK0qU1E9wrH0/colorizer_UI_components_and_V1_design?type=design&node-id=1182-2603&mode=design&t=nBl9gboJtCdu79Dh-0

Solution
========
- Adds styled components to `LandingPage.tsx` for lists of `projects`, each of which can include one or more `datasets`.
  - Projects/datasets are stored in a separate file and rendered on `LandingPage`.
  - Custom styling for links, dataset items, "in review" flags, etc.
- Adds some example links and content to the landing page for testing.
  - These will be removed when we prepare for final release.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-260/
1. Click any of the first three datasets under the first project. They should open in the same window and load different views of the Nucmorph dataset. (You will need to be on the network to access them.)

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**
https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/890c81ab-4e7c-4c50-9e8e-022b2ce2dc71

Design:
image

Implementation:
image

-----------------------
1. Start with `src/types/index.tsx`!


src/routes/LandingPageContent.ts
ShrimpCryptid, line 1: Placeholder content until we have real datasets to include here.

src/routes/LandingPageContent.ts
```diff
@@ -0,0 +1,94 @@
...
+export const landingPageContent: ProjectEntry[] = [
+  {
+    name: "NucMorph dataset name which usually is a rather long name so may wrap to two lines",
+    description:
+      "Introductory explanatory text about the dataset(s) and anything a user should know before opening in app. This should ideally only be a couple of sentences.",
+    publicationLink: new URL("https://www.google.com"),
+    publicationName:
+      "This is the name of the associated publication that the user can click to open in a new tab (Publisher name, mm/dd/yyyy)",
+    loadLink:
+      "viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json&dataset=Mama+Bear&feature=Volume&color=matplotlib-cool",
```
meganrm, line 12: I'm curious about this load link. First it seems odd to me to include the viewer path if you always know that any dataset is going to get loaded at the viewer path. this also seems like it would be tedious to edit if you decided you wanted a different default color map, for example, and had a bunch of params in this link. I feel like this could be separated out into `defaultSettings` and `collectionUrl`. Is `dataset=Mama+Bear` just the dataset name, or does that get used to load some other data?

ShrimpCryptid: Oh, breaking this up makes sense, thank you for the suggestion! I think I can reuse an existing datastructure I have that serializes the state to URL parameters.

Yes, the `dataset` URL param determines which dataset in the collection to load.

ShrimpCryptid: Change has been made! See https://github.com/allen-cell-animated/nucmorph-colorizer/pull/260/files/a438f1b1f6223d4bd6eb2e8e5c5d43d55f34193d#diff-845b967d105e866efc166956c15fa7b2a2fab3dad00bdeb978f06a17c671da81.

I am a bit worried that this will make it harder for other people to make changes to the landing page content; while it's more maintainable for the developer side, it's a higher barrier of entry for everyone else. Any thoughts @toloudis ?

interim17 (approved):
LGTM

Issue #229, opened 2024-02-21
Opening dataset from URL does not update the recent datasets dropdown in Load menu


ShrimpCryptid:
## Description

_A clear description of the bug_

## Expected Behavior

_What did you expect to happen instead?_

## Reproduction

_A minimal example that exhibits the behavior._

## Environment

_Any additional information about your environment. E.g. OS version, python version, browser etc._

PR #268, opened 2024-03-15
feature: Load button on landing page

### LINKED ISSUE #267: Add load button to landing page header
## Use Case

Add load button to the header so datasets can be loaded from the landing page.

## Acceptance Criteria

- LoadButton needs to be able to validate loading the dataset and show the user an error message if there are problems with it. That means that the dataset loading code needs to also be accessed in the LandingPage.
- Refactor internals of `LoadButton` to support the above!
- Allow landing page to pass Dataset + Collections object to the viewer so data doesn't need to be loaded twice?

ShrimpCryptid:
Problem
=======
Closes #267, adding the load button to the landing page.

*Estimated size: small, 20 minutes*

Solution
========
This required a big refactor of the loading button! Before, it would pass out the URL and wait for the viewer to try loading it and report back whether it loaded successfully. This change moves the load into the logic for the button itself, so the loading code isn't replicated between the landing page and the viewer.

I made a lot of code moves as part of this change which I could potentially split up into smaller PRs if needed.

- Refactors LoadButton to perform the test loading step, and to pass out the loaded dataset through a callback.
- Adds the LoadButton to the landing page; loading a dataset will cause it to switch to the viewer.
- Added a nice React hook for keeping track of recently loaded datasets.
  - I want to use this in a later refactor for #229 

## Type of change

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-268/
2. Open the load dialog and try loading any of the following datasets:
  - https://raw.githubusercontent.com/allen-cell-animated/nucmorph-colorizer/main/documentation/example_dataset/manifest.json
  - https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test
  - https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/3500005828_25/ (single dataset)
4. Repeat step 2 on the viewer.

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -547,30 +547,11 @@ function Viewer(): ReactElement {
    * @throws an error if the URL could not be loaded.
    * @returns the absolute path of the URL resource that was loaded.
    */
-  const handleLoadRequest = useCallback(
-    async (url: string): Promise<string> => {
-      console.log("Loading '" + url + "'.");
-      const newCollection = await Collection.loadFromAmbiguousUrl(url);
-      const newDatasetKey = newCollection.getDefaultDatasetKey();
-      const loadResult = await newCollection.tryLoadDataset(newDatasetKey);
-      if (!loadResult.loaded) {
-        const errorMessage = loadResult.errorMessage;
-
-        if (errorMessage) {
-          // Remove 'Error:' prefixes
-          const matches = errorMessage.replace(/^(Error:)*/, "");
-          // Reject the promise with the error message
-          throw new Error(matches);
-          // throw new Error(errorMessage);
-        } else {
-          throw new Error();
-        }
-      }
-
+  const handleDatasetLoad = useCallback(
+    (newCollection: Collection, newDatasetKey: string, newDataset: Dataset): void => {
       setCollection(newCollection);
       setFeatureThresholds([]); // Clear when switching collections
-      await replaceDataset(loadResult.dataset, newDatasetKey);
-      return newCollection.url || newCollection.getDefaultDatasetKey();
+      replaceDataset(newDataset, newDatasetKey);
     },
     [replaceDataset]
   );
@@ -710,7 +691,7 @@ function Viewer(): ReactElement {
         {/* <h3>Dataset Name</h3> */}
         <FlexRowAlignCenter $gap={12}>
           <FlexRowAlignCenter $gap={2}>
-            <LoadDatasetButton onRequestLoad={handleLoadRequest} currentResourceUrl={collection?.url || datasetKey} />
+            <LoadDatasetButton onLoad={handleDatasetLoad} currentResourceUrl={collection?.url || datasetKey} />
             <Export
               totalFrames={dataset?.numberOfFrames || 0}
               setFrame={setFrameAndRender}
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -9,7 +9,7 @@ import {
 import { Checkbox, notification, Slider, Tabs } from "antd";
 import { NotificationConfig } from "antd/es/notification/interface";
 import React, { ReactElement, useCallback, useContext, useEffect, useMemo, useReducer, useRef, useState } from "react";
-import { useLocation, useSearchParams } from "react-router-dom";
+import { Location, useLocation, useSearchParams } from "react-router-dom";
 
 import { ColorizeCanvas, Dataset, Track } from "./colorizer";
 import {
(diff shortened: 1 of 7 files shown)


src/Viewer.tsx
```diff
@@ -547,30 +558,11 @@ function Viewer(): ReactElement {
...
-          // Remove 'Error:' prefixes
-          const matches = errorMessage.replace(/^(Error:)*/, "");
-          // Reject the promise with the error message
-          throw new Error(matches);
-          // throw new Error(errorMessage);
-        } else {
-          throw new Error();
-        }
-      }
-
```
ShrimpCryptid, line 569: Moved into `LoadDatasetButton.tsx`

src/colorizer/utils/react_utils.ts
ShrimpCryptid, line 1: Copied over from `LoadDatasetButton.tsx`

src/components/LoadDatasetButton.tsx
```diff
@@ -122,6 +108,37 @@ export default function LoadDatasetButton(props: LoadDatasetButtonProps): ReactE
...
+        // throw new Error(errorMessage);
+      } else {
+        throw new Error();
+      }
+    }
+
+    const resourceUrl = newCollection.url || newCollection.getDefaultDatasetKey();
+    return [resourceUrl, newCollection, loadResult.dataset];
+  };
+
```
ShrimpCryptid, line 141: Copied with almost no changes from `Viewer.tsx`

interim17 (approved):
This LGTM and was working earlier. Currently I'm getting failed to fetch errors, but I don think it's related to any code changes.

src/Viewer.tsx
```diff
@@ -423,19 +425,28 @@ function Viewer(): ReactElement {
...
-      const datasetParam = initialUrlParams.dataset;
       let datasetKey: string;
-
-      if (datasetParam && urlUtils.isUrl(datasetParam) && !collectionUrlParam) {
-        // Dataset is a URL and no collection URL is provided;
-        // Make a dummy collection that will include only this dataset
-        newCollection = Collection.makeCollectionFromSingleDataset(datasetParam);
-        datasetKey = newCollection.getDefaultDatasetKey();
+      // Check if we were passed a collection + dataset from the previous page.
+      if (location.state && "collection" in location.state && "datasetKey" in location.state) {
```
meganrm, line 430: Suggestion: this amount of conditionals is something I would usually move to a function that makes it clear what you're testing to make it easier to comprehend. ie `locationHasCollectionAndDataset: () => boolean` or something more succinct :)

src/components/LoadDatasetButton.tsx
```diff
@@ -219,11 +224,11 @@ export default function LoadDatasetButton(props: LoadDatasetButtonProps): ReactE
   });
 
   // Get the URLs (keys) of any recent datasets that match the currently selected urlInput.
-  const matchingKeys = recentDatasets.filter(({ label }) => label === urlInput).map(({ url }) => url);
+  const matchingKeys = recentCollections.filter(({ label }) => label === urlInput).map(({ url }) => url);
   const datasetsDropdownProps: MenuProps = {
     onClick: (info) => {
       // Set the URL input to the label of the selected dataset
-      const dataset = recentDatasets.find(({ url }) => url === info.key);
+      const dataset = recentCollections.find(({ url }) => url === info.key);
```
meganrm, line 231: just for my own knowledge: what's the definition of a collection and a dataset? It seems like there are both things, but here it seems like you're using the terms equivalently?

ShrimpCryptid: Oops, I totally forgot to update the `dataset` variable here, I'll make that change. The change here was made because `recentDatasets` is _usually_ storing collections instead of datasets (though maybe this should be `recentResources`?).

- A dataset is a single image sequence + feature/data JSON files, all indexed by a manifest JSON file. 
- A collection is a JSON file containing paths to one or more datasets-- e.g. a collection of datasets. When a collection is loaded, you can switch between its datasets in the dropdown.

The colorizer tool can load in **both** collections and datasets directly, though internally it treats both as collection objects. (If you load in a dataset, it wraps it in a dummy collection and treats it like a collection with only a single dataset inside.)

ShrimpCryptid: Some of the language here is legacy from when we used collections/datasets interchangeably. I usually prefer using the word "dataset" for external users since it's a bit more intuitive.

src/routes/LandingPage.tsx
```diff
@@ -99,11 +104,26 @@ const InReviewFlag = styled(FlexRowAlignCenter)`
...
+    navigate(PageRoutes.VIEWER, { state: { collection: collection, datasetKey: datasetKey } as LocationState });
+  };
+
+  // Rendering
 
   // TODO: Should the load buttons be link elements or buttons?
   // Currently both the link and the button inside can be tab-selected.
   const renderDataset = (dataset: DatasetEntry, index: number): ReactElement => {
-    const viewerLink = "viewer" + paramsToUrlQueryString(dataset.loadParams);
+    const viewerLink = PageRoutes.VIEWER + paramsToUrlQueryString(dataset.loadParams);
```
meganrm, line 126: personal preference: 
```suggestion
    const viewerLink = `${PageRoutes.VIEWER}${paramsToUrlQueryString(dataset.loadParams}`;
```

meganrm (approved):
looks good

PR #270, opened 2024-03-18
feature: Handle v1.1.0 collection JSON schema

### LINKED ISSUE #220: Display collection names on the UI
## Use Case

Add a display name for datasets or collections that can be shown on the UI.

## Acceptance Criteria

## Details

- Does a collection's name for a dataset override its manifest name?

### LINKED ISSUE #248: Handle collection JSON objects (as well as arrays)
## Use Case

Add compatibility to new collection JSON object schema, as described in `colorizer-data` https://github.com/allen-cell-animated/colorizer-data/issues/35

## Acceptance Criteria

- [ ] Collection loader should handle arrays and object collections data, converting to the object format as needed.
- [ ] Collection loader should load any metadata from the new collection JSON format.
- [ ] Add unit tests for both the old collection data and the new format.

## Details

ShrimpCryptid:
Problem
=======
Closes #248 and closes #220! Updates the collection loader to handle the new v1.1.0 collection JSON schema, as described in https://github.com/allen-cell-animated/colorizer-data/pull/37, and shows the collection names (if provided) on the interface.

*Estimated size: small, 20 minutes*

Solution
========
- Adds a `collection_utils` file similar to the one that exists for datasets, which updates older legacy JSON files and updates them to the current format.
- Adds unit tests for the new collection and dataset JSON files.

## Type of change

## Testing
1. Open this preview link and open the developer console: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-270/#/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Ftest_v1.1.0%2Fcollection.json
2. You should see the name of the collection in the top header, and the collection metadata should include additional information about the collection.

Displayed name:
image

Collection metadata:
image

### DIFF
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -1,19 +1,14 @@
 import { DEFAULT_COLLECTION_FILENAME, DEFAULT_DATASET_FILENAME } from "../constants";
+import {
+  CollectionEntry,
+  CollectionFile,
+  CollectionFileMetadata,
+  updateCollectionVersion,
+} from "./utils/collection_utils";
 import { DEFAULT_FETCH_TIMEOUT_MS, fetchWithTimeout, formatPath, isJson, isUrl } from "./utils/url_utils";
 
 import Dataset from "./Dataset";
 
-/**
- * Dataset properties in a collection. Collections are defined as .json files containing an array of objects
- * with the following properties:
- *
- * - `path`: a relative path from the base directory of the collection URL, or a URL to a dataset.
- * - `name`: the display name for the dataset.
- */
-export type CollectionEntry = {
-  path: string;
-  name: string;
-};
 export type CollectionData = Map<string, CollectionEntry>;
 
 export type DatasetLoadResult =
@@ -34,6 +29,7 @@ export type DatasetLoadResult =
  */
 export default class Collection {
   private entries: CollectionData;
+  public metadata: Partial<CollectionFileMetadata>;
   /**
    * The URL this Collection was loaded from. `null` if this Collection is a placeholder object,
    * such as when generating dummy Collections for single datasets.
@@ -47,9 +43,10 @@ export default class Collection {
    * @param url the optional string url representing the source of the Collection. `null` by default.
    * @throws an error if a `path` is not a URL to a JSON resource.
    */
-  constructor(entries: CollectionData, url: string | null = null) {
+  constructor(entries: CollectionData, url: string | null = null, metadata: Partial<CollectionFileMetadata> = {}) {
     this.entries = entries;
     this.url = url ? Collection.formatAbsoluteCollectionPath(url) : url;
+    this.metadata = metadata;
 
     // Check that all entry paths are JSON urls.
     this.entries.forEach((value, key) => {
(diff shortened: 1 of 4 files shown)


src/colorizer/Collection.ts
```diff
@@ -1,19 +1,14 @@
...
- * Dataset properties in a collection. Collections are defined as .json files containing an array of objects
- * with the following properties:
- *
- * - `path`: a relative path from the base directory of the collection URL, or a URL to a dataset.
- * - `name`: the display name for the dataset.
- */
-export type CollectionEntry = {
-  path: string;
-  name: string;
-};
```
ShrimpCryptid, line 16: Moved to `collection_utils.ts`

interim17 (approved):
Thanks for adding the testing section for making review easiser!

tests/Collection.test.ts
```diff
@@ -44,33 +45,94 @@ describe("Collection", () => {
...
+    ];
+
+    for (const { version, collectionJson } of collectionFormats) {
+      describe(version, () => {
+        it("collects datasets", async () => {
+          const url = "https://e.com/collection.json";
+          const mockFetch = makeMockFetchMethod(url, collectionJson);
+          const collection = await Collection.loadCollection(url, mockFetch);
+
+          expect(collection.getDatasetKeys().length).to.equal(5);
```
meganrm, line 83: small thing: but couldn't you do:
```suggestion
          expect(collection.getDatasetKeys().length).to.equal(datasets.length);
```
?
just in case you add more data to your mock data

tests/Collection.test.ts
```diff
@@ -44,33 +45,94 @@ describe("Collection", () => {
...
+        });
+
+        it("retrieves metadata", async () => {
+          const url = "https://e.com/collection.json";
+          const mockFetch = makeMockFetchMethod(url, collectionJson);
+          const collection = await Collection.loadCollection(url, mockFetch);
+
+          if (semver.lte(version, "v0.0.0")) {
+            expect(collection.metadata).deep.equals({});
+            return;
```
meganrm, line 114: this is a personal preference, but I would have made these two different tests: one that sends in v0.0.0 data and checks for the empty object. and one that sends in v1.1.0 data and has everything below. and the description would be "it returns an empty object if the version is 0.0.0" since it's not really retrieving any metadata in this case. Also is there a possibility of there not being a version or do you add it somewhere (I'm starting here in the review)

src/colorizer/Collection.ts
```diff
@@ -224,16 +222,22 @@ export default class Collection {
...
+      const json = await response.json();
+      collection = updateCollectionVersion(json);
+    } catch (e) {
+      throw new Error(`Could not parse collections JSON data from url '${absoluteCollectionUrl}': '${e}'`);
     }
 
     // Convert JSON array into map
+    if (collection.datasets.length === 0) {
+      throw new Error(
+        `Could not retrieve collections JSON data from url '${absoluteCollectionUrl}': No datasets found.`
```
meganrm, line 236: do your tests cover this path?

meganrm (approved):
Overall looks good. Since you wrote tests, I have suggestions for some improvements in the current tests. It also seems like you're only testing for cases where everything goes well, so maybe think about some edge cases, or at least catching the errors you know you're going to throw. I don't know the data well enough to suggest more edge cases myself

tests/Collection.test.ts
ShrimpCryptid: The versions are defined in a big array up above, so here it's checking versions of the data that I've explicitly labeled as `v0.0.0`, `v1.0.0`, and `v1.1.0`. While I think it makes sense to split these into different tests as the tests are now, I'm also anticipating that more backwards-compatible versions will get added in the future. Structuring this test as it is now (where it progressively tests more and more fields) makes sense to me for that specific use case.

src/colorizer/Collection.ts
ShrimpCryptid: Good catch, they don't! Adding that now.

tests/Collection.test.ts
ShrimpCryptid: Mm, taking another look at it, I think it makes sense to split these up for now. (or at least, to make `v0.0.0` its own case!)

PR #271, opened 2024-03-18
Sample AWS deploy workflow for timelapse.allencell.org 

griffinfujioka:
Problem
=======
This PR provides a sample AWS deploy workflow for timelapse.allencell.org and staging-timelapse.allencell.org: 

[#56382596](https://github.com/orgs/AllenCellSoftware/projects/10?pane=issue&itemId=56382596)

Solution
========
This PR demonstrates how to deploy timelapse.allencell.org by pushing data to S3 and invalidating the CloudFront cache.

## Type of change

Change summary:
---------------
* Create sample `aws-deploy.yml` Github Action workflow to demonstrate deployment of timelapse.allencell.org 

Steps to Verify:
----------------
After this PR is merged, we will be able to test the workflow more thoroughly by manually triggering the workflow and verifying that adequate permissions are in place for S3 publishing and invalidating of the CloudFront cache. 

After we have proven the above concept, then developers of this project may incorporate S3 publishing and invalidating of CloudFront cache into their Github Actions worfkflows.


.github/workflows/aws-deploy.yml
```diff
@@ -0,0 +1,69 @@
...
+jobs:
+    deploy:
+        runs-on: ubuntu-latest
+
+        # Dynamically set the environment variable based on the input above:
+        environment: ${{ github.event.inputs.environment }}
+
+        steps:
+          - name: Checkout
+            uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
```
ShrimpCryptid, line 38: Any reason why we're using a specific action and not the major version specifier?

ShrimpCryptid (approved):
Nothing blocking since this is an initial test. Thank you!

.github/workflows/aws-deploy.yml
```diff
@@ -0,0 +1,69 @@
...
+                - production
+            required: true
+            type: choice
+       
+env:
+  # TODO: Verify what these values are for the allen-cell-animated Github organization
+  AWS_ACCOUNT_ID: ${{ vars.AWS_PUBLIC_DATA_RELEASES_ACCOUNT_ID }}
+  AWS_REGION: ${{ vars.AWS_DEFAULT_REGION }}
+  STAGING_CLOUDFRONT_DISTRIBUTION_ID: ${{ secrets.STAGING_CLOUDFRONT_DISTRIBUTION_ID }}
+  STAGING_S3_BUCKET: s3://staging-timelapse.allencell.org
```
toloudis, line 21: I think both cfe and simularium use the patterm `staging.appname.allencell.org` with a dot instead of hyphen.

.github/workflows/aws-deploy.yml
```diff
@@ -0,0 +1,69 @@
+---
+name: Sample workflow to demonstrate AWS deployment including publishing data to S3 and invalidating a CloudFront distribution cache
+
+on: 
+  workflow_dispatch:
```
toloudis, line 5: @ShrimpCryptid  Just noting future work: we can add "automatic deploy to staging on tagged releases" and promoting staging to prod by manual workflow_dispatch later

.github/workflows/aws-deploy.yml
```diff
@@ -0,0 +1,69 @@
...
+              elif [ "${{ github.event.inputs.environment }}" == "staging" ]; then
+                echo "S3_BUCKET=${{ env.STAGING_S3_BUCKET }}" >> $GITHUB_ENV
+                echo "CLOUDFRONT_DISTRIBUTION_ID=${{ env.STAGING_CLOUDFRONT_DISTRIBUTION_ID }}" >> $GITHUB_ENV
+              else
+                echo "Invalid environment specified"
+                exit 1
+              fi          
+
+          - name: Write an arbitrary file to S3 - just to demonstrate how. 
+            run: aws s3 sync ./index.html ${{ env.S3_BUCKET }}
```
toloudis, line 66: Are there github actions that wrap all this up into a cleaner smaller footprint here?  Maybe we could fold the above 3-4 steps into our own...

griffinfujioka: I poked around a bit and am not feeling like there are great options out there for existing Github Actions to accomplish these steps.

The best option I found was this one: https://github.com/marketplace/actions/s3-sync

.github/workflows/aws-deploy.yml
griffinfujioka: It's standard practice for us as [suggested by Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) to ensure an immutable release. 

I think [this post summarizes the vulnerability](https://www.paloaltonetworks.com/blog/prisma-cloud/unpinnable-actions-github-security/)

ShrimpCryptid: Thank you! I should probably go back and update our existing actions, then.

.github/workflows/aws-deploy.yml
toloudis: Yeah that one looks reasonable although the codebase is about 4 yrs old.  If we wrapped that all up into our own we could reuse it in at least 4 places, I think.  I have never looked into building my own action that simply aggregates steps of other actions.

toloudis: "looks reasonable" = "has the kind of front end signature I would expect"

.github/workflows/aws-deploy.yml
toloudis: If that change requires another iteration or some discussion it need not hold up this PR

toloudis (approved):

anl (approved):

.github/workflows/aws-deploy.yml
griffinfujioka: Thanks, I ended up opening [`terraform-platform#1116`](https://aicsbitbucket.corp.alleninstitute.org/projects/SW/repos/terraform-platform/pull-requests/1116/overview) to change from `staging-timelapse.allencell.org` to `staging.timelapse.allencell.org`. Should be no big deal and now is a good time to get it right.

gmoynihan88 (approved):

PR #275, opened 2024-03-19
fix: Recent datasets includes datasets loaded from URL

### LINKED ISSUE #229: Opening dataset from URL does not update the recent datasets dropdown in Load menu

ShrimpCryptid:
Problem
=======
Closes #229, "Opening dataset from URL does not update the recent datasets dropdown in Load menu". Also makes some other fixes to the `LoadDatasetButton` and dataset loading behaviors.

*Estimated review size: small, 20 minutes*

Solution
========
I did some other refactors as part of this change, so this PR is not as focused as I would have liked to be. I tried to comment everywhere I made changes to justify my reasoning!

- The viewer page now saves loaded URLs to the recent datasets dropdown.
- Adds improved error messaging when loading datasets from the URL fails.
- Hides the dropdown if there are no recent datasets.
- Made the currently loaded dataset clickable.

## Type of change

Steps to Verify:
----------------
1. Open the preview link **in incognito/private mode**: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-275/#/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json
2. Click the Load button. The current collection should be visible in the recent datasets dropdown.
3. In the same incognito/private window, open this preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-275/#/viewer?dataset=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2F3500005828_25%2Fmanifest.json
4. **Click the Load button. Both datasets should appear in the Load button dropdown.** 
image

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -29,7 +29,7 @@ import {
 } from "./colorizer/types";
 import { getColorMap, getInRangeLUT, thresholdMatchFinder, validateThresholds } from "./colorizer/utils/data_utils";
 import { numberToStringDecimal } from "./colorizer/utils/math_utils";
-import { useConstructor, useDebounce } from "./colorizer/utils/react_utils";
+import { useConstructor, useDebounce, useRecentCollections } from "./colorizer/utils/react_utils";
 import * as urlUtils from "./colorizer/utils/url_utils";
 import { DEFAULT_COLLECTION_PATH, DEFAULT_PLAYBACK_FPS } from "./constants";
 import { FlexRowAlignCenter } from "./styles/utils";
@@ -72,6 +72,7 @@ function Viewer(): ReactElement {
   const [collection, setCollection] = useState<Collection | undefined>();
   const [dataset, setDataset] = useState<Dataset | null>(null);
   const [datasetKey, setDatasetKey] = useState("");
+  const [, addRecentCollection] = useRecentCollections();
 
   const [featureName, setFeatureName] = useState("");
   const [selectedTrack, setSelectedTrack] = useState<Track | null>(null);
@@ -444,8 +445,19 @@ function Viewer(): ReactElement {
           datasetKey = newCollection.getDefaultDatasetKey();
         } else {
           // Try loading the collection, with the default collection as a fallback.
-          newCollection = await Collection.loadCollection(collectionUrlParam || DEFAULT_COLLECTION_PATH);
-          datasetKey = datasetParam || newCollection.getDefaultDatasetKey();
+          try {
+            newCollection = await Collection.loadCollection(collectionUrlParam || DEFAULT_COLLECTION_PATH);
+            datasetKey = datasetParam || newCollection.getDefaultDatasetKey();
+          } catch (error) {
+            console.error(error);
+            notificationApi["error"]({
+              message: "Error loading dataset:",
+              description: (error as Error).message,
+              placement: "bottomLeft",
+              duration: 4,
+            });
+            return;
+          }
         }
       }
 
@@ -462,6 +474,9 @@ function Viewer(): ReactElement {
         });
         return;
       }
+
+      // Add the collection to the recent collections list
+      addRecentCollection({ url: newCollection.getUrl() });
 
       // TODO: The new dataset may be null if loading failed. See TODO in replaceDataset about expected behavior.
       if (!isInitialDatasetLoaded) {
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -200,6 +200,17 @@ export default class Collection {
     return this.formatDatasetPath(collectionDirectory + "/" + datasetPath);
   }
 
+  // TODO: Refactor how dummy collections store URLs? The URL should always be a valid resource maybe?
+  /**
+   * Returns a URL to the collection or dataset.
+   */
+  public getUrl(): string {
+    if (this.url === null) {
+      return this.entries.get(this.getDefaultDatasetKey())!.path;
+    }
+    return this.url;
+  }
+
   // ===================================================================================
   // Static Loader Methods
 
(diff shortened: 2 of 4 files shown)


src/Viewer.tsx
```diff
@@ -463,6 +475,9 @@ function Viewer(): ReactElement {
         return;
       }
 
+      // Add the collection to the recent collections list
+      addRecentCollection({ url: newCollection.getUrl() });
+
```
ShrimpCryptid, line 480: Lines that actually add the new behavior

src/Viewer.tsx
```diff
@@ -444,8 +445,19 @@ function Viewer(): ReactElement {
...
+          } catch (error) {
+            console.error(error);
+            notificationApi["error"]({
+              message: "Error loading dataset:",
+              description: (error as Error).message,
+              placement: "bottomLeft",
+              duration: 4,
+            });
+            return;
+          }
```
ShrimpCryptid, line 460: Better error handling when bad datasets are loaded in-- this will pop up an error notification.

src/colorizer/Collection.ts
ShrimpCryptid, line 1: Provides the resource URL used to load this collection initially. Normally, `this.url == null` for dummy collections, which are generated to hold a single dataset.

src/components/LoadDatasetButton.tsx
ShrimpCryptid, line 1: Hide recent collections dropdown when empty

src/Viewer.tsx
```diff
@@ -449,8 +450,19 @@ function Viewer(): ReactElement {
...
           // Try loading the collection, with the default collection as a fallback.
-          newCollection = await Collection.loadCollection(collectionUrlParam || DEFAULT_COLLECTION_PATH);
-          datasetKey = datasetParam || newCollection.getDefaultDatasetKey();
+          try {
+            newCollection = await Collection.loadCollection(collectionUrlParam || DEFAULT_COLLECTION_PATH);
+            datasetKey = datasetParam || newCollection.getDefaultDatasetKey();
+          } catch (error) {
+            console.error(error);
+            notificationApi["error"]({
+              message: "Error loading dataset:",
```
toloudis, line 459: does the url of the dataset come in here via the `error`?  Would it be good to add `collectionUrlParam` to the message explicitly here?

toloudis, line 459: does the url of the dataset come in here via the error? Would it be good to add collectionUrlParam to the message explicitly here?

src/colorizer/Collection.ts
```diff
@@ -200,6 +200,17 @@ export default class Collection {
     return this.formatDatasetPath(collectionDirectory + "/" + datasetPath);
   }
 
+  // TODO: Refactor how dummy collections store URLs? The URL should always be a valid resource maybe?
```
toloudis, line 203: I'm not understanding the TODO. What is a dummy collection?  Is that when user tried to load a dataset directly and you have to internally build a fake collection to hold it?

src/colorizer/utils/react_utils.ts
```diff
@@ -199,34 +199,50 @@ export function useScrollShadow(shadowColor: string = "#00000030"): {
...
 const MAX_RECENT_COLLECTIONS = 10;
 
-export type RecentCollection = {
+export type StoredRecentCollection = {
   /** The absolute URL path of the collection resource. */
   url: string;
   /**
    * The user input for the collection resource.
-   * If `null`, uses the existing label (if already in recent datasets) or reuse the URL (if new).
+   * If `undefined`, uses the existing label (if already in recent datasets) or reuses the URL (if new).
```
toloudis, line 207: if undefined is allowed, should the type declaration for `label` have a question mark in it?

src/colorizer/utils/react_utils.ts
```diff
@@ -199,34 +199,50 @@ export function useScrollShadow(shadowColor: string = "#00000030"): {
...
    */
   label: string;
 };
 
+export type RecentCollection = Partial<StoredRecentCollection> & {
+  url: string;
+};
```
toloudis, line 214: This looks like it is adding complexity to the code, for a very simple data structure.  I think the verbose explanatory comment you put in the PR is really needed in the code too.  From the names RecentCollection and StoredRecentCollection it's hard to tell why they differ by this one url property.  Or else just simplify the type to have all 3 fields and use class methods to keep it internally consistent?

toloudis: Naively, it seems like all you need is a function like `ensureLabelAndUrlConsistent`

src/colorizer/Collection.ts
ShrimpCryptid: Correct! Currently, when I create a dummy collection I set the collection's base URL to `null`, and the only URL data that is provided is in the single dataset it wraps. I think I did this so I wouldn't be duplicating the URL, but the only reason that `null` value is stored is so I can tell if it's a dummy collection. It would likely be much clearer to separate that behavior out into its own flag.

src/colorizer/utils/react_utils.ts
ShrimpCryptid: Updated!

src/colorizer/utils/react_utils.ts
ShrimpCryptid: I ended up just removing the duplicated datastructure and just allowing the label to be undefined. This fortunately did not cause as many problems as I thought it would.

src/Viewer.tsx
ShrimpCryptid: It usually does, though it'll depend on the error message. The console log statement will have it though.
image

ShrimpCryptid: image

toloudis (approved):

interim17 (approved):
LGTM :)

PR #280, opened 2024-03-21
fix: Fixed bug where dropdown would render over modals

### LINKED ISSUE #276: Color ramp dropdown renders over modals
## Description
Reported by @interim17!

image
If clicking very fast between the color ramp dropdown and the load/export modals, you can briefly see the dropdown render over the other elements.

## Expected Behavior

Color ramp dropdown should render under the modals.

ShrimpCryptid:
Problem
=======

Closes #276, a bug where the color map dropdown would render over the modals.

image
*Horrifying!*

Solution
========
The issue was with the modals themselves. They were being contained within a div along with the buttons, which meant that they were in a [**stacking context**](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) with the buttons that prevented them from overlaying the rest of the UI. (Elements in a stacking context can apply z-indexing to change their ordering ONLY within the context.)

Removing the container also didn't work because it causes them to live outside of the styles and CSS variables given by `AppStyle`.

The solution was to add a container reference in `AppStyle` that can be  accessed through React's `Context` API. This means that modals can still be styled using rules defined in `AppStyle`, but aren't confined to the z-sorting of their root triggering buttons.

image 

### Changes:

- Add a `DocumentContext` containing an `HTMLDivElement` ref to a modal container.
- Allow `AppStyle` to provide a value for `DocumentContext`'s modal container ref.
- Update the existing modals to use the new context and ref.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-280/
2. Click on the color ramp dropdown button to pin it open.
3. Quickly click the Load or Export buttons. The color ramp dropdown should appear behind the modal.

Screenshots (optional):
-----------------------
| Before | After |
| --- | --- |
| image | image |


src/components/AppStyle.tsx
```diff
@@ -256,6 +261,18 @@ const CssContainer = styled.div`
...
+  // Force a rerender so the context provider updates with the new div ref; otherwise
+  // the value will always be null.
+  const [, setForceRerender] = useState(false);
+  useEffect(() => {
+    setForceRerender(true);
+  }, []);
+
   return (
     <CssContainer className={props.className}>
       <link rel="preconnect" href="https://fonts.gstatic.com" />
```
meganrm, line 278: curious why this is here and not in your html file. same with the font import above

meganrm (approved):
Looks like you fixed the bug. I am curious why you don't have AppStyle at a higher level? or is there no level you can get it high enough that it would have wrapped the modal?

src/components/AppStyle.tsx
```diff
@@ -256,6 +261,18 @@ const CssContainer = styled.div`
...
+  // other elements in the app, and lets us use the <Modal> component from Ant instead
+  // of the static modal API through Ant.App, which is less flexible.
+  const modalContainerRef = useRef<HTMLDivElement>(null);
+
+  // Force a rerender so the context provider updates with the new div ref; otherwise
+  // the value will always be null.
+  const [, setForceRerender] = useState(false);
+  useEffect(() => {
+    setForceRerender(true);
+  }, []);
```
frasercl, line 274: This mechanism to immediately trigger a second render after the initial render is a bit janky. A bit of research suggests a slightly nicer way to accomplish this might be through a [`ref` callback](https://react.dev/reference/react-dom/components/common#ref-callback), like so:

```tsx
const [modalContainer, setModalContainer] = useState(undefined);
// ...
return (<div ref={setModalContainer}>);
```

This has the added bonus of tying updates of `modalContainer` to the DOM element it actually captures, so if for some reason React decides to destroy and recreate that `div` during the lifecycle of this component, the reference will stay current.

I'm still not a huge fan of re-rendering this component, and potentially many of its children, twice (four times in dev mode!) on mount, but that might be the price of a fix here.

src/components/Export.tsx
```diff
@@ -445,7 +445,7 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
         centered={true}
         // Don't allow cancellation of modal by clicking off it when the recording is happening
         maskClosable={!isRecording}
-        getContainer={modalContextRef.current || undefined}
+        getContainer={modalContainerRef || undefined}
         footer={modalFooter}
       >
```
frasercl, line 450: Would it make any sense at all to create a shared `Modal` wrapper that captures and uses this context, rather than having to remember to do this for every current and future `Modal`? Absolutely valid to say "no" given how many more modals we expect to use (or not) and/or how much more complicated modals may (or may not) get.

ShrimpCryptid:
> Looks like you fixed the bug. I am curious why you don't have AppStyle at a higher level? or is there no level you can get it high enough that it would have wrapped the modal?

Yeah, I think Ant makes modals direct children of the document `body`, so there isn't a way to do it unless I directly style the root.

src/components/Export.tsx
ShrimpCryptid: Yeah, that makes sense, I'll take a stab at this and see if I can get it to work as expected. We'll still have to do it manually for any use of the static API (see #289).

src/components/AppStyle.tsx
ShrimpCryptid: I wanted to have the font family variables all live in one place for DRY reasons, though I know it's non-standard... This may also help if we ever want to embed the viewer as a component somewhere, though there's currently no plans to do so. Do you think this is worth bumping to `index.html`?

src/components/Export.tsx
```diff
@@ -115,10 +117,9 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
   // Note: notification API seems to only place notifications at the top-level under the
   // <body> tag, which causes some issues with styling.
   const { modal, notification } = App.useApp();
-  const modalContextRef = useRef<HTMLDivElement>(null);
 
   const originalFrameRef = useRef(props.currentFrame);
-  const [isLoadModalOpen, _setIsLoadModalOpen] = useState(false);
+  const [isModalOpen, _setIsModalOpen] = useState(false);
```
ShrimpCryptid, line 122: Rename here, since `isLoadModalOpen` was a holdover from when I copied this code from `LoadButton.tsx`.

src/components/Export.tsx
ShrimpCryptid: Fixed! I can't find a way to elegantly do this for the static modal calls, though I'd like to. I could provide styled versions of them (`styledStaticModal.confirm`, etc.) but I wouldn't have a way to access the ref since they don't exist inside a React component and I therefore can't use `useContext`... I've seen a few places where people assign an `id` to the wrapper container, but I'm not sure if that's a robust solution.

https://github.com/allen-cell-animated/nucmorph-colorizer/pull/280/files#diff-52e0fc4234b72537ec0c614472063a5c4b962ea1cffc66fd8a92e69616d0d571R215-R227

frasercl (approved):
Still wish there were a way to get styling through to modals without adding to React lifecycle stuff, but I trust this is the best way reasonably available

PR #282, opened 2024-03-22
feature: AWS Deployment

ShrimpCryptid:
Problem
=======
Updates #279 to push the build files to the S3 bucket, based on @griffinfujioka's template!

Test staging deploy is available here: https://staging.timelapse.allencell.org

Solution
========
- Adds an extra `build` job to the GitHub action, which creates the build files and saves them as an artifact.
- Updates the `deploy` job to retrieve and sync the build files to S3.
(This is how we are currently doing it for GitHub pages!)

## Type of change

### DIFF
--- a/.github/workflows/aws-deploy.yml
+++ b/.github/workflows/aws-deploy.yml
@@ -1,5 +1,5 @@
 ---
-name: AWS Deploy Example
+name: AWS Deployment
 
 on: 
   workflow_dispatch:
--- a/.github/workflows/aws-deploy.yml
+++ b/.github/workflows/aws-deploy.yml
@@ -26,6 +26,26 @@ permissions:
     contents: write  # Required for actions/checkout and OIDC tokens
 
 jobs:
+    build:
+      runs-on: ubuntu-latest
+
+      steps:
+        - name: Checkout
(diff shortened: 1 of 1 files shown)


griffinfujioka (approved):

.github/workflows/aws-deploy.yml
```diff
@@ -26,15 +26,33 @@ permissions:
...
+          run: npm ci && npx vite build
+  
+        - name: Upload build files
+          uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3
+          with:
+            name: aws-deploy-files
+            path: ./dist
+    
     deploy:
         runs-on: ubuntu-latest
```
griffinfujioka, line 51:
```suggestion
        runs-on: ubuntu-latest
        needs: build
```

Per [these docs](https://docs.github.com/en/actions/using-workflows/about-workflows#creating-dependent-jobs), since the default behavior is for jobs to run in parallel, then this workflow currently fails because the `deploy` job runs before the `build` job has created the artifact.

ShrimpCryptid: Just fixed this, thanks for reminding me that I can test the deployment without needing to merge it :X

toloudis (approved):

PR #289, opened 2024-03-28
feature: Small screen warning

### LINKED ISSUE #259: Add a small screen size warning
Image

ShrimpCryptid:
Problem
=======
Closes #259, adding a small screen warning for widths below 768 pixels (which is standard tablet size).

Solution
========
This one is a little more complicated than I would have liked, because Ant's `Modal` component doesn't provide an icon prop like the static API does. This means that I have to track and puppet the static API modal to make it behave like a component would.

- Adds a modal that appears when the window width dips below 768 pixels.
  - Once dismissed, the modal will not reappear until the width goes above 768 pixels.
  - The modal will automatically dismiss itself if the width returns to above 768 pixels.
- When the "Don't show this message again" checkbox is checked, the modal will not reappear.
- Adjusts modal styling based on Lyndsay's feedback.

## Type of change

Steps to Verify:
----------------
(see attached video!)
1. Open the preview link in an incognito/private window: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-289/
2. Turn on the developer console, and decrease the width of the window area.
3. At 768 pixels, the warning should appear. Increasing the window area above 768 pixels should cause it to disappear.
4. Trigger the warning again, and check the "Don't show this message again" box. The box should not reappear the next time the window width decreases below 768 pixels.

Screenshots (optional):
-----------------------

Demo video (🔊):

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/0ee23fa1-cff6-4133-8833-19c36d609e78


src/components/AppStyle.tsx
ShrimpCryptid, line 1: Minor fixes to modal header font styling here!

rugeli:
Thank you for the demo video, it was very insightful! The implementation of the checkbox to store the user's preference has made me think about the following: 
- are there scenarios where users might wish to reset this preference, or potentially other preferences?
- might there be a need to have something like a settings page to accommodate adjustments?

ShrimpCryptid:
> Thank you for the demo video, it was very insightful! The implementation of the checkbox to store the user's preference has made me think about the following:
> 
> * are there scenarios where users might wish to reset this preference(in this case, reverting the `allowSmallScreenWarning` flag back to true), or potentially other preferences?
> * might there be a need to have something like a settings page to accommodate adjustments?

I don't think so... Once a user knows that the app isn't optimized for small screens, we probably don't need to show it again in the future. It's more of an expectation of communications than a serious warning, I think.

rugeli:
> > Thank you for the demo video, it was very insightful! The implementation of the checkbox to store the user's preference has made me think about the following:
> > 
> > * are there scenarios where users might wish to reset this preference(in this case, reverting the `allowSmallScreenWarning` flag back to true), or potentially other preferences?
> > * might there be a need to have something like a settings page to accommodate adjustments?
> 
> I don't think so... Once a user knows that the app isn't optimized for small screens, we probably don't need to show it again in the future. It's more of an expectation of communications than a serious warning, I think.

Absolutely, that makes sense. I was just thinking about situations where we might implement more preferences, what if a user accidentally checks some boxes or later switches to a different display, things like that.

ShrimpCryptid:
> > > Thank you for the demo video, it was very insightful! The implementation of the checkbox to store the user's preference has made me think about the following:
> > > 
> > > * are there scenarios where users might wish to reset this preference(in this case, reverting the `allowSmallScreenWarning` flag back to true), or potentially other preferences?
> > > * might there be a need to have something like a settings page to accommodate adjustments?
> > 
> > 
> > I don't think so... Once a user knows that the app isn't optimized for small screens, we probably don't need to show it again in the future. It's more of an expectation of communications than a serious warning, I think.
> 
> Absolutely, that makes sense. I was just thinking about situations where we might implement more preferences, what if a user accidentally checks some boxes or later switches to a different display, things like that.

The small screen size warning will need to be dismissed in each new browser the user has, since we don't have a way to track it to a shared user preference or anything. (That would require us to make user logins, which, 😱)

In terms of other preferences, everything else is currently saved under the "Viewer settings" tab on the UI. In the future we could probably nest additional settings there as needed.

interim17 (commented):
Can you explain a little bit more about what this means: "Ant's Modal component doesn't provide an icon prop like the static API does"?

I'm not sure I follow which parts of the complexity are coming from this limitation.

I'm wondering what the importance is of using the icon prop, if there might be a workaround (like moving the JSX for the icon into the title?) and reducing the complexity a bit that way?

ShrimpCryptid:
> Can you explain a little bit more about what this means: "Ant's Modal component doesn't provide an icon prop like the static API does"?
> 
> I'm not sure I follow which parts of the complexity are coming from this limitation.
> 
> I'm wondering what the importance is of using the icon prop, if there might be a workaround (like moving the JSX for the icon into the title?) and reducing the complexity a bit that way?

image
^ Image of the modal design for reference.

I wanted to match the original design, where the icons lives outside the title + description text. In the `<Modal>` component, there isn't a prop where you can provide an icon, but there is one if you use Ant's static modal API, calling `modal.info({icon: some_icon_here})`.  The issue is that you have to track the instance of the static modal once it's been created, rather than having it be updated as part of the regular component rendering lifecycle.

image

You're right that it would have been easier to implement it by putting the icon into the title text instead though... argh now I'm thinking about redoing this

interim17:
At the end of the day I'll trust you if you think that's important to have it that way!

But my intuition is that if you can solve this with a little CSS/HTML in the title it will be more readable/maintainable than messing with the life cycle stuff.

ShrimpCryptid:
> At the end of the day I'll trust you if you think that's important to have it that way!
> 
> But my intuition is that if you can solve this with a little CSS/HTML in the title it will be more readable/maintainable than messing with the life cycle stuff.

image

had a shocking realization, will update this PR 😭😭😭

ShrimpCryptid:
Rerequesting @interim17! New code should be simpler 😅

interim17 (approved):
Thanks for being open to simpler means :) LGTM!

rugeli (approved):
Lgtm, too!

PR #290, opened 2024-03-28
refactor: Added styled wrapper for the static modal API

ShrimpCryptid:
Problem
=======
A comment Cameron made on the previous PR (#280) inspired this change! He had suggested I add a nice wrapper component for Ant Modals to hide the dependency on `useContext(DocumentContext)`. This worked great, but it didn't handle the use of Ant's static modal API elsewhere in the codebase.

This makes the same change but this time for Ant's static modal API! I'm using the static API in a couple places (including the upcoming #289) so this will make those PRs cleaner!

Solution
========
- Adds a drop-in replacement for `App.useApp().modal` through the new hook `useStyledModal()`.
  - The returned static methods wrap Ant's existing modal API, but inject a reference to a container for the modals if there isn't one provided.
  - This allows static methods to be called cleanly without running into the bug previously fixed in #280 where modals would render out of order.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-290/#/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fcollection.json
2. Open the Export menu, and start exporting an MP4 video. Immediately click `Cancel`, and a static modal will appear! You should be able to dismiss it and interact with it as normal.

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/components/Export.tsx
+++ b/src/components/Export.tsx
@@ -9,9 +9,9 @@ import { FlexRow } from "../styles/utils";
 import CanvasRecorder, { RecordingOptions } from "../colorizer/recorders/CanvasRecorder";
 import ImageSequenceRecorder from "../colorizer/recorders/ImageSequenceRecorder";
 import Mp4VideoRecorder, { VideoBitrate } from "../colorizer/recorders/Mp4VideoRecorder";
-import { AppThemeContext, DocumentContext } from "./AppStyle";
+import { AppThemeContext } from "./AppStyle";
 import TextButton from "./Buttons/TextButton";
-import StyledModal from "./Modals/StyledModal";
+import StyledModal, { useStyledModal } from "./Modals/StyledModal";
 import { SettingsContainer, SettingsItem } from "./SettingsContainer";
 import SpinBox from "./SpinBox";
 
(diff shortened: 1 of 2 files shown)


src/components/Export.tsx
```diff
@@ -219,12 +219,11 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
       cancelText: "Back",
       centered: true,
       icon: null,
-      getContainer: modalContainerRef || undefined,
       onOk: () => {
```
ShrimpCryptid, line 222: The drop-in replacement for the modal API worked so well that GitHub isn't showing the invocation lol. The new modal wrapper is being called on line `215` if you're curious.

frasercl (approved):
I like it!

toloudis (approved):
I think I get it, but it seems like it added a decent amount of arcane code as opposed to the "before".  Is there a runtime behavior change as a result of this?

ShrimpCryptid:
> I think I get it, but it seems like it added a decent amount of arcane code as opposed to the "before". Is there a runtime behavior change as a result of this?

No runtime behavior, makes it so we don't have to remember to get the specific container context in every additional use of the static modals API. It's currently used in two places.

PR #293, opened 2024-04-01
fix: UX updates to settings sliders

### LINKED ISSUE #211: Settings sliders UX improvements
| Brightness/Contrast | Opacity |
| --- | --- |
| image | image |

Brightness/Saturation slider show percentages in the tooltips, but opacity does not. Opacity also has a continuous integer range, while brightness/saturation are locked to 10% increments.

Also, tab navigation does not show the current values of the sliders.

## Acceptance Criteria

- Travis feedback: Change brightness/saturation to integer range w/o increments
- Change brightness to a 0-200% range
- Add back the labels to the sliders to show that their value ranges are different
- Add numeric input/spinbox to the settings sliders (currently slider tooltips are not accessible)

## Details

ShrimpCryptid:
Problem
=======
Closes #211, implementing various fixes to the slider UI.

*Estimated review size: small-med, ~20 minutes. May be slightly easier to read w/o whitespace changes*

Solution
========
This change was bigger than expected because of a UX suggestion to add input boxes to the sliders. There's a lot of similar code in the `LabeledRangeSlider` component, so rather than copy the logic I opted to update it to handle both ranges and single inputs for DRY reasons.

- Refactored `LabeledSlider` (from `LabeledRangeSlider`) to handle range type inputs and single numeric value inputs based on input props.
- Adjustments to settings sliders:
  - Settings sliders now use `LabeledSlider` under the hood and have a numeric input box.
  - Increases brightness range from 50-150% to 0-200%
  - Fixed inconsistent use of `%` labels on percentages.
  - Sliders use consistent integer stepping
  - Sliders only have one marker, at default value

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-293/#/viewer?dataset=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2F3500005828_25%2Fmanifest.json&feature=adjacent_neighbors&tab=settings&bg-key=ch0ch1&scatter-range=all
2. Go to the Viewer settings tab and adjust the values on the slider. You should see the viewer update as expected. You can also try typing into the input or giving it non-numeric/bad values.
3. Test the color ramp slider to ensure it still works as expected.

Screenshots (optional):
-----------------------
| Before | After |
| --- | --- |
| image | image |

-----------------------
1. `LabeledSlider.tsx`
4. `SettingsTab.tsx`

Thanks for contributing!


interim17 (approved):
LGTM!

src/components/LabeledSlider.tsx
```diff
@@ -1,17 +1,16 @@
-import { InputNumber, Slider } from "antd";
+import { InputNumber, Slider, SliderSingleProps } from "antd";
+import { SliderRangeProps } from "antd/es/slider";
```
meganrm, line 2: why was this import location needed? It's not exposed in the main package?

src/components/LabeledSlider.tsx
```diff
@@ -161,59 +203,86 @@ export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps):
...
-        size="small"
-        type="number"
-        style={{ width: "80px" }}
-        value={props.max}
-        onPressEnter={handleMaxInputChange}
-        onBlur={handleMaxInputChange}
-        controls={false}
-        disabled={props.disabled}
-      />
+      {props.type === "range" ? (
```
meganrm, line 273: nit: 
```suggestion
      {props.type === "range" && (
```
and then remove the null

src/components/LabeledSlider.tsx
```diff
@@ -1,17 +1,16 @@
...
+import { SliderRangeProps } from "antd/es/slider";
 import React, { ReactElement, ReactEventHandler, ReactNode, useRef } from "react";
 import styled, { css } from "styled-components";
 import { clamp } from "three/src/math/MathUtils";
 
 import { numberToStringDecimal, setMaxDecimalPrecision } from "../colorizer/utils/math_utils";
 import { excludeUndefinedValues } from "../colorizer/utils/react_utils";
 
-type LabeledRangeSliderProps = {
+type BaseLabeledSliderProps = {
```
meganrm, line 10: I'm confused why these props don't extend the antd props? and why you use the antd props down below? I'll try to read more carefully, but maybe you have a way to explain it quickly?

src/components/LabeledSlider.tsx
ShrimpCryptid: Unfortunately not... I will move the `SliderSingleProps` down with it though for consistency.

src/components/LabeledSlider.tsx
ShrimpCryptid: Looking at this code again, this is a refactor I could make! When I first wrote this component I was only interested in a subset of the Ant Slider's props, and since it was only being used in two places I didn't feel the need to provide too much flexibility. I'm also not directly passing the props to the Slider anyways, since I have to do some checks for the marks and step value size.

Is this question about L38-47 or L214?

L38-47 are still extending `BaseLabeledSliderProps`, but they define different additional parameters based on whether the `type` is set to `range` or `value`.

Below on L214, I'm extending the antd props for `sliderProps` to conditionally save either value or range-related properties! I ended up changing this to match the style of the other suggestion you'd made about making shared props for the inputs, so it should be a bit cleaner (if more verbose).

src/components/LabeledSlider.tsx
```diff
@@ -161,59 +203,86 @@ export default function LabeledRangeSlider(inputProps: LabeledRangeSliderProps):
...
+          <InputNumber
+            ref={minInput}
+            size="small"
+            style={{ width: "80px" }}
+            value={props.min}
+            onPressEnter={handleMinInputChange}
+            onBlur={handleMinInputChange}
+            controls={false}
+            disabled={props.disabled}
+          />
```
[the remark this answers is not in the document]
ShrimpCryptid: Thank you! Implemented :')

meganrm (approved):

PR #294, opened 2024-04-01
feat: Landing Page Banner

### LINKED ISSUE #263: Landing page banner
Implement landing page banner and content once designs have been finalized

ShrimpCryptid:
Problem
=======
Closes #263, adding the landing page banner.

Solution
========
- Adds the new animated landing page banner video.
  - If it can't be loaded, shows a default solid color instead (this is the fallback behavior for GitHub Pages, where the current asset hosting location doesn't work).
- Adds the feature descriptions to the landing page.
- Removes the old placeholder warning banner.

Note: I had to make a weird change because removing a `useContext` call somehow changed the CSS selector specificity for the whole app? I'm really not sure what happened but I commented the parts that were affected.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://staging.timelapse.allencell.org/
2. New banner text + feature description text should appear.

Screenshots (optional):
-----------------------
image

image


src/components/Dropdowns/AccessibleDropdown.tsx
ShrimpCryptid, line 1: Weird change here. Removing a call to `useContext` [here](https://github.com/allen-cell-animated/nucmorph-colorizer/pull/294/files#diff-35278075551402a3ec60191d0fbbea1066a18bcc00bb1533a1265fdab2cec0fcL106) caused all of my dropdown buttons to break their existing styling on the viewer page. Increasing the level of specificity by one here seems to have fixed it.

src/components/Dropdowns/AccessibleDropdown.tsx
```diff
@@ -91,18 +91,19 @@ const MainButton = styled(Button)<{ $open: boolean; $type: AccessibleDropdownPro
   ${(props) => {
     if (props.$type === "outlined") {
       return css`
-        &.ant-btn:not(:disabled) {
+        // Extra classnames to increase selector specificity; otherwise gets overridden by root styling
+        &.ant-btn.ant-btn-default:not(:disabled) {
```
meganrm, line 95: why not just add your own class? instead of multiple ant classes

ShrimpCryptid: Thank you for the suggestion! Refactored this :)

src/routes/LandingPage.tsx
```diff
@@ -12,27 +12,106 @@ import { DatasetEntry, LocationState, ProjectEntry } from "../types";
...
+  margin: 0;
+`;
+
+const BannerTextContainer = styled(FlexColumn)`
+  --padding-x: 30px;
+  padding: var(--padding-x);
+  max-width: calc(1060px - 2 * var(--padding-x));
+
+  --total-padding-x: calc(2 * var(--padding-x) + 2 * var(--container-padding-x));
+  width: calc(90vw - var(--total-padding-x));
```
frasercl, line 34: Just curious: why this width calculation? It looks like the effect of this is that, as the user resizes the screen down past the max width of the banner text box, the x margin between the edge of the screen and the edge of the text box gradually shrinks - is that intended?

frasercl (approved):
Looks good! One non-blocking behavior question.

src/routes/LandingPage.tsx
ShrimpCryptid: Yup! This is mostly based off a similar calculation I'm doing for the rest of the page content.

PR #306, opened 2024-04-03
fix: Fixed bug where dropdowns were using incorrect style

ShrimpCryptid:
Problem
=======
Follow-up to #294! It turns out that I did not fix the dropdown styling, and that the styled-components override and the ant override both had the same level of specificity. (It only seemed fixed because hot-reload was causing styled-components to create a new CSS class that then took precedence over the ant override.)

*Estimated review size: tiny, 2 minutes*

Solution
========
- Increased the specificity of the styled-components override by one to fix a bug where the dropdowns were using the incorrect style.
- Fixed an incorrectly used CSS variable.

## Type of change

Steps to Verify:
----------------
1. Open the preview link. The outlined dropdowns should be using a grey color instead of a purple: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-306/#/viewer
2. Bugged build for reference: https://allen-cell-animated.github.io/nucmorph-colorizer/main/

### DIFF
--- a/src/components/Dropdowns/AccessibleDropdown.tsx
+++ b/src/components/Dropdowns/AccessibleDropdown.tsx
@@ -89,7 +89,6 @@ const MainButton = styled(Button)<{ $open: boolean }>`
   transition: all 0.2s cubic-bezier(0.645, 0.045, 0.355, 1), width 0s;
 
   // Override Ant styling for the outlined button style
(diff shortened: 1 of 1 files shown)


meganrm (approved):
Normally I would suggest template literals over string concatenation, but in this case it's a real tossup of what's easier to read. so no strong opinion here.

Issue #329, opened 2024-04-19
`NaN` in feature data is shown as `Infinity` once loaded. (also, UI will probably crash if `Infinity` is given as a data value.)


ShrimpCryptid:
## Description

`NaN` values in feature data are replaced with `Infinity` on the UI.

![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/c88707f1-b39a-47a1-b9f6-244e3f7e99d5)

![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/ac322090-69d2-4628-8da6-031917bdd82c)

On closer inspection, it looks like it's caused by these lines in `JsonArrayLoader.ts`:
```
const nanToNull = (json: string): string => json.replace(/NaN/g, "null");

export default class JsonArrayLoader implements IArrayLoader {
  async load(url: string): Promise<JsonArraySource> {
    const response = await fetch(url);
    const text = await response.text();
    const { data, min, max }: FeatureDataJson = JSON.parse(nanToNull(text));
    console.log(url, data);
    return new JsonArraySource(data, min, max);
  }
}
```

and this logic up above in the constructor:
```
      this.array = array.map((val) => (val === null ? Infinity : val));
```

`JSON.parse` doesn't parse `NaN` values, so we are currently converting `NaN` to `null` and then `null` to `Infinity`. Note that this WILL NOT handle positive or negative `Infinity` in our data, which will cause this to crash and the dataset load to fail.

A few possible fixes:
- EASY: Update the JSON parsing to also convert `Infinity` or other values to `null`, and change `null` to be parsed as `NaN` by default.
- HARDER: Use a more robust JSON parsing library, like https://json5.org/, so we can directly represent `Infinity`/`NaN`/etc.

I'd also like to add unit tests for both cases!

## Expected Behavior

- Feature data values should match what is shown on the UI.

## Reproduction

https://dev-aics-dtp-001.int.allencell.org/nucmorph-colorizer/dist/index.html#/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fnucmorph-apr2024%2Fcollection.json&dataset=Mama+Bear&feature=time

## Environment

_Any additional information about your environment. E.g. OS version, python version, browser etc._

ShrimpCryptid:
Extra thoughts from a conversation w/ @toloudis:
[...]

PR #330, opened 2024-04-19
fix: Fix crash from `ax.dtick` NaN error

### LINKED ISSUE #291: Bad axes crash
## Description

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/0118ad95-0c60-47fe-8821-c6a4f4e05172

UI repeatedly loaded, hit an error screen, and returned while I had the reprocessed Nucmorph dataset open.

I believe this was caused by an issue with the scatterplot axis, but I have no way to be sure.

Error messages:
> ax.dtick error: NaN
> THREE.WebGLRenderer: Context Lost.
> React Router caught the following error during render `ax.dtick error: NaN`

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
*Estimated size: small, 10-15 minutes*

Closes #291, a crash with the error message `ax.dtick error: NaN` on certain datasets.

I think what was happening was that Plotly can handle _some_ `Infinity` or `NaN` values in input data, but fails when calculating the range of axis values to show if _all_ the values are `Infinity` or `NaN`. 

This happens in some datasets on certain frame + feature combinations, such as in the updated nucmorph datasets. The **time\*** feature fails on the first frame, when values are undefined for all the cells currently in frame.

(\*representing the age of a cell since mitosis)

I also opened up a related issue here: [#329, NaN in feature data is shown as Infinity once loaded.](https://github.com/allen-cell-animated/nucmorph-colorizer/issues/329)

Solution
========
- Unrepresentable `NaN` and `Infinity` values are now removed from scatter plot data before rendering.
- Adds a warning banner for when the plot rendering fails for any reason, and clears the plot.

## Type of change

Steps to Verify:
----------------
(Note: these datasets are pretty big, so they may take up to 10 seconds to load.)
1. Open the bugged build: https://dev-aics-dtp-001.int.allencell.org/nucmorph-colorizer/dist/index.html#/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fnucmorph-apr2024%2Fcollection.json&dataset=Mama+Bear&feature=time&tab=scatter_plot&scatter-range=all&scatter-x=time&scatter-y=volume
2. Note that all cells on the current frame have a value of `Infinity`.
3. Go to the scatter plot tab and change the range from **All time** to **Current frame**. The UI will crash.
4. Now open the preview link and repeat the above instructions. No error should occur. https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-330/#/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fnucmorph-apr2024%2Fcollection.json&dataset=Mama+Bear&feature=time&tab=scatter_plot&scatter-range=all&scatter-x=time&scatter-y=volume

## Screenshots:
**Video showing off the error banner (🔊):**

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/b8663610-bf0e-4472-bb94-d4fcf94ce9a5

image


toloudis (approved):
sanitize func looks good!  Still wonder about putting a try/catch around Plotly.react as an extra insurance layer.

ShrimpCryptid:
@toloudis Re-requesting because of the new error banner handler!

meganrm:
General question: is the data on this app always going through your data conversion repo? Because if so, this seems like something you'd want to fix there instead of having to check it all on the front end

src/components/Tabs/ScatterPlotTab.tsx
```diff
@@ -716,15 +740,31 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+        setIsRendering(false);
+        lastRenderedState.current = {
+          xAxisFeatureKey,
+          yAxisFeatureKey,
+          rangeType,
+          ...props,
+        };
+      });
+    } catch (error) {
+      console.error(error);
```
meganrm, line 754: could you move your sanitizeNumericDataArrays into this block so you're only cleaning the data if it fails? and then if it fails a second time show the alert?

ShrimpCryptid:
> General question: is the data on this app always going through your data conversion repo? Because if so, this seems like something you'd want to fix there instead of having to check it all on the front end

We still need to have a conversation with scientists about what the expected behavior is, but I think having NaN values is valid for certain features. (ex: age of a cell since mitosis is NaN for cells that are present on the first frame, since the time since mitosis is unknown for them.) The issue here is that Plotly can't render plots with *only* NaN/Infinity values, which can happen with certain range filters (like only showing the current frame, which happens to be a frame with only NaN values).

src/components/Tabs/ScatterPlotTab.tsx
ShrimpCryptid: The structure of how the plot is rendered might make that a bit difficult? Currently the data is being filtered by range at the start of `renderPlot`, and there are additional methods that take in the (sanitized) data and perform additional operations on it (for example, highlighting the currently selected object or track). That means that we'd have to perform all of those calculations twice, and I'd have to restructure the method to do the try/catch checking.

Do you think we're losing something from sanitizing the data preemptively?

toloudis (approved):

src/components/Tabs/ScatterPlotTab.tsx
meganrm: I was just trying to avoid mapping over the data more than is necessary,  I'm assuming these are pretty large arrays

meganrm (approved):

src/components/Tabs/ScatterPlotTab.tsx
ShrimpCryptid: That makes sense. I'll add a note to revisit this if it causes performance issues.

PR #339, opened 2024-05-07
feat: Implemented analytics events 

### LINKED ISSUE #326: GA events for interactions
## Use Case

Add GA events for key interactions like dataset load, collection load, crashes/error messages, and export menu interactions.

## Acceptance Criteria

| Event | Description |
| -- | -- |
| Dataset Load | Report when a dataset is loaded and its schema version |
| Collection loaded | Report when a collection is loaded, and schema version |
| Crashes/error messages | Report when the user hits the error page |
| Features | Track when a user switches features as a proxy for engagement |
| Export button | Reporting user using Export menu item, also hit "Export" button and complete the export (event) |
| Color map selection | Users are clicking the color map button and what maps they select |

## Details

- Track number of features * number of objects?

ShrimpCryptid:
Problem
=======
Closes #326, implementing GA tracking for interactions!

Solution
========
Much of the setup takes place in Google Tag Manager and not in the UI itself. 

- Adds GA events for the following actions:
  - Loading datasets (reports schema version, number of objects, number of features, number of frames)
  - Loading collections (reports schema version)
  - Switching features (reports schema version, feature range, and feature type)
  - Export completed (reports export type)
  - Error screen (reports error encountered)
- Adds IDs to some elements to support tag manager event detection. 
- Adds missing accessibility labels to some buttons

## Type of change

Screenshots (optional):
-----------------------

**Video talk-through (+ demoing events) (🔊):**

https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/ded27b4b-4f82-4132-a717-42374c6190af

-----------------------
1. `analytics_utils.ts`


src/colorizer/Collection.ts
```diff
@@ -252,6 +253,10 @@ export default class Collection {
       collectionData.set(entry.name, entry);
     }
 
+    triggerAnalyticsEvent(AnalyticsEvent.COLLECTION_LOAD, {
+      collectionWriterVersion: collection.metadata?.writerVersion || "N/A",
```
ShrimpCryptid, line 257: Could also use "v0.0.0" as a placeholder but could also be a custom Collection JSON that a user has written without using the writer scripts.

src/components/Dropdowns/ColorRampDropdown.tsx
```diff
@@ -17,6 +17,9 @@ import IconButton from "../IconButton";
 
 import styles from "./ColorRampDropdown.module.css";
 
+const COLOR_RAMP_BUTTON_CLASS = "color-ramp-button";
+const COLOR_PALETTE_BUTTON_CLASS = "color-palette-button";
+
```
ShrimpCryptid, line 22: Adds classes to the ramp and palette buttons so they can be used to filter click events in GA.

Other changes:
- Added missing aria-label tags
- Add key of each color ramp/palette as the ID of the button elements

src/colorizer/Dataset.ts
```diff
@@ -418,6 +419,14 @@ export default class Dataset {
       throw new Error("No features found in dataset. Is the dataset manifest file valid?");
     }
 
+    // Analytics reporting
+    triggerAnalyticsEvent(AnalyticsEvent.DATASET_LOAD, {
+      datasetWriterVersion: this.metadata.writerVersion || "N/A",
+      datasetTotalObjects: this.numObjects,
+      datasetFeatureCount: this.features.size,
+      datasetFrameCount: this.numberOfFrames,
```
toloudis, line 427: why not put manifest url here also?  wouldn't that make it easier?

src/colorizer/utils/analytics_utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+export const enum AnalyticsEvent {
+  COLLECTION_LOAD = "collection_load",
+  DATASET_LOAD = "dataset_load",
+  EXPORT_COMPLETE = "export_complete",
+  FEATURE_SELECTED = "feature_selected",
+  ROUTE_ERROR = "route_error",
+}
```
toloudis, line 22: do these have to be defined at the google analytics back end?  Or does it just take whatever we send?

src/colorizer/utils/analytics_utils.ts
toloudis, line 1: consider renaming to analytics.ts. it's already in a utils subdirectory

src/routes/ErrorPage.tsx
```diff
@@ -16,10 +17,13 @@ export default function ErrorPage(): ReactElement {
 
   if (isErrorResponse(error)) {
     errorMessage = error.status + " " + error.statusText;
+    triggerAnalyticsEvent(AnalyticsEvent.ROUTE_ERROR, { errorMessage: error.statusText, errorStatus: error.status });
   } else if (error instanceof Error) {
     errorMessage = error.message;
+    triggerAnalyticsEvent(AnalyticsEvent.ROUTE_ERROR, { errorMessage: error.message });
   } else {
     errorMessage = "Unknown error";
+    triggerAnalyticsEvent(AnalyticsEvent.ROUTE_ERROR, { errorMessage: "Unknown error" });
```
toloudis, line 26: if someone mistakenly copy/pastes a bad data set url will one of these analytics errors happen? Any concern about a disproportionate amount of these coming through?

toloudis (approved):

src/routes/ErrorPage.tsx
ShrimpCryptid: Routing errors are catastrophic errors that usually indicate a bug in the viewer code rather than an issue with the dataset. Dataset errors *should* be shown on the UI via a popup (and not trigger the `ROUTE_ERROR` event), though there's almost certainly some edge cases we haven't caught yet from weird/nonsensical dataset configurations.

src/colorizer/Dataset.ts
ShrimpCryptid: We absolutely could! I just wasn't sure if there were any concerns about data privacy in doing so (though we do collect URL params/path so technically we already have this anyways... :X)

src/colorizer/utils/analytics_utils.ts
ShrimpCryptid: Yes! Well. 
From what I understand, GA will still log the events, but they won't trigger the corresponding tag that actually captures and then does interesting stuff with that information (e.g. set it up for reports/exploration). So yes, all of these have counterparts in GA Tag Manager that I've set up.

src/colorizer/utils/analytics_utils.ts
ShrimpCryptid: Everything else in the folder has the `_utils` suffix so this makes it consistent with the standard formatting 😅

toloudis: Okay but this isn't a utils file.  It's an analytics module

PR #342, opened 2024-05-13
fix: Missing first frame in Dataset no longer throws an error

### LINKED ISSUE #214: Failed dataset load gives contextless `Error` message
## Description

image

This (unhelpful) error message occurred when the Dataset failed to load on one of Goutham's datasets, where the `Dataset.open()` operation failed. It was specifically caused by the first frame of the dataset being missing (see related issue #143).

I've replicated it with this test dataset here: https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/baby_bear/

## Expected Behavior

The error message should be more descriptive.

Ex: `Error: Could not open dataset due to missing files. Please check that the dataset is not missing frames or JSON data.`

## Reproduction

1. Open the colorizer: https://dev-aics-dtp-001.int.allencell.org/nucmorph-colorizer/dist/index.html
2. Try loading this test dataset: https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/baby_bear/

ShrimpCryptid:
Problem
=======
*Estimated review size: Small, 5-10 minutes*

Closes #214, a bug where trying to load a dataset that was missing its first frame would result in an "Error" message in the load modal.

image

Solution
========
- Changes `Dataset.open` to use `Promise.allSettled` instead of `Promise.all` when loading dataset files, which allows it to handle error cases gracefully.
  - Datasets missing their first frame are still able to be loaded.
- Updates `ColorizeCanvas` to check sizing every time a frame is loaded in case of missing frame dimensions (this can happen when the first frame is missing on a Dataset).

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-342/#/viewer?dataset=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Fbaby_bear%2Fmanifest.json
2. The first frame should show as a blank screen with "Missing image data" shown as an onscreen overlay.
3. Advancing to the next frame should show a normal frame.

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -356,13 +356,19 @@ export default class Dataset {
   }
 
   /** Loads the dataset manifest and features. */
-  public async open(manifestLoader = this.fetchJson): Promise<void> {
+  public async open(
+    options: Partial<{ manifestLoader: (url: string) => Promise<AnyManifestFile> }> = {}
+  ): Promise<void> {
     if (this.hasOpened) {
       return;
     }
+
+    const defaultOptions = { manifestLoader: this.fetchJson };
+    const fullOptions = { ...defaultOptions, ...options };
+
     this.hasOpened = true;
 
-    const manifest = updateManifestVersion(await manifestLoader(this.manifestUrl));
+    const manifest = updateManifestVersion(await fullOptions.manifestLoader(this.manifestUrl));
 
     this.frameFiles = manifest.frames;
     this.outlierFile = manifest.outliers;
@@ -392,7 +398,7 @@ export default class Dataset {
       this.loadFeature(data)
     );
 
-    const result = await Promise.all([
+    const result = await Promise.allSettled([
       this.loadToBuffer(FeatureDataType.U8, this.outlierFile),
       this.loadToBuffer(FeatureDataType.U32, this.tracksFile),
       this.loadToBuffer(FeatureDataType.U32, this.timesFile),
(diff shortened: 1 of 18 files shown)


src/colorizer/Dataset.ts
```diff
@@ -395,7 +394,7 @@ export default class Dataset {
       this.loadFeature(data)
     );
 
-    const result = await Promise.all([
```
ShrimpCryptid, line 398: Promise.all fails if any of the promises fail, while Promise.allSettled waits for all promises to succeed OR fail and then return the resulting statuses.

ShrimpCryptid: Not shown in the edits, but this allows line 404, `this.loadFrame(0)` to be fine if it throws an error.

src/colorizer/Dataset.ts
```diff
@@ -406,15 +405,42 @@ export default class Dataset {
...
+    if (centroids.status === "rejected") {
+      console.warn("Failed to load centroids: ", centroids.reason);
+    } else {
+      this.centroids = centroids.value;
+    }
+    if (bounds.status === "rejected") {
+      console.warn("Failed to load bounds: ", bounds.reason);
+    } else {
+      this.bounds = bounds.value;
+    }
```
meganrm, line 433:
```suggestion
    const successData = [this.outliners, this.trackIds, this.times, this centroids, this.bounds]
    results.forEach((result, index) => {
        if (result.status) == "rejected" {
            console.warn(`failed to load: ${result.reason}`
        } else if (index < successResults.length) {
            successResult[index] = result.value
         } else {
             const [key, data] = result.value;
             this.features.set(key, data)
         }
    
    }
```

meganrm: I wouldn't commit this, it was just easier to write out as code. It just seems like if you have an array as your return it's nice to continue to treat it as an array and reduce repeated code. If it's really important that you also have the name of the data in the warning message you could also make an array of names somewhere.

src/colorizer/Dataset.ts
```diff
@@ -406,15 +405,42 @@ export default class Dataset {
...
 
     // Keep original sorting order of features by inserting in promise order.
-    featureResults.forEach(([key, data]) => {
-      this.features.set(key, data);
+    featureResults.forEach((result) => {
+      if (result.status === "rejected") {
+        console.warn("Failed to load feature: ", result.reason);
+        return;
```
meganrm, line 439: you don't need this return here since you have the else

src/colorizer/Dataset.ts
ShrimpCryptid: Gotcha. Some of these types are required, so I'll need to add an extra step for whether we should throw an error if that data is missing. This makes sense though.

src/colorizer/Dataset.ts
```diff
@@ -406,15 +405,42 @@ export default class Dataset {
...
+    } else {
+      this.outliers = outliers.value;
+    }
+    if (tracks.status === "rejected") {
+      console.warn("Failed to load tracks: ", tracks.reason);
+    } else {
+      this.trackIds = tracks.value;
+    }
+    if (times.status === "rejected") {
+      throw new Error("Time data could not be loaded. Is the dataset manifest file valid?");
```
rugeli, line 420: Just curious, why didn't we include `reason` in this error message for `times`?

ShrimpCryptid: Ah, including it as a console warning! The error that's shown here would be directly shown on the UI.

src/colorizer/Dataset.ts
ShrimpCryptid: Uhh, looking at this further I'm not sure this would actually work. This wouldn't play nicely with typing (different files are different types, including unsigned 8/16/32-bit integer arrays) and I don't think there's actually a way to keep references to data fields as an array and assign them like you show here unless I'm storing, say, a setter method.

I do get wanting to remove the repeated code here, so I tried adding a handler that will retrieve the promise value and throws a warning if it doesn't exist.

rugeli (approved):
the last commit for refactoring the error handling looks much cleaner!

meganrm (approved):

PR #344, opened 2024-05-16
feat: Remove Nucmorph as the default dataset, add banner popup for missing dataset

### LINKED ISSUE #228: Remove nucmorph as the default dataset
- Either have a blank screen or have a preloaded example dataset
  - Confirm behavior with UX

ShrimpCryptid:
Problem
=======
Closes #228, removing nucmorph as the default dataset!

*Estimated size: small, 5-10 minutes*

Solution
========
- Removes the default nucmorph dataset that was used when datasets were missing.
- Adds an info banner for when only the empty viewer page is loaded.
- Fixes a bug where the banners would be immediately cleared on page load, leading to no banners being shown.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/nucmorph-colorizer/pr-preview/pr-344/#/viewer
2. You should see an empty page with the info banner at the top:

image

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -32,7 +32,7 @@ import { getColorMap, getInRangeLUT, thresholdMatchFinder, validateThresholds } 
 import { numberToStringDecimal } from "./colorizer/utils/math_utils";
 import { useConstructor, useDebounce, useRecentCollections } from "./colorizer/utils/react_utils";
 import * as urlUtils from "./colorizer/utils/url_utils";
-import { DEFAULT_COLLECTION_PATH, DEFAULT_PLAYBACK_FPS } from "./constants";
+import { DEFAULT_PLAYBACK_FPS } from "./constants";
 import { FlexRowAlignCenter } from "./styles/utils";
 import { LocationState } from "./types";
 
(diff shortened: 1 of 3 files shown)


src/Viewer.tsx
```diff
@@ -149,7 +149,7 @@ function Viewer(): ReactElement {
   };
   const [notificationApi, notificationContextHolder] = notification.useNotification(notificationConfig);
 
-  const { bannerElement, showAlert } = useAlertBanner([dataset]);
+  const { bannerElement, showAlert, clearBanners } = useAlertBanner();
```
ShrimpCryptid, line 152: Manually clear the banners instead of clearing whenever `dataset` changes. This caused a bug because the `useEffect` that listens for changes to `dataset` would fire immediately after the first render, causing the banners emitted by any error handlers to be cleared before they could be shown to the user.

interim17 (approved):
LGTM!

rugeli (approved):
I was wondering if the UI immediately reflects the cleared state without any flicker or delay when `clearBanners` is called manually. Other than that, this fix looks reasonable to me!

ShrimpCryptid:
> I was wondering if the UI immediately reflects the cleared state without any flicker or delay when `clearBanners` is called manually. Other than that, this fix looks reasonable to me!

It should update in the next state update, so within a fraction of a second or less depending on whether React is chugging.

PR #377, opened 2024-06-07
fix: Bad feature values are now shown as NaN instead of Infinity

### LINKED ISSUE #329: `NaN` in feature data is shown as `Infinity` once loaded. (also, UI will probably crash if `Infinity` is given as a data value.)

ShrimpCryptid:
Problem
=======
Closes #329, a bug where `NaN` values are shown as `Infinity`.

Solution
========
- Added a comment explaining why `Infinity` values are stored internally.
- Added a check that replaces `Infinity` values with `NaN` on the hover tooltip.
- Deleted unused shader.
- Minor formatting changes to shader code.

## Type of change

Steps to Verify:
----------------
1. Open the preview link and hover over the cells. Greyed-out cells should appear as `NaN` instead of `Infinity`: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-377/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fnucmorph-apr2024%2Fcollection.json&dataset=Mama+Bear&feature=time&t=396
2. Compare with main build: https://allen-cell-animated.github.io/timelapse-colorizer/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Fnucmorph-apr2024%2Fcollection.json&dataset=Mama+Bear&feature=time&t=396
3. 
image


src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -1,4 +1,5 @@
 precision highp usampler2D;
+precision highp sampler2D;
```
toloudis, line 2: I suggest removing this until we have a PR where it actually matters

src/colorizer/shaders/colorize_R32.frag
```diff
@@ -60,7 +60,7 @@ void main() {
 
   // Data buffer starts at 0, non-background segmentation IDs start at 1
   float featureVal = getFeatureVal(id - 1);
-  uint outlierVal = getOutlierVal(id - 1)
+  uint outlierVal = getOutlierVal(id - 1);
```
toloudis, line 63: ??? how was this shader even compiling without the semicolon

toloudis (approved):
The main changes look good! But I bet you can remove the modifications from at least 1 and maybe 2 of the files

src/colorizer/shaders/colorize_R32.frag
toloudis: If this is a totally unused shader then maybe we should just delete the whole file

ShrimpCryptid: Currently unused, yeah. I'll make that change.

interim17 (approved):

PR #388, opened 2024-06-20
fix: URL no longer updates while dataset is still unloaded

### LINKED ISSUE #387: URL changes when dataset fails to load and breaks refresh
## Description

When a dataset fails to load, the URL changes to remove the dataset. This causes a refresh of the page to end up in a different state (with a, "there is no dataset loaded") message instead of a load failure.

## Expected Behavior

The URL should not change unless the dataset was loaded successfully.

ShrimpCryptid:
Problem
=======
Closes #387, "URL changes when dataset fails to load and breaks refresh".

Solution
========
- Prevents URL from updating while dataset is unloaded or fails to load.
- Also removes old TODOs.

## Type of change

Steps to Verify:
----------------
1. Open the preview link:
2. 

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
3. other important file

Thanks for contributing!

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -248,10 +248,12 @@ function Viewer(): ReactElement {
     scatterPlotConfig,
   ]);
 
-  // Update url whenever the viewer settings change
-  // (but not while playing/recording for performance reasons)
+  // Update url whenever the viewer settings change, with a few exceptions:
+  // - The URL should not change while playing/recording for performance reasons.
+  // - The URL should not change if the dataset hasn't loaded yet, or if it failed to load.
+  //   that way, users can refresh the page to try again.
   useEffect(() => {
-    if (!timeControls.isPlaying() && !isRecording) {
+    if (!timeControls.isPlaying() && !isRecording && isInitialDatasetLoaded) {
       setSearchParams(getUrlParams(), { replace: true });
     }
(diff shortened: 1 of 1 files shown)


interim17 (approved):

src/Viewer.tsx
```diff
@@ -248,10 +248,12 @@ function Viewer(): ReactElement {
...
+  // Update url whenever the viewer settings change, with a few exceptions:
+  // - The URL should not change while playing/recording for performance reasons.
+  // - The URL should not change if the dataset hasn't loaded yet, or if it failed to load.
+  //   that way, users can refresh the page to try again.
   useEffect(() => {
-    if (!timeControls.isPlaying() && !isRecording) {
+    if (!timeControls.isPlaying() && !isRecording && isInitialDatasetLoaded) {
       setSearchParams(getUrlParams(), { replace: true });
     }
   }, [timeControls.isPlaying(), isRecording, getUrlParams]);
```
frasercl, line 259: should `isInitialDatasetLoaded` be a dependency?

frasercl (approved):

src/Viewer.tsx
ShrimpCryptid: Yes, good catch!

Issue #20, opened 2023-06-02
Load features asynchronously


toloudis:
## Use Case

If I have a dataset with a large number of features, I may not want to consume extra resources by loading them all initially.

## Acceptance Criteria

Only load feature data on-demand.

## Details

Could discard old feature data (or optionally keep around a very limited number of them in a local cache)

ShrimpCryptid:
Starting to see big slowdowns on the nucmorph dataset. May experimentally implement this.

ShrimpCryptid:
Should we close this now that feature load is multithreaded?

toloudis:
I agree we can close it and create a new ticket if we have concerns later.  Binary+worker loading was a big win

ShrimpCryptid:
Closed by #420 and #409, which solve the underlying performance concerns.

PR #393, opened 2024-06-24
refactor: Make FrameCache generic

ShrimpCryptid:
Problem
=======
Part 1 of 3 implementing #20, performance improvements when handling large numbers of features. This change changes the `FrameCache` previously implemented by Cameron into a generic cache that can handle other types of data. This will eventually be used to store feature data and renamed to `DataCache`.

*Estimated review size: small, 15 minutes.*

Solution
========
- Refactored `FrameCache` to allow generic value types when caching.
- Wrote unit tests for `FrameCache`.

## Type of change

Steps to Verify:
----------------
1. Open the preview link in a private/incognito window so it isn't using cached data:
2. Let the video play up until about frame 60, then reset it to 0. It should play faster on the second round, indicating that the cache is still working.

### DIFF
--- a/src/colorizer/FrameCache.ts
+++ b/src/colorizer/FrameCache.ts
@@ -14,7 +14,7 @@ type CacheEntry<E extends DisposableValue> = {
  * Generic LRU cache for data, intended for Texture or other GPU resources.
  * Calls `dispose` on eviction to keep GPU memory under control.
  */
-export default class DataCache<E extends DisposableValue> {
+export default class FrameCache<E extends DisposableValue> {
   private data: Map<string, CacheEntry<E>>;
   private first: CacheEntry<E> | null;
   private last: CacheEntry<E> | null;
--- a/src/colorizer/FrameCache.ts
+++ b/src/colorizer/FrameCache.ts
@@ -96,10 +96,6 @@ export default class DataCache<E extends DisposableValue> {
     key = key.toString();
     const currentEntry = this.data.get(key);
 
-    if (size > this.maxSize) {
-      throw new Error(`Attempted to insert a value of size ${size} into a cache with maxSize ${this.maxSize}`);
-    }
-
     if (currentEntry !== null && currentEntry !== undefined) {
       // NOTE: This assumes that the value's size HAS NOT CHANGED. If it has,
       // the size of the cache will be incorrect and the old value will not be disposed of.
--- a/src/colorizer/FrameCache.ts
+++ b/src/colorizer/FrameCache.ts
@@ -14,7 +14,7 @@ type CacheEntry<E extends DisposableValue> = {
  * Generic LRU cache for data, intended for Texture or other GPU resources.
  * Calls `dispose` on eviction to keep GPU memory under control.
  */
-export default class FrameCache<E extends DisposableValue> {
+export default class DataCache<E extends DisposableValue> {
   private data: Map<string, CacheEntry<E>>;
   private first: CacheEntry<E> | null;
   private last: CacheEntry<E> | null;
@@ -39,7 +39,9 @@ export default class FrameCache<E extends DisposableValue> {
       this.last.next.prev = null;
     }
 
-    this.last.value.dispose && this.last.value.dispose();
+    if (this.last.value.dispose) {
+      this.last.value.dispose();
+    }
     this.data.delete(this.last.key);
     this.currentSize -= this.last.size;
     this.last = this.last.next;
@@ -82,7 +84,8 @@ export default class FrameCache<E extends DisposableValue> {
    * If the cache is full, the least recently used entry is evicted and its `dispose()` function is
    * called if it exists.
    *
-   * Throws an error if the size of the value is greater than the cache's maxSize.
+   * Throws an error if the `size` is greater than the cache's `maxSize`, as defined in the
+   * constructor.
    */
   public insert(key: string | number, value: E, size: number = 1): void {
     key = key.toString();
--- a/src/colorizer/FrameCache.ts
+++ b/src/colorizer/FrameCache.ts
@@ -21,6 +21,11 @@ export default class DataCache<E extends DisposableValue> {
   private currentSize: number;
   private maxSize: number;
 
+  /**
+   * Creates a new data cache with a maximum size.
+   * @param maxSize Arbitrary max size. Can be in number of entries or bytes; if bytes,
+   * entries should have a size defined in bytes when inserted.
+   */
   constructor(maxSize: number = 30) {
     this.data = new Map();
     this.first = null;
--- a/src/colorizer/FrameCache.ts
+++ b/src/colorizer/FrameCache.ts
@@ -2,7 +2,7 @@ type DisposableValue = {
   dispose?: () => void;
 };
 
-type CacheEntry<E extends DisposableValue> = {
+type CacheEntry<E> = {
   key: string;
   value: E;
   size: number;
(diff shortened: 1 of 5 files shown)


src/colorizer/FrameCache.ts
```diff
@@ -1,48 +1,52 @@
...
-  private data: CacheEntryNullable[];
-  private first: CacheEntryNullable;
-  private last: CacheEntryNullable;
-  private numItems: number;
+/**
+ * Generic LRU cache for data, intended for Texture or other GPU resources.
+ * Calls `dispose` on eviction to keep GPU memory under control.
+ */
+export default class FrameCache<E extends DisposableValue> {
+  private data: Map<string, CacheEntry<E>>;
```
ShrimpCryptid, line 18: Now uses string keys instead of integers which index into an array, which allows more flexible data to be stored (e.g. features)

src/colorizer/FrameCache.ts
```diff
@@ -69,78 +73,65 @@ export default class FrameCache {
...
-    if (index >= this.data.length) {
-      return;
+  /**
+   * Inserts a value into the cache with the key. If the key already exists, the value is replaced.
+   * If the cache is full, the least recently used entry is evicted and its `dispose()` function is
+   * called if it exists.
+   *
+   * Throws an error if the size of the value is greater than the cache's maxSize.
+   */
+  public insert(key: string | number, value: E, size: number = 1): void {
```
ShrimpCryptid, line 87: Variable size is useful if we want to define the cache's capacity in terms of bytes in the future.

src/colorizer/FrameCache.ts
ascibisz, line 1: I am bad with naming best practices, but should this file be renamed to `DataCache.ts` as well, since it now only contains the generic `DataCache` class?

ascibisz (approved):

src/colorizer/FrameCache.ts
ShrimpCryptid: Yes, it should be! I left it as `FrameCache.ts` and renamed it in a subsequent PR so GitHub still picks up on the diff while reviewing :')

src/colorizer/FrameCache.ts
```diff
@@ -1,48 +1,54 @@
...
 
-/** LRU cache for textures. Calls `dispose` on eviction to keep GPU memory under control. */
-export default class FrameCache {
-  private data: CacheEntryNullable[];
-  private first: CacheEntryNullable;
-  private last: CacheEntryNullable;
-  private numItems: number;
+/**
+ * Generic LRU cache for data, intended for Texture or other GPU resources.
+ * Calls `dispose` on eviction to keep GPU memory under control.
```
toloudis, line 15: curious from reading this comment: what do you do if the gpu needs a resource but it's been evicted?  

I guess you're going on the assumption that the most recently used set of assets required by the gpu for a given rendered frame will always fit in cache.

src/colorizer/FrameCache.ts
```diff
@@ -1,48 +1,54 @@
...
+export default class DataCache<E extends DisposableValue> {
+  private data: Map<string, CacheEntry<E>>;
+  private first: CacheEntry<E> | null;
+  private last: CacheEntry<E> | null;
+  private currentSize: number;
   private maxSize: number;
 
-  constructor(length: number, maxSize: number = 30) {
-    this.data = new Array(length).fill(null);
+  constructor(maxSize: number = 30) {
```
toloudis, line 24: shouldn't this maxSize be a byte count?

src/colorizer/FrameCache.ts
toloudis: interesting how you did this.  I like it but I would suggest that sizes=1 are not very useful when they can range widely in how much memory they consume.

src/colorizer/FrameCache.ts
```diff
@@ -69,78 +75,66 @@ export default class FrameCache {
...
-      if (this.numItems >= this.maxSize) {
+      this.currentSize += size;
+      while (this.currentSize > this.maxSize && this.last) {
         this.evictLast();
-      } else {
-        if (!this.last) {
-          this.last = newEntry;
-        }
-        this.numItems++;
+      }
```
toloudis, line 111: sanity check: does evictLast update currentSize?

tests/DataCache.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  class DisposableString {
+    public value: string;
+    public hasBeenDisposed: boolean = false;
+
+    constructor(value: string) {
+      this.value = value;
+    }
+
+    dispose(): void {
+      this.hasBeenDisposed = true;
```
toloudis, line 17: it would also be interesting to test caching a type that doesn't have a dispose method, like a simple Array or ArrayBuffer.

src/colorizer/FrameCache.ts
```diff
@@ -69,78 +75,66 @@ export default class FrameCache {
...
-      return;
+  /**
+   * Inserts a value into the cache with the key. If the key already exists, the value is replaced.
+   * If the cache is full, the least recently used entry is evicted and its `dispose()` function is
+   * called if it exists.
+   *
+   * Throws an error if the `size` is greater than the cache's `maxSize`, as defined in the
+   * constructor.
+   */
+  public insert(key: string | number, value: E, size: number = 1): void {
```
toloudis, line 90: it could also be interesting if the `E` type had an optional function it could call to return its size, rather than explicitly passing it in separately.  (Both scenarios are potentially interesting. I'm thinking about things like Arrays or ArrayBuffers that know their size)

src/colorizer/FrameCache.ts
ShrimpCryptid: Agreed! The `size=1` matches the previous behavior for frames, where the max capacity was defined in terms of the number of frames in the cache, but we should revisit this in the future. I'll add a TODO about it.

src/colorizer/FrameCache.ts
ShrimpCryptid: Okay, made a change so the frame cache now has a max limit of 64 MB, based on raw image size. With the default nucmorph dataset, this is about 16 frames. The browser is also doing a lot of caching though, so I'm not confident whether this will have a measurable impact.

src/colorizer/FrameCache.ts
ShrimpCryptid: It does!

tests/DataCache.test.ts
ShrimpCryptid: Added!

src/colorizer/FrameCache.ts
ShrimpCryptid: Also, I've updated some behavior so frames larger than the cache size can still be cached; without this, certain datasets with very large frames (like the endothelial flow datasets, which have a raw frame size of 100+ MB) exceed the cache size.

src/colorizer/FrameCache.ts
ShrimpCryptid: Ended up skipping this for now, because I think it's more general if the size is an optional property on the insertion method than on the objects themselves. This is definitely making the assumption that entries do not change size.

tests/DataCache.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    expect(cache.size).toBe(18);
+    expect(cache.get("1")?.value).toBe(undefined);
+    expect(cache.get("2")?.value).toBe("B");
+  });
+
+  it("allows values that are non-disposable", () => {
+    const cache = new DataCache<string[]>(2);
+    cache.insert("1", ["A"]);
+    cache.insert("2", ["B", "B"]);
+    expect(cache.size).toBe(2);
```
toloudis, line 147: optional/nit: I think this test makes more sense if the cached sizes are actually proportional to the lengths of the arrays (or needs a separate test for "non-disposable with sizes") -- semi-related to my comment about the cache being able to ask the object for its size automatically.

toloudis (approved):

tests/DataCache.test.ts
ShrimpCryptid: Gotcha. I think I'll just revert these to being length 1, since there's already a test for different sizes. I don't expect the disposal code to interact with size at all.

PR #394, opened 2024-06-24
feature: Reserve cache keys

ShrimpCryptid:
Problem
=======
Part 2/3 for #20, handling large numbers of features. This code change updates the `DataCache` to allow certain keys to be exempted from eviction; this way, the feature currently being viewed will not have its Texture data unloaded if other parts of the UI are interacted with\*.

*Estimated review size: small, 15 minutes. Lots of unit tests.*

\*_Okay so interesting technical note, it's possible that this may not be an issue because Three.js will not unload Texture objects that are currently being used by a scene. However, it's not behavior that I want to rely on!_

Solution
========
- Added the ability to mark certain keys in the `DataCache` as reserved. This will prevent them from being evicted until they are removed from the list of reserved keys.
- Added unit tests for validating new behavior.
- Renamed `FrameCache.ts` to `DataCache.ts`.

## Type of change


tests/DataCache.test.ts
```diff
@@ -131,4 +131,88 @@ describe("DataCache", () => {
...
+    const cache = new DataCache<DisposableString>(3);
+    cache.insert("1", new DisposableString("A"));
+    cache.insert("2", new DisposableString("B"));
+    cache.insert("3", new DisposableString("C"));
+    cache.setReservedKeys(new Set(["1", "2"]));
+    expect(cache.size).toBe(3);
+
+    // 3 should be evicted because 1 and 2 are reserved.
+    cache.insert("4", new DisposableString("D"));
+    expect(cache.size).toBe(3);
```
meganrm, line 145: why is the cache size still 3 here? it seems like you added another item?

meganrm: oh! I get it I think, the size remains constant and because you added a new one it overwrote "3" in order to add 4:D

tests/DataCache.test.ts
```diff
@@ -131,4 +131,88 @@ describe("DataCache", () => {
...
+    expect(cache.size).toBe(2);
+
+    cache.insert("3", new DisposableString("C"));
+    expect(cache.size).toBe(2);
+    // 2 is evicted
+    expect(cache.get("1")?.value).toBe("A");
+    expect(cache.get("2")?.value).toBe(undefined);
+    expect(cache.get("3")?.value).toBe("C");
+
+    // Remove A from the reserved keys. It will be added to the front of the list.
```
meganrm, line 166: nit: you're sometimes referring to the key and sometimes the value as the thing. Ie, up above you say "2 is evicted" but here you refer to "Remove A..."

meganrm (approved):
looks good, maybe take a fresh look at your comments in the unit tests and see if you can make anything clearer on first read through

ShrimpCryptid:
> looks good, maybe take a fresh look at your comments in the unit tests and see if you can make anything clearer on first read through

Thanks, will do!

tests/DataCache.test.ts
ShrimpCryptid: I'll add a comment on this! Sorry for the puzzle 😅

src/colorizer/DataCache.ts
```diff
@@ -54,16 +69,17 @@ export default class DataCache<E extends DisposableValue | Object> {
...
     if (this.first) {
       this.first.next = entry;
     }
     entry.prev = this.first;
+    entry.next = null;
     this.first = entry;
   }
 
-  /** Moves an entry currently in the list to the front */
+  /** Moves an EXISTING entry currently in the list to the front */
```
frasercl, line 82: Like all this clarification!

src/colorizer/DataCache.ts
```diff
@@ -124,13 +188,14 @@ export default class DataCache<E extends DisposableValue | Object> {
   public get(key: string | number): E | undefined {
```
frasercl, line 188: Should the whole class, in addition to being generic over its element type `E`, be generic over some `K extends string | number` so that instances can narrow keys to only `number`s or `string`s?

tests/DataCache.test.ts
frasercl, line 1: This test file might be getting long enough to merit sorting the tests into a few `describe`s, but that's not a chore that needs to block this PR.

frasercl (approved):
My comments are more musings than problems, and none need to block this PR. This is neat!

src/colorizer/DataCache.ts
ShrimpCryptid: Made this change!

PR #395, opened 2024-06-24
feature: Load features asynchronously

ShrimpCryptid:
Problem
=======
Part 3/3 to close #20, handling large numbers of features! This solves a big performance problem I was seeing on initial load where the dataset would stall.

Before, the Dataset would **fetch the feature data for EVERY feature** on load. This would cause it to block for up to 10-15 seconds on the latest NucMorph datasets. This change now allows it to only load in feature data on demand, making initial startup MUCH faster.

However, there are still some interactions that could use loading spinners + better feedback when loading is happening, which I've described in #390.

This builds off of the previous changes to the `DataCache` that allows it to also store feature data and protect certain keys from eviction.

*Estimated size: large, ~40 minutes*

Solution
========

- Dataset now asynchronously loads and stores feature data in a `DataCache`.
- When a dataset is opened (loaded), only the first feature will be fetched and loaded. Other features will be loaded on demand.
- The `Viewer` now stores the selected feature's data (with the `featureData` state) instead of just the feature's key. This allows it to do synchronous data lookups.
- Updates other components to use async logic when handling feature data fetching.

## Type of change
* New feature (non-breaking change which adds functionality)

Steps to Verify:
----------------
1. Open the dataset preview link in an incognito/private browser from the main build: https://allen-cell-animated.github.io/timelapse-colorizer/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Large (this takes about 10 seconds on my computer)
3. Close the previous browser page. Open the PR preview link in a new incognito/private browser, which should load much faster: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-395/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Large
4. Click around both. Adding/removing filters or switching features will cause new feature data to be loaded in.

Screenshots (optional):
-----------------------
Show-n-tell images/animations here

Keyfiles (delete if not relevant):
-----------------------
1. `Dataset.ts`
5. `Viewer.tsx`

Thanks for contributing!

PR #396, opened 2024-06-26
feature: Loading spinner for datasets

ShrimpCryptid:
Problem
=======
The nucmorph datasets are loading noticeably slowly (10-15 seconds), so this is a stopgap measure until async feature loading (see #395) can be merged.

Solution
========
- Adds a loading spinner to the main view when datasets are being loaded or switched between.
- Fixes some styling on the loading spinner based on feedback from UX.

## Type of change

Steps to Verify:
----------------
1. Open the PR preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-396/
2. Click on the "Exploratory" dataset, which is the largest and slowest to load. You should see the loading spinner appear.
3. The loading spinner will disappear once the dataset loads.
4. Try switching to a different dataset.

Screenshots (optional):
-----------------------
image

image

### DIFF
--- a/src/routes/LandingPageContent.tsx
+++ b/src/routes/LandingPageContent.tsx
@@ -35,6 +35,36 @@ export const landingPageContent: ProjectEntry[] = [
     // publicationName: "<NucMorph manuscript> (Publisher name, mm/dd/yyyy)",
     datasets: [
       {
+        name: "Baseline colonies dataset",
+        description:
+          "Minimally filtered tracked nuclei with quantitative single-timepoint features (subset of exploratory analysis dataset)",
+        loadParams: {
+          collection:
+            "https://allencell.s3.amazonaws.com/aics/nuc-morph-dataset/timelapse_feature_explorer_datasets/baseline_colonies_dataset/collection.json",
+          time: 15,
+        },
+      },
+      {
+        name: "Full-interphase dataset",
+        description:
+          "Nuclei tracked throughout interphase (subset of baseline colonies analysis dataset, with added growth trajectory features)",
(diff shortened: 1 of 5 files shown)


interim17 (commented):
1) Looks good when loading dataset and seeing wheel over the main canvas, but I'm wondering if/when we should be seeing the wheel over the plots panel?

2) Possible bug/unrelated? When I use the mouse scrool wheel and I'm hovering over the canvas I can scroll into some blank space. Is that intentional?
Screenshot 2024-06-26 at 10 55 57 AM

ShrimpCryptid:
> 1. Looks good when loading dataset and seeing wheel over the main canvas, but I'm wondering if/when we should be seeing the wheel over the plots panel?
> 2. Possible bug/unrelated? When I use the mouse scrool wheel and I'm hovering over the canvas I can scroll into some blank space. Is that intentional?
> 
> Screenshot 2024-06-26 at 10 55 57 AM

1. Yup, that's behavior that was existing before. The main scatterplot panel is set to pause while playback is happening to avoid a performance slowdown. Thinking about it more, I might be able to speed it up in the near future?
2. Scrolling behavior is intended, but we had a discussion about how much the user should be able to view the margins. I'll write it up as an issue ticket and report back here.

interim17 (approved):

src/routes/LandingPageContent.tsx
ShrimpCryptid, line 1: Moved the baseline/other colony datasets to be first because the exploratory dataset takes the most time to load.

meganrm (approved):

PR #404, opened 2024-07-09
feat: Added progress circle for the dataset load

### LINKED ISSUE #403: Show progress bar instead of loading spinner on dataset load
## Use Case

Current NucMorph datasets load very slowly! Showing users progress will help it feel less frustrating.

## Acceptance Criteria

- Replace the loading spinner with a progress bar as dataset resources are loaded in.

## Details

ShrimpCryptid:
Problem
=======
Closes #403, show progress bar instead of loading spinner.

Solution
========
- Updates `LoadingSpinner` to accept an integer percent `progress`. When set, it replaces the loading spinner with the progress bar.
- Updates `Dataset.open` to report loading status via callbacks.
  - These are captured by `Collection` and reported up to the `Viewer`, which stores it as state and passes it into `CanvasWrapper` which controls the loading spinner.
- Moves old console statements out of `Dataset` so they don't trigger during unit testing.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-404/
2. Click on any of the provided datasets. (The exploratory dataset is the largest.)
3. The progress bar should appear as the dataset loads.
4. Switch datasets from the dropdown.

Screenshots (optional):
-----------------------
image

-----------------------
1. `Dataset.tsx`
5. `Viewer.tsx`
6. `LoadingSpinner.tsx`

Thanks for contributing!


ShrimpCryptid:
TODO after talking with Lyndsay:
- [ ] On completion, show same purple color and 100% text instead of checkmark
- [ ] Increase delay before the loading spinner is shown for datasets.

src/Viewer.tsx
```diff
@@ -355,6 +358,16 @@ function Viewer(): ReactElement {
...
   // DATASET LOADING ///////////////////////////////////////////////////////
+
+  const handleProgressUpdate = useCallback(
+    (complete: number, total: number): void => {
+      startTransition(() => {
+        setDatasetLoadProgress(Math.round((complete / total) * 100));
+      });
+    },
+    [setDatasetLoadProgress]
+  );
```
ShrimpCryptid, line 369: The progress value is expected to update very quickly and trigger a lot of re-renders, so I've wrapped this in a transition to allow the render to be interrupted when the progress value updates.

src/Viewer.tsx
```diff
@@ -400,6 +413,7 @@ function Viewer(): ReactElement {
       setSelectedTrack(null);
       setDatasetOpen(true);
       setFeatureThresholds(validateThresholds(newDataset, featureThresholds));
+      console.log("Dataset metadata:", newDataset.metadata);
```
ShrimpCryptid, line 416: Moved from `Dataset.open` so this doesn't trigger during unit tests.

src/colorizer/Collection.ts
```diff
@@ -124,18 +124,32 @@ export default class Collection {
...
+    const onLoadCompleted = (): void => {
+      completedLoadItems++;
+      onLoadProgress?.(completedLoadItems, totalLoadItems);
+    };
+
     // TODO: Override fetch method
     try {
       const dataset = new Dataset(path);
-      await dataset.open();
+      await dataset.open(undefined, onLoadStart, onLoadCompleted);
```
ShrimpCryptid, line 152: This `undefined` here makes me cringe a little... It's an argument that overrides the fetching method `dataset.open` uses, and in this case I don't want to touch. (It's mostly used in unit tests to spoof `fetch` operations.)

Open to input on if I can make this cleaner.

src/Viewer.tsx
```diff
@@ -514,7 +528,8 @@ function Viewer(): ReactElement {
       }
 
       setCollection(newCollection);
-      const datasetResult = await newCollection.tryLoadDataset(datasetKey);
+      setDatasetLoadProgress(null);
```
frasercl, line 531: Question unrelated to this change: I'm noticing this async function is defined within a `useEffect` callback, then the only other thing the callback does is call this function. Could the callback itself not be async?

src/Viewer.tsx
frasercl, line 369: I believe state setters from `useState` are guaranteed to be stable? If so, should this dependency be removed?

src/colorizer/Dataset.ts
```diff
@@ -406,18 +418,27 @@ export default class Dataset {
...
     this.frames = new DataCache(MAX_CACHED_FRAME_BYTES);
 
+    // Wrap an async operation and report progress when it starts + completes
+    const reportLoadProgress = async <T>(promise: Promise<T>): Promise<T> => {
+      onLoadStart?.();
+      return promise.then((result) => {
+        onLoadComplete?.();
+        return result;
+      });
+    };
```
frasercl, line 428: Clever!

src/components/LoadingSpinner.tsx
```diff
@@ -64,20 +80,43 @@ const LoadSpinnerIconContainer = styled.div<{ $fontSize: number }>`
...
+  // would flash the progress bar at 100% right as it vanished.
+  const showProgressBar = useDebounce(props.progress !== undefined && props.progress !== null, VANISH_DURATION_MS);
+
+  // Disable completion checkmark by forcing value to always be shown as a number %
+  const progressFormatter = (percent?: number): ReactNode => {
+    if (percent === undefined) {
+      return "";
+    }
+    return `${percent}%`;
+  };
```
frasercl, line 94: nit: doesn't necessarily need to be defined within the component body

src/colorizer/Dataset.ts
```diff
@@ -111,6 +111,7 @@ export default class Dataset {
 
   private resolveUrl = (url: string): string => `${this.baseUrl}/${url}`;
 
+  // TODO: Make this part of `urlUtils`?
```
frasercl, line 114: Makes sense to me - I'm strongly in favor of moving class methods that never use `this` out of their class.

frasercl (approved):
Looks good, makes the app a lot nicer to use!

src/Viewer.tsx
frasercl: This comment was motivated by seeing this function (`handleProgressUpdate`) in a `useCallback` dependency array further down, thinking for a sec it was a `useEffect`, and wondering what that might mean semantically (change datasets whenever the function that changes the progress bar changes...?). Code makes a lot more sense when you read it right 🙃 and this is ultimately much less important a question than I thought.

ShrimpCryptid: Still agree that this dependency can be removed!

src/Viewer.tsx
ShrimpCryptid: Ah, `useEffect` does not allow async callbacks.

```
Argument of type '() => Promise<void>' is not assignable to parameter of type 'EffectCallback'.
```

src/components/LoadingSpinner.tsx
ShrimpCryptid: Fixed!

src/colorizer/Collection.ts
meganrm: one thing would just be to make it not optional and send in the loader you want here. the other way would be to send in a config object that has the optional settings, instead of a list of arguments, and the config object could have just the onLoadStart and onLoadCompleted, and you wouldn't need to set anything to undefined

src/colorizer/Dataset.ts
```diff
@@ -373,8 +367,21 @@ export default class Dataset {
...
+   * @returns A Promise that resolves when loading completes.
+   */
+  public async open(
+    manifestLoader = urlUtils.fetchManifestJson,
+    onLoadStart?: () => void,
+    onLoadComplete?: () => void
+  ): Promise<void> {
+    if (manifestLoader === undefined) {
+      manifestLoader = urlUtils.fetchManifestJson;
+    }
```
meganrm, line 384: doesn't line 378 already do this?

src/components/LoadingSpinner.tsx
```diff
@@ -59,25 +75,48 @@ const LoadSpinnerIconContainer = styled.div<{ $fontSize: number }>`
...
-            <LoadSpinnerIconContainer $fontSize={props.iconSize}>
-              {/* Make larger loading icon by multiple of them together */}
-              {/* TODO: Make custom loading spinner SVG? */}
-              <LoadingOutlined />
-              <LoadingOutlined rotate={90} />
-              <LoadingOutlined rotate={135} />
-            </LoadSpinnerIconContainer>
-          }
-        ></Spin>
+        {showProgressBar && props.progress !== undefined && props.progress !== null ? (
```
meganrm, line 99: showProgressBar already has a check for props.progress being defined, is it an issue with the debounce that makes you have to recheck of that there? either way, I would move these conditionals into showProgressBar, 
```JavaScript
const showProgressBar = useDebounce(props.progress !== undefined && props.progress !== null, VANISH_DURATION_MS)  &&  props.progress !== undefined && props.progress !== null;
```

ShrimpCryptid: Unfortunately for some reason TypeScript doesn't catch this... I ended up just adding check to cast `null` values to `undefined`, even though both values should never reach the offending line.

src/colorizer/Collection.ts
ShrimpCryptid: Went the route of including a config object!

meganrm (approved):

PR #409, opened 2024-07-12
feature: Accept binary `.parquet` data

### LINKED ISSUE #401: Use binary format for feature data
## Use Case

Loading features is VERY slow currently in our example datasets and we can definitely do better. Features are currently stored as JSON data with an array of numbers; changing this to a binary format would significantly decrease the size of feature files and allow datasets to load faster.

The tradeoff is that they will be harder to read, but if we maintain backwards compatibility with the JSON format it should not be an issue.

## Acceptance Criteria

- Update `colorizer-data` to write to a binary format, most likely 4-byte integer or float values.
  - Will we need to include headers in the data? Are there libraries we can use for this already?
- Update TFE to parse binary data for the feature data.

## Details

- Feature files currently contain a min and max field, which could be moved to the feature metadata stored in the manifest JSON file. This actually would speed up a few interactions on the UI since we wouldn't need to load the feature file to know the min/max (namely feature thresholds).

ShrimpCryptid:
Problem
=======

Closes #401, allowing the viewer to read binary data! See related PR in `colorizer-data` here that writes to the Parquet data format: https://github.com/allen-cell-animated/colorizer-data/pull/53

When testing on our exploratory dataset, the feature files went from 110 MB to only 10.8 MB using Parquet! This lets us open datasets MUCH faster and greatly improves the user experience.

Before: All data stored in JSON files, take up a large amount of space due to being saved as string characters.
```
{
  "data": [1.4, 2.3, 5.9,...] 
}
```

After: Data is stored in a columnar format in binary. Each value should take up no more than 32 bits, since we save values as 32-bit floats. Parquet does a lot of additional optimization so our storage size may be much smaller.
| data |
| --- |
| 1.4 |
| 2.3 | 
| 5.9 |
| ... |

Solution
========
- Adds the `hyparquet` and `hyparquet-compressors` packages to the project for Parquet file parsing.
- Updates the viewer to recognize `min` and `max` for feature data provided in the manifest, instead of needing it to be provided in the feature data file. (Viewer is still backwards-compatible with datasets that do, however.)
  - This lets us use parquet column datasets, which don't have a way to provide a min/max value.
- Refactors `JsonArrayLoader` into `UrlArrayLoader`, which loads and parses both `.json` and `.parquet` filetypes.

## Type of change

Steps to Verify:
----------------
1. Load the example here in your browser:  https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-409/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Fparquet%2Fcollection.json&dataset=Medium+JSON
2. Switch between each dataset to observe the load times.

NOTE: I may swap this out for a link that accesses s3 data once I get access to our test bucket. For now, you can also try throttling the network activity from the network tab to see the difference.

NOTE: There's a known bug I've been trying to track down where Parquet datasets that load very quickly (like in this example) don't show the progress bar. I'm fairly certain this is because too many state updates are being requested, so React is blocking or discarding the state updates. It then only passes the updated progress prop when it reaches 100 and stops updating. There are potentially some hacky workarounds, like detecting the time of last update and then batching progress updates together until a certain time interval is hit, but I don't want to block this PR while I work on that. I also expect that this won't be an issue in environments where the dataset is loading more slowly (e.g. our s3 bucket).

-----------------------
1. main file/entry point
4. other important file

Thanks for contributing!


package-lock.json
ShrimpCryptid, line 1: Ran `npm audit fix`

src/colorizer/loaders/JsonArrayLoader.ts
```diff
@@ -1,6 +1,9 @@
+import { parquetRead } from "hyparquet";
```
ShrimpCryptid, line 1: I will rename this file to `UrlArrayLoader` once reviews come in! Leaving this as `JsonArrayLoader` so it's easier to diff.

src/colorizer/Dataset.ts
```diff
@@ -442,8 +446,10 @@ export default class Dataset {
       }
     });
 
-    if (this.features.size === 0) {
-      throw new Error("No features found in dataset. Is the dataset manifest file valid?");
+    if (this.features.size !== manifest.features.length) {
+      throw new Error(
+        "One or more features could not be loaded. This may be because of an unsupported format or a missing file. See console for details."
+      );
```
ShrimpCryptid, line 452: Behavior question: If one feature fails to load, should the whole dataset load be counted as a failure?

src/Viewer.tsx
```diff
@@ -355,6 +358,13 @@ function Viewer(): ReactElement {
   );
 
   // DATASET LOADING ///////////////////////////////////////////////////////
+
+  const handleProgressUpdate = useCallback((complete: number, total: number): void => {
+    startTransition(() => {
+      setDatasetLoadProgress(Math.round((complete / total) * 100));
+    });
```
ShrimpCryptid, line 365: Many updates are expected to occur, so use a transition to mark the render as interruptible and non-critical.

src/colorizer/Dataset.ts
```diff
@@ -449,16 +453,18 @@ export default class Dataset {
...
         this.features.set(key, data);
       }
     });
 
-    if (this.features.size === 0) {
-      throw new Error("No features found in dataset. Is the dataset manifest file valid?");
+    if (this.features.size !== manifest.features.length) {
+      console.warn(
+        "One or more features could not be loaded. This may be because of an unsupported format or a missing file. See console for details."
+      );
```
frasercl, line 467: "see console" when the user is already in the console? 🤔

src/colorizer/Dataset.ts
```diff
@@ -449,16 +453,18 @@ export default class Dataset {
     }
 
     // Keep original sorting order of features by inserting in promise order.
-    featureResults.forEach((result) => {
-      const featureValue = this.getPromiseValue(result, "Failed to load feature: ");
+    featureResults.forEach((result, index) => {
+      const featureValue = this.getPromiseValue(result, `Failed to load feature ${index}`);
```
frasercl, line 457: Not included in this PR, but `getPromiseValue` looks a lot more like a util function than a class method to me. It doesn't look like it refers to `this` anywhere, though perhaps it might in the future?

src/colorizer/loaders/JsonArrayLoader.ts
```diff
@@ -57,13 +60,37 @@ class JsonArraySource implements ArraySource {
...
+      return new UrlArraySource(data, min ?? jsonMin, max ?? jsonMax);
+    } else if (url.endsWith(".parquet")) {
+      const result = await fetch(url);
+      const arrayBuffer = await result.arrayBuffer();
+      let data: number[] = [];
+      await parquetRead({
+        file: arrayBuffer,
+        compressors,
+        onComplete: (loadedData: number[][]) => {
+          data = loadedData.map((row) => Number(row[0]));
```
frasercl, line 88: 1. Since the array may be large, it might be worth benchmarking `map` against a traditional `for` loop or iterator (`for row of loadedData) { ... }`
2. If you have to loop through the data anyways, could you also be accumulating into a `min` and `max` now?
3. In fact, is there a way the `min` and `max` can come along with the data, as it can in json?

frasercl (approved):
Hooray for binary!

src/colorizer/loaders/JsonArrayLoader.ts
ShrimpCryptid: 1. Ooh, good point, I'll do that.
2. That's a good idea in case the user didn't provide a min and max, though `colorizer-data` should guarantee that parquet data has one. `min` and `max` are meant to be overridden by the user, though, for example in the case where outliers would blow the min/max out of proportion.
3. It might be possible if I include a "min" and "max" column, but if we ever pursue async loading again it's actually much better to have the feature min/max live inside the manifest JSON. That way we have immediate access to the range even if the feature data hasn't been loaded in.

src/colorizer/Dataset.ts
ShrimpCryptid: I don't expect it would ever need to use class-related content... I'll move it

src/colorizer/Dataset.ts
ShrimpCryptid: Ah, yeah. I think I was originally debating between showing this as a full error or just a console warning. Ideally, I think I need `Dataset.open` to report issues via warnings or banners on the viewer page. Ended up opening  https://github.com/allen-cell-animated/timelapse-colorizer/issues/411 to capture this.

interim17 (approved):

PR #420, opened 2024-08-08
feature: Use multithreading when loading feature data

### LINKED ISSUE #416: Fix blocked progress bar while handling Parquet data -> add multithreading support
## Description

Progress bar is blocked and stops at one or two numbers while loading Parquet data, leading to a bad user experience.

## Expected Behavior

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
Closes #416, a bug where the progress bar would be blocked when loading Parquet data. The underlying problem was that the Parquet data would all arrive at once and need to be decoded by the main thread, causing the UI to block. This change adds multithreading using workers to parallelize this step!

Here's some stats on the new load times:
| Dataset | JSON Load | Parquet Load |
| --- | --- | --- |
| Small | 4.54s | 1.38s |
| Medium | 5.25s | 1.66s |
| Large | 5.85s | 1.68s | 

*Estimated size: medium, 20-30 minutes*

Solution
========
This one ended up a tad bit bigger than I would have liked, but that's the magic of ✨TYPE WRANGLING✨! In my initial attempt at this, I found that the UI was still stalling when the workers finished because they were copying the data arrays back to the main thread. To solve this, I wanted to transfer the arrays, which required that the arrays be `TypedArrays`.

This kicked off a big types refactor, but ultimately the `ArraySource`, `IArrayLoader`, and other interfaces/classes were updated to be generic on a `FeatureDataType` provided at initialization. This meant that I could specify the required data array type when handing off work to the worker threads and receive the expected type when they completed.

## Changes
- Installs the `workerpool` library for handling workers.
- Moves loading + parsing for both JSON and Parquet data to worker threads.
- Edits many of the Array loading types to specify a `FeatureDataType` at initialization.
- Data arrays no longer use `Infinity` as a placeholder for `NaN` values for data at rest.
  -  `NaN`s are converted to `Infinity` only when writing to Textures.

I have one more change I'd like to make where I also move the packing of Textures into squares to the worker threads, but I'll handle that later once these initial changes are approved.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-420/
2. Copy in the following URL and click load: `https://animatedcell-test-data.s3.us-west-2.amazonaws.com/parquet/collection.json`.
3. Switch between the JSON and Parquet datasets from the dropdown to compare the load times!

-----------------------
1. `ILoader.ts` -> changes to interface to account for new type interactions
5. `UrlArrayLoader.ts` -> big changes to `UrlArraySource` (it's much simpler than before, basically a data container!), removes old parsing code
6. `urlLoadWorker.ts` -> The new worker script. Copied code from `UrlArrayLoader` with updated handlers for types.

Thanks for contributing!

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -130,7 +130,12 @@ export default class Dataset {
     const url = this.resolveUrl(metadata.data);
     const featureType = this.parseFeatureType(metadata.type);
 
-    const source = await this.arrayLoader.load(url, FeatureDataType.F32);
+    const source = await this.arrayLoader.load(
+      url,
+      FeatureDataType.F32,
+      metadata.min ?? undefined,
+      metadata.max ?? undefined
+    );
 
     const featureCategories = metadata?.categories;
     // Validation
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -8,7 +8,7 @@ import { ManifestFile, ManifestFileMetadata, updateManifestVersion } from "./uti
 import * as urlUtils from "./utils/url_utils";
 
 import DataCache from "./DataCache";
-import { IArrayLoader, IFrameLoader } from "./loaders/ILoader";
+import { IArrayLoader, ITextureImageLoader } from "./loaders/ILoader";
 import ImageFrameLoader from "./loaders/ImageFrameLoader";
 import UrlArrayLoader from "./loaders/UrlArrayLoader";
 import Track from "./Track";
@@ -49,12 +49,12 @@ const defaultMetadata: ManifestFileMetadata = {
 const MAX_CACHED_FRAME_BYTES = 1_000_000_000; // 1 GB
 
 export default class Dataset {
-  private frameLoader: IFrameLoader;
+  private frameLoader: ITextureImageLoader;
   private frameFiles: string[];
   private frames: DataCache<number, Texture> | null;
   private frameDimensions: Vector2 | null;
 
-  private backdropLoader: IFrameLoader;
+  private backdropLoader: ITextureImageLoader;
   private backdropData: Map<string, BackdropData>;
   // TODO: Implement caching for overlays-- extend FrameCache to allow multiple frames per index -> string name?
   // private backdrops: Map<string, FrameCache | null>;
@@ -91,7 +91,7 @@ export default class Dataset {
    * @param frameLoader Optional.
    * @param arrayLoader Optional.
    */
-  constructor(manifestUrl: string, frameLoader?: IFrameLoader, arrayLoader?: IArrayLoader) {
+  constructor(manifestUrl: string, frameLoader?: ITextureImageLoader, arrayLoader?: IArrayLoader) {
     this.manifestUrl = manifestUrl;
 
     this.baseUrl = urlUtils.formatPath(manifestUrl.substring(0, manifestUrl.lastIndexOf("/")));
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -130,7 +130,7 @@ export default class Dataset {
     const url = this.resolveUrl(metadata.data);
     const featureType = this.parseFeatureType(metadata.type);
 
-    const source = await this.arrayLoader.load(url);
+    const source = await this.arrayLoader.load(url, FeatureDataType.F32);
 
     const featureCategories = metadata?.categories;
     // Validation
@@ -148,8 +148,8 @@ export default class Dataset {
       {
         name,
         key,
-        tex: source.getTexture(FeatureDataType.F32),
-        data: source.getBuffer(FeatureDataType.F32),
+        tex: source.getTexture(),
+        data: source.getBuffer(),
         min: source.getMin(),
         max: source.getMax(),
         unit: metadata?.unit || "",
@@ -269,8 +269,8 @@ export default class Dataset {
     }
     try {
       const url = this.resolveUrl(fileUrl);
-      const source = await this.arrayLoader.load(url);
-      return source.getBuffer(dataType);
+      const source = await this.arrayLoader.load(url, dataType);
+      return source.getBuffer();
     } catch (e) {
       return null;
     }
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -60,6 +60,7 @@ export default class Dataset {
   // private backdrops: Map<string, FrameCache | null>;
 
   private arrayLoader: IArrayLoader;
+  private cleanupArrayLoaderOnDispose: boolean;
   // Use map to enforce ordering
   /** Ordered map from feature keys to feature data. */
   private features: Map<string, FeatureData>;
@@ -104,6 +105,7 @@ export default class Dataset {
     this.backdropLoader = frameLoader || new ImageFrameLoader(RGBAFormat);
     this.backdropData = new Map();
 
+    this.cleanupArrayLoaderOnDispose = !arrayLoader;
     this.arrayLoader = arrayLoader || new UrlArrayLoader();
     this.features = new Map();
     this.metadata = defaultMetadata;
@@ -479,6 +481,10 @@ export default class Dataset {
   public dispose(): void {
     Object.values(this.features).forEach(({ tex }) => tex.dispose());
     this.frames?.dispose();
+    // Cleanup array loader if it was created in the constructor
+    if (this.cleanupArrayLoaderOnDispose) {
+      this.arrayLoader.dispose();
+    }
   }
 
   /** get frame index of a given cell id */
--- a/src/colorizer/loaders/ILoader.ts
+++ b/src/colorizer/loaders/ILoader.ts
@@ -18,5 +18,5 @@ export interface IFrameLoader {
 }
 
 export interface IArrayLoader {
-  load<T extends FeatureDataType>(url: string, type: T): Promise<ArraySource<T>>;
+  load<T extends FeatureDataType>(url: string, type: T, min?: number, max?: number): Promise<ArraySource<T>>;
 }
--- a/src/colorizer/loaders/ILoader.ts
+++ b/src/colorizer/loaders/ILoader.ts
@@ -2,21 +2,21 @@ import { Texture } from "three";
 
 import { FeatureArrayType, FeatureDataType } from "../types";
 
-export interface ArraySource {
+export interface ArraySource<T extends FeatureDataType> {
   /** Create a `TypedArray` of the specified type from this data source */
-  getBuffer<T extends FeatureDataType>(type: T): FeatureArrayType[T];
+  getBuffer(): FeatureArrayType[T];
   /** Create a square texture of the specified type from this data source */
-  getTexture(type: FeatureDataType): Texture;
+  getTexture(): Texture;
   /** Get the minimum value of the contained data */
   getMin(): number;
   /** Get the maximum value of the contained data */
   getMax(): number;
 }
 
-interface ILoader<DataType> {
-  /** Begin loading data */
-  load(url: string): Promise<DataType>;
+export interface IFrameLoader {
+  load(url: string): Promise<Texture>;
 }
 
-export interface IFrameLoader extends ILoader<Texture> {}
-export interface IArrayLoader extends ILoader<ArraySource> {}
+export interface IArrayLoader {
+  load<T extends FeatureDataType>(url: string, type: T): Promise<ArraySource<T>>;
+}
--- a/src/colorizer/loaders/ILoader.ts
+++ b/src/colorizer/loaders/ILoader.ts
@@ -19,4 +19,5 @@ export interface IFrameLoader {
 
 export interface IArrayLoader {
   load<T extends FeatureDataType>(url: string, type: T, min?: number, max?: number): Promise<ArraySource<T>>;
+  dispose(): void;
 }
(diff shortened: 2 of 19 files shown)


src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -1,105 +1,84 @@
...
-          data = Array(loadedData.length);
-          for (let i = 0; i < loadedData.length; i++) {
-            const value = Number(loadedData[i][0]);
-            dataMin = dataMin === undefined ? value : Math.min(dataMin, value);
-            dataMax = dataMax === undefined ? value : Math.max(dataMax, value);
-            data[i] = value;
-          }
-        },
-      });
-      return new UrlArraySource(data, min ?? dataMin, max ?? dataMax);
```
ShrimpCryptid, line 100: This code was copied with changes to `urlLoadWorker.ts`.

src/colorizer/loaders/workers/urlLoadWorker.ts
```diff
@@ -0,0 +1,103 @@
...
+  } else if (url.endsWith(".parquet")) {
+    result = await loadFromParquetUrl(url, type);
+  } else {
+    throw new Error(`Unsupported file format: ${url}`);
+  }
+
+  const { min, max, data } = result;
+
+  // TODO: Also generate textures for the data on the worker thread too?
+  // Would need access to the underlying array buffer to transfer
```
ShrimpCryptid, line 96: I'm going to make a follow-up PR that does this! This one is getting a little unwieldy already.

src/colorizer/utils/json_utils.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Replaces all NaN in string text (such as the string representation of a JSON
+ * object) with null. Can be used to safely parse JSON objects with NaN values.
+ */
+export const nanToNull = (json: string): string => json.replace(/NaN/g, "null");
```
ShrimpCryptid, line 5: Had to be moved to its own file because of a dependency loop:
`urlLoadWorker` -> `data_utils` -> `Dataset` -> `UrlArrayLoader` -> `urlLoadWorker`

src/colorizer/Dataset.ts
```diff
@@ -130,7 +132,12 @@ export default class Dataset {
     const url = this.resolveUrl(metadata.data);
     const featureType = this.parseFeatureType(metadata.type);
 
-    const source = await this.arrayLoader.load(url);
+    const source = await this.arrayLoader.load(
+      url,
+      FeatureDataType.F32,
+      metadata.min ?? undefined,
+      metadata.max ?? undefined
```
ShrimpCryptid, line 139: Oops, bug that I missed!

src/components/Tabs/ScatterPlotTab.tsx
```diff
@@ -496,6 +496,8 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
     // Sort data into buckets
     for (let i = 0; i < xData.length; i++) {
       const objectId = objectIds[i];
+      const isMinMaxNaN = Number.isNaN(colorMaxValue) && Number.isNaN(colorMinValue);
```
ShrimpCryptid, line 499: Caused by a change I mentioned where we're no longer using `Infinity` as our placeholder value.

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -1,105 +1,80 @@
...
-  constructor(array: number[] | boolean[], min?: number | boolean, max?: number | boolean) {
-    if (isBoolArray(array)) {
-      this.array = array.map(Number);
-      this.isBool = true;
-    } else {
-      // Must store Infinity values internally because WebGL states that NaN behavior is undefined.
-      // This can cause shaders to not detect NaN, and operations like isnan() fail.
-      // On the UI, however, Infinity should be parsed as NaN for display.
-      this.array = array.map((val) => (val === null ? Infinity : val));
-      this.isBool = false;
```
ShrimpCryptid, line 34: Now handled by `texture_utils.ts`

src/colorizer/loaders/workers/urlLoadWorker.ts
toloudis: how much savings will there be in creating the textures on the worker?  If this is not necessary then why bother?  Save it for later?

src/colorizer/utils/texture_utils.ts
```diff
@@ -1,29 +1,52 @@
...
-import { FeatureDataType, featureTypeSpecs } from "../types";
+import { FeatureArrayType, FeatureDataType, featureTypeSpecs } from "../types";
 
 /**
- * Calculate the squarest possible texture that `data` can fit into, extend it with
- * `emptyVal` to fit perfectly into that space, and return a tuple of `[width, height]`
+ * Calculate the squarest possible texture that `data` can fit into and return a tuple of `[width, height]`
  */
-function fitIntoSquare<T>(data: T[], emptyVal: T): [number, number] {
+function getSquareDimensions<T extends FeatureDataType>(data: FeatureArrayType[T] | number[]): [number, number] {
```
toloudis, line 8: I would rename this to something like `getBestTextureDimensions`... calling it square when it will almost never actually be square can be misleading.

src/colorizer/utils/texture_utils.ts
```diff
@@ -1,29 +1,52 @@
...
+    data = new featureTypeSpecs[type].ArrayConstructor(data);
+  }
 
   const spec = featureTypeSpecs[type];
-  const buffer = new spec.ArrayConstructor(data);
+  const buffer = padToLength(data, type, 0, length);
+
+  // Convert all NaNs in the buffer to Infinity before texture conversion, as WebGL has undefined
+  // behavior for NaNs. This assumes that the data is float-based.
+  for (let i = 0; i < buffer.length; i++) {
```
toloudis, line 45: ouch, that's too bad

toloudis (commented):
workers are super useful.  I have one or two actionable things (like renaming a func) but mostly this is looking good

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -1,105 +1,79 @@
-import { parquetRead } from "hyparquet";
-import { compressors } from "hyparquet-compressors";
+// sort-imports-ignore
 import { DataTexture } from "three";
+import workerpool from "workerpool";
 
-import { FeatureArrayType, FeatureDataType, featureTypeSpecs } from "../types";
-import { nanToNull } from "../utils/data_utils";
+// @ts-ignore Ignore missing file
+import WorkerUrl from "./workers/urlLoadWorker?url&worker";
```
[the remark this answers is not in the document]
ShrimpCryptid: Yes!

src/colorizer/Dataset.ts
ShrimpCryptid: It would use the min/max calculated from the data instead of letting the metadata min/max override it. There were a few features in NucMorph that excluded outliers from the final min/max calculation to avoid blowing out the range.

src/colorizer/Dataset.ts
```diff
@@ -269,8 +276,8 @@ export default class Dataset {
     }
     try {
       const url = this.resolveUrl(fileUrl);
-      const source = await this.arrayLoader.load(url);
-      return source.getBuffer(dataType);
+      const source = await this.arrayLoader.load(url, dataType);
```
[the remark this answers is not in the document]
ShrimpCryptid: This is used to be the outliers, tracks, times, centroids, and bounds files, which don't have associated min/max metadata.

src/colorizer/loaders/ILoader.ts
```diff
@@ -2,21 +2,22 @@ import { Texture } from "three";
...
   /** Get the minimum value of the contained data */
   getMin(): number;
   /** Get the maximum value of the contained data */
   getMax(): number;
 }
 
-interface ILoader<DataType> {
-  /** Begin loading data */
-  load(url: string): Promise<DataType>;
+export interface IFrameLoader {
```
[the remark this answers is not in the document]
ShrimpCryptid: I like TextureImageLoader!

src/colorizer/loaders/workers/urlLoadWorker.ts
ShrimpCryptid: This was a good reminder for me to go back and validate my claims 😄

image

It's adding about 2-3 milliseconds each time a feature is loaded. It's small relative to the React re-render but it's there. I think it will be a relatively small lift though.

src/colorizer/utils/texture_utils.ts
ShrimpCryptid: We were doing this implicitly before but I'm glad it's being handled closer to the Textures now!

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -1,105 +1,80 @@
-import { parquetRead } from "hyparquet";
-import { compressors } from "hyparquet-compressors";
+// sort-imports-ignore
 import { DataTexture } from "three";
+import workerpool from "workerpool";
 
-import { FeatureArrayType, FeatureDataType, featureTypeSpecs } from "../types";
-import { nanToNull } from "../utils/data_utils";
+// Vite import directive for worker files! See https://v3.vitejs.dev/guide/features.html#import-with-query-suffixes.
```
ShrimpCryptid, line 5: Note that Vite's documentation claims that `import` in workers will be fine in production but fail everywhere but Chrome in dev mode? But I've tested this locally in Firefox and Edge without seeing any issues.

ShrimpCryptid: Wait nevermind that was vite v3

toloudis (approved):

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -1,105 +1,80 @@
...
-    }
     return this.max;
   }
 }
 
 export default class UrlArrayLoader implements IArrayLoader {
+  private workerPool: workerpool.Pool;
+
+  constructor() {
+    // TODO: Maintain a single worker pool for all loaders/all asynchronous operations in the app?
```
frasercl, line 48: This would be neat!

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -1,105 +1,80 @@
...
+    this.workerPool = workerpool.pool(WorkerUrl, {
+      workerOpts: {
+        // Set worker type to undefined (classic)  in production to fix a Vite issue where the application.
```
frasercl, line 51: where the application what?

src/colorizer/utils/texture_utils.ts
```diff
@@ -1,29 +1,54 @@
...
+  const length = width * height;
+
+  // If provided a regular array of numbers, convert to TypedArray
+  if (Array.isArray(data)) {
+    data = new featureTypeSpecs[type].ArrayConstructor(data);
+  }
 
   const spec = featureTypeSpecs[type];
-  const buffer = new spec.ArrayConstructor(data);
+  const buffer = padToLength(data, type, 0, length);
```
frasercl, line 43: If `data` is a regular JS array, it gets copied into a typed array twice here, since `padToLength` has to create a new buffer with the desired length and copy in the data. It looks like the method it uses to do this (`TypedArray.set`) can accept a JS array as well. Could we forego the first copy and extend the type signature of `padToLength` to accept a regular array too? (If so, `padToLength` might need a new name to reflect the extra responsibility, like `intoPaddedTypedArray` or something.)

src/colorizer/Dataset.ts
```diff
@@ -104,6 +105,7 @@ export default class Dataset {
     this.backdropLoader = frameLoader || new ImageFrameLoader(RGBAFormat);
     this.backdropData = new Map();
 
+    this.cleanupArrayLoaderOnDispose = !arrayLoader;
```
frasercl, line 108: This is smart and I like it! Reminds me a bit of some hacks I've seen around the neuroglancer codebase, which invents its own reference-counting types to ensure shared resources get cleaned up properly.

frasercl (approved):
Looks great overall!

vite.config.js
```diff
@@ -13,6 +13,13 @@ process.env = {
 
 export default defineConfig({
   build: {
+    // This quiets the "module has been externalized for browser compatibility" warnings that
+    // vite throws when building for production.
+    // commonjsOptions: {
+    //   // Ignore built-in modules in Node.js. Fix copied from the workerpool documentation:
+    //   // https://github.com/josdejong/workerpool/blob/master/examples/vite/vite.config.ts
+    //   ignore: ["os", "child_process", "worker_threads"],
+    // },
```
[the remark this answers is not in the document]
ShrimpCryptid: Dan and I had a discussion about this; I'm going to keep it in case we ever need to track down where that error is coming from.

src/colorizer/loaders/UrlArrayLoader.ts
ShrimpCryptid: Fixed :X

ShrimpCryptid: :X

src/colorizer/utils/texture_utils.ts
ShrimpCryptid: This turned out to be easier than expected!

PR #428, opened 2024-08-23
refactor: Use react-select for selection dropdowns

### LINKED ISSUE #408: Align dropdowns with WCAG rules (keyboard inputs)
## Description

Dropdowns do not allow for arrow key selection, and also don't automatically show where the user last selected. Either revert to using just the Ant dropdown or implement the missing behaviors?

See https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

## Expected Behavior

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
Closes #408, aligning dropdowns with WCAG guidelines. Dropdowns now respond to arrow key inputs and can be searched directly from the input area. It also makes them much easier to navigate instead of pressing `Tab` repeatedly.

*Estimated review size: large, 40 minutes. (~300 lines are just dependencies)*

This took me so much longer than I thought it would (sorry Dan 🙏). 

This first PR focuses on changing `SelectionDropdown` to use the `react-select` component internally. A second PR will add improved keyboard navigation to the `AccessibleDropdown.tsx` component (which is only used in the Help menu). There will probably also need to be a third at some point that implements this for the ColorRampDropdown. **This change will not be merged and made public until all of these are completed.**

Just for discussion, I did try and use Ant's existing `Select` component as well for this; changes can be seen on the `refactor/use-ant-select` branch. However, it broke down in a few specific ways:
- Inconsistent tooltip behavior on the main selection input
- Not fully implementing the WCAG/aria spec
- Tooltips on options would not respond to focus inputs and had weird positioning issues
- Current selection would be offscreen due to bad scroll behavior if I modified the option height at all. This was the big killer because it meant you couldn't see what you were currently selecting.

Solution
========
- Adds `StyledSelect.tsx`, a wrapper around the `react-select` `Select` component which applies styling to make it mimic the existing style.
- Replaced use of `AccessibleDropdown` in `SelectionDropdown` with `StyledSelect`
  - `SelectionDropdown` now supports custom tooltips when hovering over options.
  - Also added support for the `aria-labelledby` attribute.
- Updated usage of `SelectionDropdown` (due to change from `key` to `value` field).

## Type of change

Steps to Verify:
----------------
1. Open https://timelapse.allencell.org. Open any of the datasets, then use tab navigation to select the feature dropdown.
2. Try to select the last feature in the dropdown list.
3. Open the new build and open the same dataset. https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-428/
4. Repeat the same steps. You should be able to use the arrow keys to open the features dropdown.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/c5cf1a1b-5893-42a2-9826-55b8b171b6d5

-----------------------
1. SelectionDropdown.tsx
2. StyledSelect.tsx

Thanks for contributing!


src/components/Dropdowns/StyledSelect.tsx
```diff
@@ -0,0 +1,225 @@
...
+const DropdownIndicator = (props: DropdownIndicatorProps): ReactElement => {
+  return (
+    <components.DropdownIndicator {...props}>
+      <DropdownSVG style={{ width: "12px", height: "12px" }} viewBox="0 0 12 12" />
+    </components.DropdownIndicator>
+  );
+};
+
+// TODO: Add open/close animation to Menu; see similar implementation in
+// ColorRampDropdown.css.
```
ShrimpCryptid, line 202: Captured as #429

interim17 (commented):
I'm reading through the code right now, mostly through it and things are looking good and making sense! Will leave comments if I find anything specific.

However, when I tab through the website using tab to advance and enter to select, I am able to open the Help and Color Map dropdowns, but the Dataset and Feature dropdowns aren't working for me.

An input cursor shows up like it is expecting me to type in the input and the dropdowns don't open.

Similarly, if I click to open those dropdowns and then try to tab advance through the options it moves focus to the next item after the dropdown, whereas if I click to open the Help menu, I am then able to tab through the options.

interim17 (approved):
Ok, LGTM, thanks for answering my questions.

interim17:
> Ok, LGTM, thanks for answering my questions.

ShrimpCryptid:
> I'm reading through the code right now, mostly through it and things are looking good and making sense! Will leave comments if I find anything specific.
> 
> However, when I tab through the website using tab to advance and enter to select, I am able to open the Help and Color Map dropdowns, but the Dataset and Feature dropdowns aren't working for me.
> 
> An input cursor shows up like it is expecting me to type in the input and the dropdowns don't open.
> 
> Similarly, if I click to open those dropdowns and then try to tab advance through the options it moves focus to the next item after the dropdown, whereas if I click to open the Help menu, I am then able to tab through the options.

We talked about this in person but I'll leave a reply here!
- Yes, this behavior is intended. You can press the spacebar (which emulates clicking for keyboard users) or press down to open.
  - See matching ARIA spec here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/ (enter only closes the listbox and doesn't open it)
- The input cursor is intended! You can type into that field to search. 
- The Help menu is actually incorrect and still needs to be fixed to be consistent with this pattern

src/components/Dropdowns/SelectionDropdown.tsx
```diff
@@ -1,107 +1,123 @@
...
+    >
+      <div>
+        <components.Option {...props} />
+      </div>
+    </Tooltip>
+  );
 };
 
+/** Converts an array of strings or SelectItems to an array of SelectItems. */
+function formatAsSelectItems(items: string[] | SelectItem[]): SelectItem[] {
```
meganrm, line 79: I'm a little confused why this is necessary. Is this inconsistently in the datasets themselves? Could a change be made in your data processing to just always have this array bye of type "SelectItems" instead of doing a conversion here?

src/components/Tabs/ScatterPlotTab.tsx
```diff
@@ -796,7 +795,7 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
           label={"X"}
           selected={xAxisFeatureKey || ""}
           items={menuItems}
-          onChange={(key) => props.updateScatterPlotConfig({ xAxis: key })}
+          onChange={(key: string) => props.updateScatterPlotConfig({ xAxis: key })}
```
meganrm, line 798: for consistency: should this still be `key`? or should it be changed to `value` as well?

meganrm (approved):
Looks good, I had one question about data processing but if you agree will probably still be a separate set of work.

src/components/Dropdowns/SelectionDropdown.tsx
ShrimpCryptid: Ah, this is for API convenience. I like that it lets components that use `SelectionDropdown` be unaware of the `SelectItem` type.

ShrimpCryptid:
Merging into `fix/all-wcag-dropdowns`!

PR #431, opened 2024-08-27
feat: Canvas now uses click + drag to pan, ignores scroll inputs

### LINKED ISSUE #398: Don't pan with scroll
## Use Case

@interim17 pointed out in #396 that behavior when scrolling over the main viewport is unexpected, as it causes grey space to be visible around the margins.

image

Instead, we could restrict the panning feature so the corners of the viewport must be within the canvas (unless stretched due to aspect ratio etc.).

The minimum zoom should be 100%, because users likely will not need to zoom out further.

Also, should scrolling when the canvas is at its limits cause the page to scroll instead?

## Acceptance Criteria

- Panning stays within the edges of the frame
- Zooming is capped at 100%

## Details

ShrimpCryptid:
Problem
=======
Closes #398, "don't pan with scroll."

This change takes some ideas from mouse interaction on Google Maps! Clicking and dragging now pans the canvas, and scrolling scrolls the whole page unless the Ctrl or Cmd keys are held. I've also added a little extra buffer zone when panning, so clicks that accidentally move the cursor slightly won't cause the page to pan. (the same behavior happens on Google Maps too)

Solution
========
- Changes panning to be activated on left click and drag.
  - The mouse must move a total of at least 5 pixels before panning activates and replaces object selection behavior. This adds in extra safety for users on laptops or with motor control issues. 
- Removes unused SVG assets.

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


src/components/CanvasWrapper.tsx
```diff
@@ -422,27 +421,24 @@ export default function CanvasWrapper(inputProps: CanvasWrapperProps): ReactElem
...
-        // to be a drag and disable track selection.
+        // to be a drag operation. Start panning and disable track selection.
         totalMouseDrag.current.x += Math.abs(event.movementX);
         totalMouseDrag.current.y += Math.abs(event.movementY);
-        if (!isMouseDragging.current && totalMouseDrag.current.length() > MIN_DRAG_THRESHOLD_PX) {
+        if (totalMouseDrag.current.length() > MIN_DRAG_THRESHOLD_PX) {
           isMouseDragging.current = true;
+          handlePan(event.movementX, event.movementY);
         }
+      }
```
ShrimpCryptid, line 433: We only start panning once the mouse has moved more than a total of `MIN_DRAG_THRESHOLD_PX`. If a user clicks and their mouse moves slightly, this means that the canvas won't immediately pan.

src/assets/images/index.ts
```diff
@@ -1,8 +1,7 @@
-import AicsLogoSVGAsset from "./AICS-logo.svg?react";
 import AicsLogoAndNameSVGAsset from "./AICS-logo-and-name.svg?react";
+import AicsLogoSVGAsset from "./AICS-logo.svg?react";
```
toloudis, line 2: static analysis wants to sort these imports.

toloudis (commented):
just autosort the imports in a couple places and LGTM

src/assets/images/index.ts
ShrimpCryptid: The girls* are fighting
(*eslint and prettier)

toloudis (approved):

ascibisz (approved):
LGTM!

PR #434, opened 2024-08-28
feature: Multithread texture packing

### LINKED ISSUE #422: Use workers when packing textures
## Use Case

Extension of #416 as described in #420. Currently, there's a slowdown when a worker thread finishes and returns the feature data to the main thread, because the main thread does some work when handling textures to pack them into a square. This could easily be done on the worker threads and the underlying array buffer transferred to the main thread.

## Acceptance Criteria

- Move Texture calculation steps to the worker thread
- If possible, transfer Texture ownership to the main thread.
  - Refactor the texture handling functions to return a reference to the underlying array buffer.

## Details

ShrimpCryptid:
Problem
=======

Closes #422, use workers when packing textures. This skips a VERY minor slowdown on the main UI-handling thread when the texture data was being processed.

On average (n=18 for new vs. old loader) the exploratory datasets finish processing about 30-150ms faster, but there's a lot of noise in the time trials I collected based on network and computer performance. S3 may have also been caching and serving files faster because of the repeated requests for them.

image

Solution
========
- Perform the texture packing step on the worker instead of the main thread.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-434/
2. Open any of the datasets and verify that the feature data loads and looks correct.


src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -7,29 +7,31 @@ import workerpool from "workerpool";
...
     this.max = max;
+    this.texture = texture;
   }
 
   getBuffer(): FeatureArrayType[T] {
     return this.array;
   }
 
   getTexture(): DataTexture {
-    return packDataTexture(this.array, this.type);
```
ShrimpCryptid, line 32: This step, `packDataTexture`, is now done once on the worker thread instead of whenever `getTexture()` is called.

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -71,8 +73,11 @@ export default class UrlArrayLoader implements IArrayLoader {
     if (!url.endsWith(".json") && !url.endsWith(".parquet")) {
       throw new Error(`Unsupported file format for URL array loader: ${url}`);
     }
-    const { data, min: newMin, max: newMax } = await this.workerPool.exec("load", [url, type]);
-    return new UrlArraySource<T>(data, type, min ?? newMin, max ?? newMax);
+    const { data, texImage, min: newMin, max: newMax } = await this.workerPool.exec("load", [url, type]);
+    // Reconstruct instance of DataTexture on the main thread
+    const tex = typedArrayToDataTexture(texImage.data, type, texImage.width, texImage.height);
```
frasercl, line 78: `Texture` has a lot of properties defined on it. If you're just using the data, width, and height here on the main thread side, why not just send those?

src/colorizer/loaders/workers/urlLoadWorker.ts
```diff
@@ -86,12 +87,12 @@ async function load(url: string, type: FeatureDataType): Promise<Transfer> {
   }
 
   const { min, max, data } = result;
-
-  // TODO: Also generate textures for the data on the worker thread too?
-  // Would need access to the underlying array buffer to transfer
+  // Cannot directly transfer the DataTexture, since the class methods are not transferred.
+  // Instead, transfer the underlying image and reconstruct it on the main thread.
+  const tex = packDataTexture(data, type);
```
frasercl, line 92: Following on from above: might make sense to inline most of what `packDataTexture` is doing here:

```typescript
const [width, height] = getSquarestTextureDimensions(data);
const buffer = padToLength(data, type, 0, length);
// (convert NaN to Infinity - maybe break that out into a utility method too?)
```
... then transfer `{ min, max, width, height, data, buffer }`

frasercl (approved):
Looking good overall! Some comments on simplifying your message type that I don't think are blocking.

src/colorizer/loaders/UrlArrayLoader.ts
ShrimpCryptid: Did some refactoring, let me know what you think!

frasercl: I like it!

meganrm (approved):

PR #436, opened 2024-08-29
fix: Fixed blurry bold text caused by Lato not loading from Google Fonts

ShrimpCryptid:
Fixes an issue where bolded fonts looked blurry. The root issue was that the font (Lato) was not actually being loaded from Google Fonts, and it was using a fallback font file (without bolding) instead. The browser was using a default bolding algorithm that looked bad at smaller sizes.

Before:
image

After:
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,6 +14,10 @@
       })(window, document, "script", "dataLayer", "GTM-KB579ZN7");
     </script>
 
+    <style>
+      @import url("https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400;0,700;0,900;1,400;1,700;1,900&display=swap");
+    </style>
+
     <title>Timelapse Feature Explorer</title>
     <link rel="icon" type="image/svg+xml" href="/images/aics-logo.svg" />
     <link rel="stylesheet" href="src/styles/cookie-banner.css" />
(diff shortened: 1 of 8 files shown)


index.html
```diff
@@ -14,6 +14,10 @@
       })(window, document, "script", "dataLayer", "GTM-KB579ZN7");
     </script>
 
+    <style>
+      @import url("https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400;0,700;0,900;1,400;1,700;1,900&display=swap");
+    </style>
+
```
toloudis, line 20: not sure I understand the fix: is it moving the reference into index.html, or is it those `wght@0,400,700...` numbers?

ShrimpCryptid: It's moving the reference to `index.html`. Before, it was being handled in `main.tsx`, which I think caused React to escape or ignore the import.

Specifying the weights is optional but also decreases the size of the asset bundle Google Fonts will send over, since it won't try to load every font weight.

toloudis (approved):

src/components/GlossaryPanel.tsx
ascibisz, line 1: Does this PR need a rebase or something? This feels out of scope based on what's described in the PR summary...

ShrimpCryptid: oops! Yes, I forgot that one of my other PRs merged into this one and that it was still outstanding.

The changes are just these two files:
- index.html
- src/components/AppStyle.tsx

ascibisz (approved):

PR #444, opened 2024-09-26
fix: Arrow keys no longer advance time when focused on an input

### LINKED ISSUE #308: Bug: Arrow key keybinds still change time when other inputs are being used
## Description

Split out from #98.

- The left/right arrow keys are currently mapped so they advance the current timestamp. 
- When selecting/editing text in input boxes (or on range sliders, etc.), pressing the left or right arrow keys causes the current timestamp to also advance or reverse.

## Expected Behavior

- Time should not be affected when other inputs are focused. Add a listener to check for the currently focused item...?

List of inputs that should override time keybinding behavior:
- Number and string inputs
- Sliders

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
Closes #308, "arrow keys still change time when other inputs are being used."

*Estimated review size: tiny, 1-2 minutes*

## Type of change

Steps to Verify:
----------------
1. Open TFE and open the export field. Try editing the prefix and pressing the left or right arrow keys. The timestamp will change. https://timelapse.allencell.org/
2. Repeat with fixed (!!!) preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-444/ (The timestamp should not change!)

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -688,10 +688,13 @@ function Viewer(): ReactElement {
   timeControls.setFrameCallback(setFrame);
 
   const handleKeyDown = useCallback(
-    ({ key }: KeyboardEvent): void => {
-      if (key === "ArrowLeft" || key === "Left") {
+    (e: KeyboardEvent): void => {
+      if (e.target instanceof HTMLInputElement) {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


rugeli (approved):
LGTM

PR #453, opened 2024-10-15
feature: Improved load errors

ShrimpCryptid:
Problem
=======
Closes #411, "Show more informative errors when datasets fail to load".

This change adds additional catches for errors and adds more informative messages that can be shown to users. It also adds explicit onscreen warnings for issues that were previously only logged to the console.

*Estimated size: medium, 30 minutes*

Solution
========
- Added callbacks for errors and warnings in the Viewer which can be passed to the Collection + Dataset classes during load operations.
- Added server error handlers in file loaders for 404 errors, etc.
- Datasets can now load even when the times data is missing. (This causes the scatterplot and plot views to not work correctly.)

New warnings and errors:
- Added warnings for missing and duplicated feature and data files.
- Added explicit error messages for missing properties in collection and dataset files.
- The load dialog now displays error messages for both collection and dataset files when the file type is ambiguous.

## Type of change
* New feature (non-breaking change which adds functionality)

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
Preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-453

| Name | Test Steps | Expected behavior/error message | 
| -- | -- | -- |
| 404 error | 1. Open the preview link and open the load dialog. <br/>2. Paste in this URL: https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/bad-data/no-exist | "Error: Could not load the provided URL as either a collection or a dataset: server returned a 404 (Not Found) code." |
| Empty collection | 1. Open the preview link and open the load dialog. <br/>2. Paste in this URL: https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/bad-data/empty-collection.json | "The collection JSON was loaded but no datasets were found. At least one dataset must be defined in the collection." |
| Duplicate collection keys | 1. Open this link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-453/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Fbad-data%2Fcollection-with-duplicates.json | You should see a warning banner with the text "Duplicate dataset names were found in the collection." |
[...]

PR #454, opened 2024-10-17
feature: Error and alert formatting

ShrimpCryptid:
Problem
=======
Follow up to #453 that applies some additional styling. Adds bullet point formatting and styles error message notifications.

*Estimated size: small, 10-15 minutes*

Solution
========
- Adds the new `renderStringArrayAsJsx` function which can parse string arrays into lists of `p` and `li` elements.
  - Also added unit tests for the new function.
- Styled popup notifications with the same fill + outline colors as the alert banners

## Type of change

Steps to Verify:
----------------

1. Open https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-454/ and try loading the URL `https://dev-aics-dtp-001.int.allencell.org/dan-data/colorizer/data/test/bad-data/empty-ambiguous.json`. You should see an error message with two bullet points.
2. Open https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-454/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fdan-data%2Fcolorizer%2Fdata%2Ftest%2Fbad-data%2Fcollection.json and click through all the different datasets. The alert notification popups and warning banners will appear with the new formatting.

Screenshots (optional):
-----------------------
New bullet points:
image
image

New popup notification styling:

image

image

### DIFF
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -331,7 +331,7 @@ export default class Collection {
    * @returns a new Collection, where the only dataset is that of the provided `datasetUrl`.
    * The `url` field of the Collection will also be set to `null`.
    */
-  public static async makeCollectionFromSingleDataset(datasetUrl: string): Promise<Collection> {
+  public static makeCollectionFromSingleDataset(datasetUrl: string): Collection {
     // Add the default filename if the url is not a .JSON path.
     if (!isJson(datasetUrl)) {
       datasetUrl = formatPath(datasetUrl) + "/" + DEFAULT_DATASET_FILENAME;
@@ -419,7 +419,7 @@ export default class Collection {
     // Could not load as a collection, attempt to load as a dataset.
     if (!result) {
       try {
-        const collection = await Collection.makeCollectionFromSingleDataset(url);
+        const collection = Collection.makeCollectionFromSingleDataset(url);
         // Attempt to load the default dataset immediately to surface any loading errors.
         const loadResult = await collection.tryLoadDataset(collection.getDefaultDatasetKey());
         if (!loadResult.loaded) {
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -306,6 +306,7 @@ export default class Collection {
       throw new Error(LoadErrorMessage.COLLECTION_HAS_NO_DATASETS);
     }
     const collectionData: Map<string, CollectionEntry> = new Map();
+    const duplicateDatasetNames = new Set<string>();
     for (const entry of collection.datasets) {
       collectionData.set(entry.name, entry);
     }
@@ -331,7 +332,7 @@ export default class Collection {
    * @returns a new Collection, where the only dataset is that of the provided `datasetUrl`.
    * The `url` field of the Collection will also be set to `null`.
    */
-  public static makeCollectionFromSingleDataset(datasetUrl: string): Collection {
+  public static async makeCollectionFromSingleDataset(datasetUrl: string): Promise<Collection> {
     // Add the default filename if the url is not a .JSON path.
     if (!isJson(datasetUrl)) {
       datasetUrl = formatPath(datasetUrl) + "/" + DEFAULT_DATASET_FILENAME;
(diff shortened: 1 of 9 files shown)


src/components/AppStyle.tsx
```diff
@@ -29,10 +29,14 @@ const palette = {
   link: "#0094FF",
   linkDark: "#007FD6",
   warning: "#faad14",
-  successLight: "#b7eb8f",
-  errorLight: "#ffa39e",
-  infoLight: "#91d5ff",
-  warningLight: "#ffe58f",
+  successMedium: "#b7eb8f",
+  successLight: "#f6ffed",
```
interim17, line 33: Something sort of sad to me about a color named `successLight`
*Non blocking comment, just don't want to paint my room with it*

src/utils/formatting.tsx
```diff
@@ -0,0 +1,59 @@
...
+      currListElements.push(<li key={currListElements.length}>{item.substring(2)}</li>);
+      continue;
+    } else {
+      if (currListElements.length > 0) {
+        elements.push(<ul key={i - 1}>{currListElements}</ul>);
+        currListElements = [];
+      }
+      elements.push(<p key={i}>{item}</p>);
+    }
+  }
```
interim17, line 52: This is a neat utility. It's only called in a couple places and the behavior is pretty bespoke, so given the amount of test code that's necessary I'm curious the rationale for adding it, as opposed to wrapping `map` calls in the new `RenderedStringContainer`?

Non-blocking, just curious.

interim17 (approved):
LGTM, left a non-blocking question a question for my own curiosity

src/utils/formatting.tsx
ShrimpCryptid: The old map calls didn't render list items as `li` or `ul` elements, just as `p` elements, so we would have lost some of implicit HTML styling/accessibility attributes we get from lists. I had gotten some feedback from UX about having the text look more like bullet points, so I figured it would be better to explicitly transform them into HTML unordered lists.

If this ever needs to handle more cases it'll probably be better to use a Markdown to HTML renderer.

Test code was optional but figured I might as well throw it in!

interim17: Thanks for context! Always keeping an eye on your custom tooling, for use in other repos, especially in V2 era.

src/utils/formatting.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const elements: ReactNode[] = [];
+  let currListElements: ReactNode[] = [];
+
+  for (let i = 0; i < items.length; i++) {
+    const item = items[i].trim();
+    if (item.startsWith("- ")) {
+      currListElements.push(<li key={currListElements.length}>{item.substring(2)}</li>);
+      continue;
```
frasercl, line 44: nit: redundant `continue`?

src/colorizer/Collection.ts
```diff
@@ -331,7 +331,7 @@ export default class Collection {
    * @returns a new Collection, where the only dataset is that of the provided `datasetUrl`.
    * The `url` field of the Collection will also be set to `null`.
    */
-  public static makeCollectionFromSingleDataset(datasetUrl: string): Collection {
+  public static async makeCollectionFromSingleDataset(datasetUrl: string): Promise<Collection> {
```
frasercl, line 334: Q: I see a couple new `await`s to support this `async` (farther down this file, in Viewer.tsx, and in tests), but no new `await`s within this function to justify the `async`. Is this in anticipation of some async behavior you plan to add later?

frasercl (approved):

src/utils/formatting.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * @param containerStyle optional CSS properties object that will be applied to the container div.
+ * @returns A list of one or more text elements, wrapped in a styling div.
+ * String items will be returned as one of the following:
+ * - If it starts with "- ", the item will be rendered as a list item (`li`) element. Groups of `li` elements
+ * will be wrapped in a unordered list (`ul`) element.
+ * - Otherwise, the item is wrapped in a `p` element.
+ */
+export function renderStringArrayAsJsx(items: string[] | string | undefined): ReactElement {
+  if (!items || items.length === 0) {
+    return <></>;
```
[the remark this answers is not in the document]
ShrimpCryptid: Changed the return type to `ReactElement | undefined` and added an explicit check in the tests for `undefined` values. Thanks!

src/colorizer/Collection.ts
ShrimpCryptid: Ah, removed this. I thought I'd reversed this change, but I may have lost it while I was squashing some commits on the previous branch.

Issue #456, opened 2024-10-18
Allow scatterplot to update in real time


ShrimpCryptid:
## Use Case

I previously prevented the scatterplot from updating during playback to prevent slowdown/lag, which shows as a big loading spinner.

![image](https://github.com/user-attachments/assets/ed910e2a-cdf9-4bc5-8100-2572fbbcad6f)

If we move some of the calculation steps that determine color and point binning to a worker, it may be performant enough to run in real time. There may be other optimizations I can make too (like only recalculating certain steps if its dependencies change)

## Acceptance Criteria

_Please describe how you know this is done_

## Details

_Please provide any helpful specifications_

ShrimpCryptid:
Also, if we can move the crosshair rendering to a layout shape, then Plotly will not need to re-render the entire plot to show the crosshair moving along the track.

Issue #423, opened 2024-08-19
Improve filter/threshold performance


ShrimpCryptid:
## Use Case

Each time the filter UI is interacted with, a new texture is calculated behind the scenes to determine which IDs are in range and which are not.

If N is the number of object IDs in the dataset and M is the number of filters, in the worst case it will run in `O(MN)` time, which can cause noticeable delay in the UI.

It would be great to offload this work to a worker thread so we don't block the main UI thread.

## Acceptance Criteria

- Move the filter calculation to a worker.
- Interrupt existing worker when the filter is updated again.
- Show a loading spinner(?) when the filter is being updated and takes more than a second to change.

## Details

_Please provide any helpful specifications_

PR #458, opened 2024-10-21
refactor: Generalize worker pool

ShrimpCryptid:
Problem
=======
First step towards #456 and #423, and for improving the scatterplot performance. This change makes the worker pool more general and allows for it to be extended with new background tasks in the near future. Most of the line count is just code being moved around!

Solution
========
- Moved worker-related logic out of `UrlArrayLoader` and into a new class, `SharedWorkerPool`.
- Renamed `urlLoadWorker` to `worker.ts` and moved most of the loading logic into its own utils file.
- Renamed some methods to be more specific.
- Fixes to auto-import sorting rules.

Steps to Verify:
----------------
1. Load any dataset from the example URL: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-458/

### DIFF
--- a/src/colorizer/loaders/UrlArrayLoader.ts
+++ b/src/colorizer/loaders/UrlArrayLoader.ts
@@ -1,11 +1,10 @@
-// sort-imports-ignore
 import { DataTexture } from "three";
 
 import { FeatureArrayType, FeatureDataType } from "../types";
 import { infoToDataTexture } from "../utils/texture_utils";
 
-import { ArraySource, IArrayLoader } from "./ILoader";
 import SharedWorkerPool from "../workers/SharedWorkerPool";
+import { ArraySource, IArrayLoader } from "./ILoader";
 
 export class UrlArraySource<T extends FeatureDataType> implements ArraySource<T> {
   array: FeatureArrayType[T];
@@ -41,10 +40,10 @@ export class UrlArraySource<T extends FeatureDataType> implements ArraySource<T>
 
 export default class UrlArrayLoader implements IArrayLoader {
   private workerPool: SharedWorkerPool;
-  private disposeOfWorkerPool: boolean;
+  private cleanupWorkerPoolOnDispose: boolean;
 
   constructor(workerPool?: SharedWorkerPool) {
-    this.disposeOfWorkerPool = workerPool === undefined;
+    this.cleanupWorkerPoolOnDispose = workerPool === undefined;
     this.workerPool = workerPool ?? new SharedWorkerPool();
   }
 
@@ -65,14 +64,14 @@ export default class UrlArrayLoader implements IArrayLoader {
         `Encountered unsupported file format when loading data. URL must end in '.parquet' or '.json': ${url}`
       );
     }
-    const { data, textureInfo, min: newMin, max: newMax } = await this.workerPool.load(url, type);
+    const { data, textureInfo, min: newMin, max: newMax } = await this.workerPool.loadUrlData(url, type);
 
     const tex = infoToDataTexture(textureInfo);
     return new UrlArraySource<T>(data, tex, type, min ?? newMin, max ?? newMax);
   }
 
   dispose(): void {
-    if (this.disposeOfWorkerPool) {
+    if (this.cleanupWorkerPoolOnDispose) {
       this.workerPool.terminate();
     }
   }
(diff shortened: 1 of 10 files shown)


src/colorizer/utils/data_load_utils.ts
```diff
@@ -1,27 +1,29 @@
...
 };
 
-async function loadFromJsonUrl<T extends FeatureDataType>(url: string, type: T): Promise<LoadedData<T>> {
+const isBoolArray = (arr: number[] | boolean[]): arr is boolean[] => typeof arr[0] === "boolean";
+
+/**
+ * Replaces all NaN in string text (such as the string representation of a JSON
+ * object) with null. Can be used to safely parse JSON objects with NaN values.
+ */
+export const nanToNull = (json: string): string => json.replace(/NaN/g, "null");
```
ShrimpCryptid, line 24: Copied from `json_utils.ts`

src/colorizer/utils/data_load_utils.ts
```diff
@@ -85,26 +87,3 @@ async function loadFromParquetUrl<T extends FeatureDataType>(url: string, type:
...
-  // Instead, transfer the underlying image and reconstruct it on the main thread.
-  const textureInfo = arrayToDataTextureInfo(data, type);
-
-  // `Transfer` is used to transfer the data buffer to the main thread without copying.
-  return new workerpool.Transfer({ min, max, data, textureInfo }, [data.buffer, textureInfo.data.buffer]);
-}
-
-workerpool.worker({
-  load: load,
-});
```
ShrimpCryptid, line 110: Moved to `data_load_utils.ts`

src/colorizer/workers/worker.ts
```diff
@@ -0,0 +1,29 @@
...
+  // Instead, transfer the underlying image and reconstruct it on the main thread.
+  const textureInfo = arrayToDataTextureInfo(data, type);
+
+  // `Transfer` is used to transfer the data buffer to the main thread without copying.
+  return new workerpool.Transfer({ min, max, data, textureInfo }, [data.buffer, textureInfo.data.buffer]);
+}
+
+workerpool.worker({
+  loadUrlData: loadUrlData,
+});
```
frasercl, line 29: optional


```suggestion
workerpool.worker({
  loadUrlData,
});
```

src/colorizer/loaders/UrlArrayLoader.ts
```diff
@@ -44,18 +39,12 @@ export class UrlArraySource<T extends FeatureDataType> implements ArraySource<T>
...
-      workerOpts: {
-        // Set worker type to undefined (classic) in production to fix a Vite issue where the application
-        // crashes when loading a module worker.
-        // Copied from https://github.com/josdejong/workerpool/tree/master/examples/vite.
-        type: import.meta.env.PROD ? undefined : "module",
-      },
-    });
+  constructor(workerPool?: SharedWorkerPool) {
+    this.cleanupWorkerPoolOnDispose = workerPool === undefined;
+    this.workerPool = workerPool ?? new SharedWorkerPool();
```
frasercl, line 47: idk if it's the right pattern for the `SharedWorkerPool` ecosystem you're looking to build, but an alternative approach worth keeping in mind for this might be reference counting the worker pool, like:

```typescript
class SharedWorkerPool {
  private workerPool: workerpool.Pool;
  private references: number;

  addReference(): void {
    this.references++;
  }

  removeReference(): void {
    this.references--;
    if (this.references <= 0) {
      this.workerPool.terminate();
    }
  }
}

class SomeSharedWorkerPoolUser {
  private workerPool: SharedWorkerPool;

  constructor(pool: SharedWorkerPool) {
    this.workerPool = pool;
    pool.addReference();
  }

  dispose(): void {
    pool.removeReference();
  }
}
```

The interface in that second class could even be made an `extends`-able base class for `SharedWorkerPool` sharers?

All optional again though. It seems just as possible from what little I know that the existing pattern, where the first class to use the workerpool effectively owns it and takes care of getting rid of it, is what makes the most sense here.

frasercl (approved):
Looks good!

src/colorizer/loaders/UrlArrayLoader.ts
ShrimpCryptid: I'll definitely keep this in mind! I think the goal for me is to have one SharedWorkerPool instance that's owned by the top-level app and is active throughout its lifetime, so it should never need to be garbage collected. I mostly have the disposal handling for testing purposes, though I don't think I have unit tests for `UrlArrayLoader` so that may be moot anyways.

rugeli (approved):

PR #474, opened 2024-11-12
fix: Outline is no longer affected by object opacity

### LINKED ISSUE #467: Inconsistent opacity behavior between track path and outline
## Description

When changing the opacity of the foreground layer, the outline fades with opacity but the track path does not.

image

## Expected Behavior

- I'm not sure if it would be better for the outline to be fully opaque or the track path to be transparent, but both should have the same behavior.

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
Closes #467. Object outlines are no longer affected by the object opacity slider, which is consistent with track paths 

Solution
========
- Updates shader slightly to ignore opacity multiplier on selected object outlines.

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/colorizer/shaders/colorize_RGBA8U.frag
+++ b/src/colorizer/shaders/colorize_RGBA8U.frag
@@ -144,7 +144,7 @@ vec4 getBackdropColor(vec2 sUv) {
   return vec4(backdropRgb, backdropColor.a);
 }
 
-vec4 getObjectColor(vec2 sUv) {
+vec4 getObjectColor(vec2 sUv, float opacity) {
   // This pixel is background if, after scaling uv, it is outside the frame
   if (isOutsideBounds(sUv)) {
     return vec4(canvasBackgroundColor, 1.0);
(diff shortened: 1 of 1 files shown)


src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -144,7 +144,7 @@ vec4 getBackdropColor(vec2 sUv) {
   return vec4(backdropRgb, backdropColor.a);
 }
 
-vec4 getObjectColor(vec2 sUv) {
+vec4 getObjectColor(vec2 sUv, float opacity) {
```
frasercl, line 147: Thought: could use the `objectOpacity` uniform rather than an argument here? Honestly don't know which is cleaner

frasercl (approved):
👍

interim17 (approved):

Issue #448, opened 2024-10-01
3D Centroids


ShrimpCryptid:
## Use Case

Requested by @Leigh-Harris during discussion of #42, XY movement delta of every cell. EMT scientists would like to be able to view direction + magnitude but also want it to include movement in the Z-dimension.

We would need to start including 3 coordinates for centroids, in the X, Y, and Z planes.

## Acceptance Criteria

- `colorizer-data` outputs 3D coordinates for centroids
- Backwards-compatibility with old format which does not include Z-axis coordinates

## Details

_Please provide any helpful specifications_

ShrimpCryptid:
Additionally, we can also consider adding a vector data type for 2D/3D vector data in `colorizer-data`.

ShrimpCryptid:
Feedback from Leigh:
- XY is a good place to start
- Would like to have an option for how many deltas to average over...
  - Special feature ("velocity") that users can override, otherwise it's calculated by the delta of centroids
  - Vector feature type that users can select. Users can visualize and scale the magnitude
- How do we get these values?
  - Leigh: Yes... we can precompute them, makes sense for users to compute them. I have not worked with TFE yet using tracks.
  - Dan: Would be okay for very simple math functions


**Insert cool whiteboard design here**

ShrimpCryptid:
![IMG_20241023_154209](https://github.com/user-attachments/assets/fbd05536-bf11-42e7-bc1d-11f3c8ddb153)

Design idea:
- Add "vector" field in the dataset manifest, which can reference a "vector feature" type. Vector features are structured like the feature data files, but have `n >= 2` dimensions for each cell ID.
- In the settings tab, add a new dropdown for "Visualize vectors" or something similar
  - Hide
  - Movement delta (auto-calculated)
    - Selecting this will show an additional input slider for amount of smoothing (avg. number of timestamps to average over)
  - one or more custom user vectors, loaded from dataset manifest
  - Also, users should be able to scale the vector magnitudes from the settings.

ShrimpCryptid:
TODO:
- [ ] Research how other vector field visualizations work in terms of coloring (by angle? by magnitude? not at all?)

ShrimpCryptid:
See #468

PR #479, opened 2024-11-20
feature: Show vector values in tooltip

### LINKED ISSUE #473: Show vector components (or angle + magnitude) in hover tooltip
## Use Case

Extension of #469. Show the vector components of the hovered object in the hover tooltip as angle + magnitude (or components) when vector arrows are enabled.

Ask scientists if they have a strong preference for one or the other? Or if we should include a toggle in the UI for both.

## Acceptance Criteria

## Details

ShrimpCryptid:
Problem
=======
Closes #473, "Show vector values in tooltip".

Solution
========
- Adds an additional line to the tooltip when vectors are toggled on that shows either the magnitude and angle or the XY components.
  - Note that units are currently always in frame pixels; future work (such as #448) may allow for custom units.
- Adds a toggle to the Vector Arrows setting area to switch between magnitude + angle or XY components.

## Type of change

Steps to Verify:
----------------
1. Open PR preview and open a dataset: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-479/
2. Turn on vector arrows from the viewer settings tab
3. Hover over any cell
4. Toggle the mode from magnitude/angle to xy components.

Screenshots (optional):
-----------------------

Magnitude + angle:
image

Components:
image


interim17 (approved):
LGTM, works when I run it :)

src/Viewer.tsx
```diff
@@ -853,6 +855,44 @@ function Viewer(): ReactElement {
...
+        </p>
+      );
+    } else {
+      // {vector}: ({x}, {y}) {unit}
+      const allowIntegerTruncation = Number.isInteger(motionDelta[0]) && Number.isInteger(motionDelta[1]);
+      hoverTooltipContent.push(
+        <p key="vector_components">
+          {vectorName}: ({numberToStringDecimal(motionDelta[0], 3, allowIntegerTruncation)},{" "}
+          {numberToStringDecimal(motionDelta[1], 3, allowIntegerTruncation)}) px
+        </p>
```
meganrm, line 891: I don't typically like to have this much data grooming mixed up with rendering. I think this would be a lot easier to follow if you had a separate function that checked for and got the appropriate data and another one that created the <p> and pushed it to the array.

meganrm (approved):
I think the code in `Viewer.tsx` could use some work, but I'm also pretty late on this review, so up to you

src/Viewer.tsx
ShrimpCryptid: Done!

Issue #484, opened 2024-12-02
Annotation labels - prototype


ShrimpCryptid:
## Use Case

First pass at #439. Implements basic UI for annotations, with no saving functionality yet!

![image](https://github.com/user-attachments/assets/6b5c6470-5440-43ee-ab72-390b3f92d504)

## Acceptance Criteria

_Please describe how you know this is done_

## Details

_Please provide any helpful specifications_

ShrimpCryptid:
Note: I'm just going to implement a first pass where only individual objects (not tracks) are annotated.

ShrimpCryptid:
![image](https://github.com/user-attachments/assets/954bb7cd-a75f-4675-88e2-a5ea5c344555)

PR #491, opened 2024-12-06
refactor: Moved hover tooltip rendering logic into component

ShrimpCryptid:
Problem
=======
Refactors the rendering logic for hover tooltips into a separate component. As part of this change, I also changed some variable names for clarity and moved component styling.

This will eventually let me add a tag tooltip that follows the mouse as part of #484!

*Estimated size: small, 10-20 minutes (faster if you compare deleted changes!!!)*

Solution
========
- Moves the rendering of tooltip contents out of `Viewer.tsx` and into a new component, `CanvasHoverTooltip.tsx`.
- Moves some styling of the tooltip out of `HoverTooltip.tsx` and into `CanvasHoverTooltip.tsx`.
- Renames some state variables in `Viewer.tsx` for clarity.

Steps to Verify:
----------------
1. Open PR preview, load a dataset, and hover over cells. https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-491/
2. Compare behavior with our public build here: https://timelapse.allencell.org


src/Viewer.tsx
```diff
@@ -203,9 +200,8 @@ function Viewer(): ReactElement {
    */
   const [frameInput, setFrameInput] = useState(0);
   const [findTrackInput, setFindTrackInput] = useState("");
-  // Prevent jarring jumps in the hover tooltip by using the last non-null value
-  const [lastHoveredId, setLastHoveredId] = useState<number | null>(null);
-  const [showHoveredId, setShowHoveredId] = useState(false);
+  const [lastValidHoveredId, setLastValidHoveredId] = useState<number>(-1);
+  const [showObjectHoverInfo, setShowObjectHoverInfo] = useState(false);
```
ShrimpCryptid, line 204: `-1` is okay here as a placeholder value, because `showObjectHoverInfo` is only set to true when `lastValidHoveredId` can be set to a valid ID number.

src/Viewer.tsx
```diff
@@ -864,55 +842,6 @@ function Viewer(): ReactElement {
...
-    } else {
-      const allowIntegerTruncation = Number.isInteger(motionDelta[0]) && Number.isInteger(motionDelta[1]);
-      const x = numberToStringDecimal(motionDelta[0], 3, allowIntegerTruncation);
-      const y = numberToStringDecimal(motionDelta[1], 3, allowIntegerTruncation);
-      return `${vectorName}: (${x}, ${y}) px
-     `;
-    }
-  };
-
-  // TODO: Move to a separate component?
```
ShrimpCryptid, line 905: :)

src/components/Tooltips/CanvasHoverTooltip.tsx
```diff
@@ -0,0 +1,112 @@
...
+import HoverTooltip from "./HoverTooltip";
+
+type CanvasHoverTooltipProps = {
+  dataset: Dataset | null;
+  featureKey: string;
+  lastValidHoveredId: number;
+  showObjectHoverInfo: boolean;
+  motionDeltas: Float32Array | null;
+  config: ViewerConfig;
+};
```
ShrimpCryptid, line 16: We could really use redux or a state management framework here...

rugeli (approved):

src/Viewer.tsx
toloudis: since you called this out, it's definitely odd to call it a "valid" hovered id when -1 is allowed!

toloudis (approved):

PR #494, opened 2024-12-07
feature: Snap scatterplot hover to currently selected track

### LINKED ISSUE #493: Difficult to hover + click current track in scatter plot
## Use Case

It's currently very difficult to hover over the values of the track you currently have selected. Also, the scatterplot tooltip changes to grey even if the value is not an outlier.

No snapping to track:
image

Grey tooltip:
image

Users would like to have snapping behavior but don't want to have to click the "clear track" button. They would be okay with a checkbox that says "Snap to current track only"

## Acceptance Criteria

## Details

I tinkered with this a bit before, but the big issue I found was that if you locked hover behavior to only the current track, there was no way to deselect a track or select another point. There are some maybe hacky solutions around this (ex: set another click listener on the plotly div, and every time a click event is received wait a few ms to see if plotly also reports a click).

ShrimpCryptid:
Problem
=======
Closes #493, "difficult to hover + click points in a track in the scatterplot"!

Solution
========
- Hover tooltips are now disabled for all points except the track when a track is selected.
- Clicking off of the track deselects the track.

## Type of change

Steps to Verify:
----------------
1. Open the preview link:
2. Switch to the scatter plot and click on a point.
3. Try clicking any points along the track.
4. Deselect the track by clicking off of it.

Screenshots (optional):
-----------------------


ascibisz (approved):
Even as a novice user, this is a much better user experience! Seems like an adventure ~fighting~ working with plotly to get this ready 😆, I appreciate all the comments. LGTM!

meganrm (approved):

PR #497, opened 2024-12-12
fix: Adds language property to 404 HTML page, adds viewer build target to package.json

ShrimpCryptid:
Tiny fix! Saw that the `lang="en"` tag was missing from our 404 HTML page.

*Estimated review size: tiny, 1 minute*

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "checks": "npm run lint & npm run typeCheck & npx vitest run",
     "build": "vite build",
     "build-internal": "vite build --base=/nucmorph-colorizer/ --config config/vite.internal-build.config.js",
-    "build-documentation": "vite build --base=/viewer/",
+    "build-colorizer-data-docs": "vite build --base=/viewer/",
     "preview": "vite preview",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
     "checks": "npm run lint & npm run typeCheck & npx vitest run",
     "build": "vite build",
     "build-internal": "vite build --base=/nucmorph-colorizer/ --config config/vite.internal-build.config.js",
+    "build-documentation": "vite build --base=/viewer/",
```
ShrimpCryptid, line 15: Also threw this in for building for the colorizer-data documentation

interim17 (approved):

toloudis (approved):

package.json
ShrimpCryptid: It's a build target for the Getting Started Guide, which contains a built copy of the viewer, stored in the subdirectory `viewer/`.

toloudis: I see - maybe there should be a comment that colorizer-data DEPENDS on this being here?

PR #499, opened 2024-12-16
feature: Annotation data model (annotations pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of 5ish for #484, annotation prototyping!

This PR implements the data model used for annotations. It syncs a pure TS class with React state using a hook to encapsulate all changes. This isn't wired up to anything yet in this PR, but will be used extensively in the next couple PRs for this feature.

Note that this change is missing track information, which means the API will change slightly in the near future.

*Estimated review size: medium, 30 minutes (line count is higher due to unit testing)*

(Also, you can see the end result of all 5 of the annotation prototype PRs here: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-498/)

Solution
========
- Adds the new `AnnotationData` class, which stores the current annotation state in the app.
  - Stores label data (name, color, and object IDs it currently labels).
  - Provides methods for interacting with label data.
- Adds the new `useAnnotations` hook in `react_utils.ts`.
  - Contains an instance of `AnnotationData`.
  - Wraps mutator methods for `AnnotationData` so state is updated accordingly.
  - Adds additional UI-based state (currently selected label, whether annotation mode is enabled, etc.)
- Adds unit tests 

## Type of change

### DIFF
--- a/src/colorizer/AnnotationData.ts
+++ b/src/colorizer/AnnotationData.ts
@@ -11,6 +11,16 @@ export type LabelData = {
 };
 
 export interface IAnnotationData {
+  /**
+   * Returns the array of label data objects.
+   * @returns an array of `LabelData` objects, containing:
+   * - `name`: The name of the label.
+   * - `color`: The color of the label.
+   * - `ids`: A set of object IDs that have the label applied.
+   * Object properties SHOULD NOT be modified directly.
+   */
+  getLabels(): LabelData[];
+
   /** Creates a new label and returns its index in the array of labels (as
    * returned by `getLabels()`).
    * @param name The name of the label. If no name is provided, a default name
@@ -21,21 +31,21 @@ export interface IAnnotationData {
   createNewLabel(name?: string, color?: Color): number;
 
   /**
-   * Returns the indices of all labels that have been applied to the provided
-   * object id. (Indices are ordered as returned by `getLabels()`.)
+   * Returns the indices of all labels that have been applied to the object id.
+   * (Indices are ordered as returned by `getLabels()`.)
    * @param id The object ID to look up label indices for.
    * @returns an array of label indices. Empty if no labels have been applied to
    * the ID.
    */
-  getLabelIdxById(id: number): number[];
+  getLabelsAppliedToId(id: number): number[];
 
   /**
-   * Returns all object IDs that have the given label (by index) applied.
+   * Returns all object IDs that the label has been applied to.
    * @param labelIdx The index of the label to look up object IDs for.
    * @returns an array of object IDs. Empty if the label has not been applied to
    * any object IDs.
    */
-  getIdsByLabelIdx(labelIdx: number): number[];
+  getLabeledIds(labelIdx: number): number[];
 
   /**
    * Returns a map from a frame number to the labeled object IDs present at that
@@ -47,25 +57,27 @@ export interface IAnnotationData {
    *
    * @param dataset The dataset to use for time information.
    * @returns a map from time to a record of label indices to IDs.
+   * @example
+   * ```typescript
+   * // Let's say we have two labels (0 and 1). There are objects with IDs 1, 2,
+   * // and 3 at time 0, and IDs 4 and 5 at time 1.
+   * // Label 0 has been applied to objects 1, 2, 3, and 4.
+   * // Label 1 has been applied to objects 3 and 5.
+   *
+   * const timeToLabelMap = getTimeToLabelIdMap(some_dataset);
+   * timeToLabelMap.get(0); // { 0: [1, 2, 3], 1: [3] }
+   * timeToLabelMap.get(1); // { 0: [4], 1: [5]}
+   * ```
    * */
-  getTimeToLabelIds(dataset: Dataset): Map<number, Record<number, number[]>>;
+  getTimeToLabelIdMap(dataset: Dataset): Map<number, Record<number, number[]>>;
 
-  /**
-   * Returns the array of labels and their associated data (name, color, labeled
-   * IDs), in index order.
-   * @returns an array of label data objects, containing:
-   * - `name`: The name of the label.
-   * - `color`: The color of the label.
-   * - `ids`: A set of object IDs that have the label applied.
-   */
-  getLabels(): LabelData[];
   setLabelName(labelIdx: number, name: string): void;
   setLabelColor(labelIdx: number, color: Color): void;
   deleteLabel(labelIdx: number): void;
 
-  toggleLabelOnId(labelIdx: number, id: number): void;
   applyLabelToId(labelIdx: number, id: number): void;
   removeLabelFromId(labelIdx: number, id: number): void;
+  toggleLabelOnId(labelIdx: number, id: number): void;
 }
 
 export class AnnotationData implements IAnnotationData {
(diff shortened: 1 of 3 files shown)


src/colorizer/utils/react_utils.ts
```diff
@@ -250,14 +254,15 @@ export const useRecentCollections = (): [RecentCollection[], (collection: Recent
...
+ * debounced motion delta array for the given dataset and vector field
+ * configuration.
  * @param dataset The dataset to calculate motion deltas for.
  * @param workerPool The worker pool to use for asynchronous calculations.
  * @param config The vector field configuration to use.
  * @param debounceMs The debounce time in milliseconds. Defaults to 100ms.
- * @returns The motion delta array or `null` if the dataset is invalid. Data will be
- * asynchronously updated as calculations complete.
+ * @returns The motion delta array or `null` if the dataset is invalid. Data
+ * will be asynchronously updated as calculations complete.
```
ShrimpCryptid, line 265: Tidied this comment slightly

src/colorizer/AnnotationData.ts
```diff
@@ -0,0 +1,222 @@
...
+   */
+  getLabeledIds(labelIdx: number): number[];
+
+  /**
+   * Returns a map from a frame number to the labeled object IDs present at that
+   * frame, represented as a record mapping label indices to IDs.
+   *
+   * The record's keys are an index of a label (as returned by `getLabels()`),
+   * and the values are arrays of object IDs present for that time frame that
+   * have that label applied.
```
ShrimpCryptid, line 56: Let me know if you can think of a tidier way to word this!

src/colorizer/AnnotationData.ts
```diff
@@ -0,0 +1,222 @@
...
+   * @example
+   * ```typescript
+   * // Let's say we have two labels (0 and 1). There are objects with IDs 1, 2,
+   * // and 3 at time 0, and IDs 4 and 5 at time 1.
+   * // Label 0 has been applied to objects 1, 2, 3, and 4.
+   * // Label 1 has been applied to objects 3 and 5.
+   *
+   * const timeToLabelMap = getTimeToLabelIdMap(some_dataset);
+   * timeToLabelMap.get(0); // { 0: [1, 2, 3], 1: [3] }
+   * timeToLabelMap.get(1); // { 0: [4], 1: [5]}
```
interim17, line 69: Nit: you could increase clarity by using 0 and 1 for the time/frame number, and a different value for the label ids.

src/colorizer/AnnotationData.ts
interim17: It's a hard thing to say tidily, wouldn't have commented if you didn't ask but I think this might be slightly clearer?? Just two cents, take it or leave it:

```
Returns a timeToLabelIdMap.
Each time (by frame number) maps to a record of labeled object 
IDs present at that time. The record's keys are the labels (by index), 
and the values are whatever objects (by ID) have that label applied 
at that time, as an array. 

A dataset with single frame, and a single label, and a single object 
that DOES have that label would return:
{0: 0: [0]} for {time: label: [objectWithLabel]}
```

interim17 (approved):
Looks great!

src/colorizer/AnnotationData.ts
ShrimpCryptid: Oh, I like this a lot! Thank you!

src/colorizer/utils/react_utils.ts
```diff
@@ -302,3 +307,103 @@ export const useMotionDeltas = (
...
+  setIsAnnotationModeEnabled: (enabled: boolean) => void;
+  visible: boolean;
+  setVisibility: (visible: boolean) => void;
+  /*
+  * Increments every time the annotation data has changed (both label metadata
+  * and IDs the labels are applied to). Can be used as a dependency in React
+  * hooks like `useEffect` or `useMemo` to trigger updates when the annotation
+  * data changes.
+  */
+  annotationDataVersion: number;
```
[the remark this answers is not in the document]
ShrimpCryptid: Hmm, maybe something like `annotationDataUpdated` would make more sense... Let me think on this a bit.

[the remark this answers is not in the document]
frasercl: Another thought, which I haven't necessarily convinced myself is better or worse than this: could the whole object returned by the hook be created in a `useMemo` with a dependency on `annotationDataVersion`, and hooks that depend on annotation data can just depend on the entire object?

I like this because it removes this odd jagged edge of the API while also slightly mitigating the fact that making this hook work requires recreating a largeish object on every render. I don't like it because it potentially replaces an explicit behavior with an implicit, not-so-obvious one. Not sure which one of those wins - I suppose I definitely think that you should `useMemo`, but am unsure whether that means you should remove `annotationDataVersion`.

src/colorizer/utils/react_utils.ts
```diff
@@ -4,9 +4,13 @@ import { useLocalStorage } from "usehooks-ts";
 
 import { VectorConfig } from "../types";
 
+import { AnnotationData, IAnnotationData } from "../AnnotationData";
 import Dataset from "../Dataset";
 import SharedWorkerPool from "../workers/SharedWorkerPool";
 
+// TODO: Move this to a folder outside of `colorizer`.
```
frasercl, line 11: +1 to this - still like the idea of keeping `colorizer` independent of React

src/colorizer/AnnotationData.ts
```diff
@@ -0,0 +1,222 @@
...
+    this.toggleLabelOnId = this.toggleLabelOnId.bind(this);
+    this.applyLabelToId = this.applyLabelToId.bind(this);
+    this.removeLabelFromId = this.removeLabelFromId.bind(this);
+  }
+
+  // Getters
+
+  getLabels(): LabelData[] {
+    return [...this.labelData];
+  }
```
frasercl, line 115: 1. Why does this copy, rather than just returning `this.labelData`?
2. Is it a problem that this shallow-copies? Modifying a member of the returned array will (if I'm not mistaken) modify the same `LabelData` object stored in `this.labelData`.

src/colorizer/AnnotationData.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+    if (!name) {
+      name = `Label ${this.numLabelsCreated + 1}`;
+    }
+
+    this.labelData.push({
+      name: name,
+      color: color,
+      ids: new Set(),
+    });
```
frasercl, line 175: Optional: can use implicit names

```suggestion
    this.labelData.push({
      name,
      color,
      ids: new Set(),
    });
```

src/colorizer/AnnotationData.ts
```diff
@@ -0,0 +1,222 @@
...
+    this.timeToLabelIdMap = null;
+  }
+
+  toggleLabelOnId(labelIdx: number, id: number): void {
+    if (!this.labelData[labelIdx].ids.has(id)) {
+      this.applyLabelToId(labelIdx, id);
+    } else {
+      this.removeLabelFromId(labelIdx, id);
+    }
+  }
```
frasercl, line 221: This is also optional, but I'm just a tiny bit hesitant about these methods, particularly `toggleLabelOnId`, whose behavior is dependent on state which the caller doesn't actually observe. It's less "set this data" and more "take this action," if that makes sense - it maps well onto what the UI does, but not necessarily onto this class's role as a data container. I might consider something like `setLabelOnId(labelIdx: number, id: number, applied: boolean)` and `getIdHasLabel(labelIdx: number, id: number): boolean`. Toggling could then be performed like `setLabelOnId(labelIdx, id, !getIdHasLabel(labelIdx, id))`. (As a bonus, this means fewer setter methods to pass through `wrapFunctionInUpdate`.)

frasercl (approved):
Liking this overall! Approving preemptively since I left it hanging over the break and because all my comments are optional/matters of opinion, and you're the expert on where this code goes from here and what it needs. A couple of those optional comments are relatively substantial though.

src/colorizer/AnnotationData.ts
ShrimpCryptid: I guess I wanted to protect the label data and make it readonly, but you're 100% right that you can still mutate the underlying objects. At the very least this prevents changing of the label ordering which would mess up all of the annotation data.

src/colorizer/utils/react_utils.ts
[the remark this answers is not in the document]
ShrimpCryptid: Talked through this with Cameron! We've moved all the getters into a memoized `data` property, which updates when the underlying data does. It can now be used directly as a dependency!

src/colorizer/AnnotationData.ts
```diff
@@ -48,39 +48,56 @@ export interface IAnnotationData {
...
-  applyLabelToId(labelIdx: number, id: number): void;
-  removeLabelFromId(labelIdx: number, id: number): void;
-  toggleLabelOnId(labelIdx: number, id: number): void;
+  setLabelOnId(labelIdx: number, id: number, value: boolean): void;
 }
 
-export class AnnotationData implements IAnnotationData {
+export type IAnnotationData = IAnnotationDataGetters & IAnnotationDataSetters;
+
+export class AnnotationData implements IAnnotationDataGetters, IAnnotationDataSetters {
```
frasercl, line 100: Just curious: can `AnnotationData` just implement `IAnnotationData` directly? If not, would defining `IAnnotationData` as an interface like...

```typescript
export interface IAnnotationData extends IAnnotationDataGetters, IAnnotationDataSetters {}
```

...make it possible?

frasercl (approved):
Looks good!

PR #500, opened 2024-12-19
refactor: Moved tab rendering logic (annotations pt. 2)

ShrimpCryptid:
Problem
=======
Part 2 of ??? for #484, prototyping annotations.

This change is a simple code move, with a little extra placeholder for the annotation tab thrown in.

*Estimated review size: small, <10 minutes*

Solution
========
- Moved tab rendering logic out of returned JSX for readability.
- Added placeholder for Annotation tab that's only visible in internal build mode.

## Type of change

Steps to Verify:
----------------
1. Open internal preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-500/
2. Compare with regular build. The annotation tab should not appear. https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-500/

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -45,7 +45,7 @@ import { getColorMap, getInRangeLUT, thresholdMatchFinder, validateThresholds } 
 import { useConstructor, useDebounce, useMotionDeltas, useRecentCollections } from "./colorizer/utils/react_utils";
 import * as urlUtils from "./colorizer/utils/url_utils";
 import { SCATTERPLOT_TIME_FEATURE } from "./components/Tabs/scatter_plot_data_utils";
-import { DEFAULT_PLAYBACK_FPS } from "./constants";
+import { DEFAULT_PLAYBACK_FPS, INTERNAL_BUILD } from "./constants";
 import { FlexRow, FlexRowAlignCenter } from "./styles/utils";
 import { LocationState } from "./types";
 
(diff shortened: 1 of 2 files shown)


src/Viewer.tsx
```diff
@@ -852,6 +852,94 @@ function Viewer(): ReactElement {
...
+            colorRampMin={colorRampMin}
+            colorRampMax={colorRampMax}
+            colorRamp={getColorMap(colorRampData, colorRampKey, colorRampReversed)}
+            categoricalPalette={categoricalPalette}
+            inRangeIds={inRangeLUT}
+            viewerConfig={config}
+            scatterPlotConfig={scatterPlotConfig}
+            updateScatterPlotConfig={updateScatterPlotConfig}
+            showAlert={showAlert}
+          />
```
rugeli, line 898: This is already clean and readable, I just have a thought for future maintainability if we anticipate adding more props: would it be less error prone to group related props into a single config object instead of passing them individually?

For instance: 
```
const scatterPlotDisplayConfig = {
dataset,
currentFrame,
...
};

const scatterPlotColorConfig = {
colorRampMin,
colorRampMax,
...
};

<ScatterPlotTab
  displayConfig={scatterPlotDisplayConfig}
...
>
```

rugeli (approved):
Just have one non-blocking comment about prop management. Looks good overall!

src/Viewer.tsx
ShrimpCryptid: Oooh, yes! We already kind of do this for the `scatterPlotConfig` and `viewerConfig`, but you're right that the color ramps etc. could be bundled together into an object as well. I'll open an issue ticket for this and update here. :')

ShrimpCryptid: Opened as https://github.com/allen-cell-animated/timelapse-colorizer/issues/502

ascibisz (approved):
LGTM!

Issue #447, opened 2024-09-27
Feature Correlation plot


toloudis:
## Use Case

We can plot a correlation matrix in the viewer and have it be clickable.  Click could take you to a scatterplot with the two features on x and y axes.

## Acceptance Criteria

TBD whether work will continue

## Details

inefficient prototype/PoC implementation here : https://github.com/allen-cell-animated/timelapse-colorizer/tree/feature/correlations

toloudis:
![Timelapse Feature Explorer 2024-09-27 14-43-08](https://github.com/user-attachments/assets/57d7223d-6bea-416e-a00b-779068900ba6)

ShrimpCryptid:
Currently blocked by needing an internal build switch!

PR #524, opened 2025-01-14
feat: Async calculation of correlations on workers (correlations pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of 2 for #447. 

This change adds a new worker task for calculating correlations between features in a dataset, based on work started by @toloudis. In part 2, this will be wired up to a new correlation plot tab.

Solution
========
- Adds a new `utils/correlations.ts` file containing logic for calculating the Pearson correlation coefficient from arrays of data.
- Adds a new handler to `SharedWorkerPool.ts` for asynchronously calculating the correlation coefficient matrix for features in a dataset.

## Type of change

### DIFF
--- a/src/colorizer/utils/correlation.ts
+++ b/src/colorizer/utils/correlation.ts
@@ -15,6 +15,8 @@ export function pearson(x: Float32Array, y: Float32Array): number {
   }
 
   const arrayLength = x.length;
+  // TODO: Could potentially optimize here by not re-initializing the arrays on
+  // every call.
   const xy = new Float32Array(arrayLength);
   const x2 = new Float32Array(arrayLength);
   const y2 = new Float32Array(arrayLength);
@@ -79,6 +81,7 @@ export function computeCorrelations(featureData: Float32Array[]): number[][] {
   // Optimization: Only calculate coefficients for [x][y] where y < x. [x][x] is
   // always 1, and [x][y] == [y][x], so we only need to calculate it once for
   // each x, y pair.
+  // TODO: Potential optimization by parallelizing this using shared memory.
   for (let colx = 0; colx < featureData.length; colx++) {
     const row = [];
     for (let coly = 0; coly < colx; coly++) {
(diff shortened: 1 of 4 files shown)


ascibisz (approved):
Looks good to me! My only suggestion would be to maybe add some tests, or make a separate ticket to add some tests? The math _looks_ right to me, but I don't completely trust myself without running it on some sample data

ShrimpCryptid:
> Looks good to me! My only suggestion would be to maybe add some tests, or make a separate ticket to add some tests? The math _looks_ right to me, but I don't completely trust myself without running it on some sample data

Added unit tests! I had to add a special case for arrays with no/minimal variance in them.

src/colorizer/utils/correlation.ts
```diff
@@ -0,0 +1,102 @@
...
+  if (x.length !== y.length) {
+    throw new Error(
+      "correlations.ts/pearson: Data arrays must have the same length when calculating correlation coefficient."
+    );
+  }
+
+  const arrayLength = x.length;
+  const xy = new Float32Array(arrayLength);
+  const x2 = new Float32Array(arrayLength);
+  const y2 = new Float32Array(arrayLength);
```
toloudis, line 20: a future optimization would be to not create these three each time you call this function - they can be reused by the higher level loop

src/colorizer/utils/correlation.ts
```diff
@@ -0,0 +1,102 @@
...
+    const row = [];
+    for (let coly = 0; coly < colx; coly++) {
+      // get list of values for colx and coly
+      const xvals = featureData[colx];
+      const yvals = featureData[coly];
+      const val = pearson(xvals, yvals);
+      row.push(val);
+    }
+    out.push(row);
+  }
```
toloudis, line 92: if this ever gets slow these nested loops are candidates for parallelization since they only read from the data arrays.  however, we'd need to use shared memory

toloudis (approved):

ascibisz (approved):

src/colorizer/utils/correlation.ts
ShrimpCryptid: Will add a TODO comment!

Issue #516, opened 2025-01-08
Annotation - add track list view


ShrimpCryptid:
## Use Case

When selecting sections of a track, the current CSV table view gets filled with hundreds of entries and is hard to navigate. Grouping track annotations together would make it easier for users to understand at a glance.

## Acceptance Criteria

- Add a radio toggle between the current table view and a list view.
- In the list view, each track that is selected will be shown as a collapsible row.
  - Expanding the track row will show a table containing only that track's objects.
- Add a visual indicator using a custom image with hard-stop gradients to denote what regions are selected. (See image)
  - Hovering over the image should display a text tooltip with a text description of the ranges selected?
  - Hover over specific sections of the image?

## Details

![image](https://github.com/user-attachments/assets/9ffc0017-f40d-4993-b584-e3e5008116c6)

Example from GitHub: 
![image](https://github.com/user-attachments/assets/020a9dc4-c294-42ad-976a-96984592bbef)

PR #529, opened 2025-01-15
refactor: Moved annotation table to a separate component

ShrimpCryptid:
Problem
=======
Refactor to support #516, "Annotation - add track list view."

This change moves the table view in the Annotation tab into its own component, which will be reused in a few places. This has been split out from the list view PR for ease of review!

Solution
========
- Moved annotation table-related logic into a new `AnnotationTable.tsx` component.

Preview here if you want to validate: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-529

### DIFF
--- a/src/components/Tabs/Annotation/AnnotationTable.tsx
+++ b/src/components/Tabs/Annotation/AnnotationTable.tsx
@@ -1,0 +1,118 @@
+import { CloseOutlined } from "@ant-design/icons";
+import { Table, TableProps } from "antd";
+import React, { ReactElement, useMemo } from "react";
+import styled from "styled-components";
+
+import { TagIconSVG } from "../../../assets";
+import { Dataset } from "../../../colorizer";
+import { FlexColumnAlignCenter, VisuallyHidden } from "../../../styles/utils";
+
+import IconButton from "../../IconButton";
+
+export type TableDataType = {
+  key: string;
+  id: number;
+  track: number;
+  time: number;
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Annotation/AnnotationTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+          </VisuallyHidden>
+        </IconButton>
+      ),
+    },
+  ];
+
+  const tableData: TableDataType[] = useMemo(() => {
+    const dataset = props.dataset;
+    if (dataset) {
+      return props.ids.map((id) => {
```
toloudis, line 83: Suggestion: make this a method on dataset, so you can do `return dataset.getAnnotationTableData(props.ids)` in order to keep the data wrangling separate from the front end ui. look for other places to do the same type of thing.

toloudis (approved):

rugeli (approved):

src/components/Tabs/Annotation/AnnotationTable.tsx
ShrimpCryptid: I started thinking about this one but it feels weird to have Dataset know about the specific use case for annotation tables? The data it's retrieving (track and time) are general enough that it seems okay to calculate with the existing API jere, but if this was doing anything more complex I would move it into a utilities file.

PR #544, opened 2025-02-05
fix: Fixed flickering issue by memoizing dropdown items

### LINKED ISSUE #540: Dropdown hover flickers during playback
## Description

When playback is occurring, hovering in a selection dropdown will cause it to rapidly switch between styling the actual hovered item and the first hovered item in the list.

## Expected Behavior

The hovered item should be the only item that is styled. Playback should not affect dropdowns.

## Reproduction

1. Go to https://timelapse.allencell.org/
2. Open a dataset
3. Start time playback
4. Open any dropdown and hover over an item (other than the first).

## Environment

ShrimpCryptid:
Problem
=======
Closes #540, "dropdown hover flickers during playback."

Solution
========
- Fixed a bug where the AnnotationData prop wouldn't update when annotation is enabled and a new label is created.
- Changed all dropdown items to be memoized to prevent a flickering bug caused by excessive re-rendering.

## Type of change

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

### DIFF
--- a/src/components/Dropdowns/SelectionDropdown.tsx
+++ b/src/components/Dropdowns/SelectionDropdown.tsx
@@ -96,11 +96,14 @@ function formatAsSelectItems(items: string[] | SelectItem[]): SelectItem[] {
  * Options can be searched by typing in the dropdown input.
  *
  * Uses react-select internally but mimics the style of Ant Design for consistency.
+ *
+ * NOTE: Items must be memoized or else flickering may occur when the page is rapidly
+ * re-rendered (e.g. during playback).
  */
 export default function SelectionDropdown(inputProps: SelectionDropdownProps): ReactElement {
   const props = { ...defaultProps, ...inputProps };
 
-  const options = formatAsSelectItems(props.items);
+  const options = useMemo(() => formatAsSelectItems(props.items), [props.items]);
 
   // TODO: Show loading spinner?
   const [_isPending, startTransition] = useTransition();
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -51,6 +51,7 @@ import {
   useRecentCollections,
 } from "./colorizer/utils/react_utils";
 import * as urlUtils from "./colorizer/utils/url_utils";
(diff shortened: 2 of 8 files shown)


src/components/Dropdowns/SelectionDropdown.tsx
```diff
@@ -96,11 +96,14 @@ function formatAsSelectItems(items: string[] | SelectItem[]): SelectItem[] {
  * Options can be searched by typing in the dropdown input.
  *
  * Uses react-select internally but mimics the style of Ant Design for consistency.
+ *
+ * NOTE: Items must be memoized or else flickering may occur when the page is rapidly
+ * re-rendered (e.g. during playback).
```
toloudis, line 101: I see things being memoized all over, and not just in this function.  IMO this comment needs to explain why memoizing fixes the problem and why we had to do it for items before they are passed in.   I accept that it DID fix it but why?

ShrimpCryptid: I'll update the comment here too with more detail. React-select does some kind of re-calculation when the `option` prop changes, which caused the flickering behavior I noted.

My main reason for wanting to move memoization out of this component instead of inside of it is for performance.
- If we do the memoization check inside of `SelectionDropdown`, we need to store a copy of the list in the component itself (O(N) add'l memory footprint) and compare both arrays on every render (O(N)). (Also, the components outside of `SelectionDropdown` may recalculate their lists each render as well, which is also O(N)).
- Doing the memoization check outside of `SelectionDropdown` means that, on renders where the list hasn't changed, we don't incur a performance cost (O(1)). 

Theoretically the second check won't be a problem for most of the UI since we have <50 options for most dropdowns, and would only really cause problems on arrays with thousands of elements. But I do think memoizing outside of the dropdown *is* best practice and I want to push anyone using this component towards that behavior.

ascibisz (approved):

toloudis (approved):

PR #558, opened 2025-02-07
refactor: Added `vole-core`, updated dependencies

ShrimpCryptid:
Ran `npm install @aics/vole-core` and `npm audit fix`. Separating this out from future PRs to prevent line count bloat.

*Estimated review size: small, <5 minutes*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "timelapse-colorizer",
       "version": "1.1.0",
       "dependencies": {
+        "@aics/vole-core": "^3.12.4",
         "@ant-design/icons": "^5.2.5",
         "@fortawesome/fontawesome-svg-core": "^6.5.1",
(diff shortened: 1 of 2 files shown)


frasercl (approved):
Looks good, though thought: should this go into a feature branch?

toloudis (approved):
2 things: 
* vole-core needs a version bump after a critical bugfix that was preventing shaders from importing correctly in any client
* I kinda agree about a feature branch, but slightly mitigated by the fact that we probably aren't doing a public release anytime soon. Also, if implemented correctly, this will incur no behavior changes on the main code paths.  The 3d view should only kick in if a special type of dataset is being loaded.

ShrimpCryptid:
> Looks good, though thought: should this go into a feature branch?

Made a new branch, `feature/3d-viewport`!

Issue #559, opened 2025-02-07
Annotations - clear annotations when switching datasets


ShrimpCryptid:
## Use Case

Annotations are currently not cleared when switching between datasets, which causes strange behavior due to tracks not having the same object IDs.

## Acceptance Criteria

- [x] When switching datasets while annotations are still present, pop up a confirmation dialog under the Datasets window. 
- [x] Clear annotation data on dataset switch.

## Details

![Image](https://github.com/user-attachments/assets/a56f664b-abd7-485d-ac23-4123ac47a20f)
See similar popup when deleting labels.

ShrimpCryptid:
Bumping this as important after some testing by Chantelle. Annotations can get really messed up when switching between datasets :(

PR #561, opened 2025-02-07
feature: Select time ranges in annotation (select range pt. 1)

### LINKED ISSUE #514: Annotation - select range of track with shortcut keys
## Use Case

Requested by @SergeEParent! Scientists want to be able to label ranges of a track

## Acceptance Criteria

- In annotation edit mode, holding shift will select a range of objects in a track from last time it was selected to the current frame.
  - The flow would work like: Select a cell in a track on a certain frame, then jump forward by some number of frames, hold shift, and click the same track again.
- A popup should appear when hovering over the new cell ("Select track {T} from frame {T0} to frame {T1}").

## Details

- Store the last object that a label was applied to (null if a label was last removed from an object).
- Check if the currently hovered object and the last object have the same track ID.

### LINKED ISSUE #531: Optimize track building in `Dataset`
## Use Case

Building the `Track` structure is comparatively slow, since the entire dataset needs to be traversed in O(N) time. Some operations (like annotation) need to request many tracks on each render.

`Dataset` should save or cache the built `Track` so it takes amortized O(1).

## Acceptance Criteria

Repeated calls to `dataset.buildTrack` should return the same instance of a Track.

## Details

ShrimpCryptid:
Problem
=======
Part 1 of 3 for #514, "Annotation - select range of track with shortcut keys." (Also closes #531, "optimize track building in Dataset".)

Adds a range annotation mode to select a range of objects within the same track. Click a track, go to a different timepoint, and click again to label (or unlabel) all of the objects in between.

I'll be following up on this PR with a change that visualizes the affected range.

*Estimated review size: medium, 20-25 minutes*

Solution
========
- Moved annotation click handling behavior out of `Viewer.tsx` and into a new callback, `handleAnnotationClick`.
- Added a new React hook, `useShortcutKey`, which tracks whether a given key is pressed down.
- Added a new range selection mode where, when annotation editing is enabled, clicking one an object in a track and then clicking another object at a different timepoint will annotate the segments of the track in between those two points.
  - Clicking again on the same object will toggle the range.
  - Range selection mode can be temporarily enabled from the time selection mode by holding shift. (This will be documented in a tooltip in select range pt. 3)
- Added a tag in the hover tooltip that appears when range selection mode is used and a range of cells will be edited.

Misc:
- Implemented a `clear()` method in `AnnotationData` to support future work in #559
- `Dataset` now caches built tracks for performance, as described in #531.

## Type of change

Steps to Verify:
----------------
1. Open PR preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-561/
2. Load a dataset and activate annotation edit mode. Change selection mode to range.
3. Click on a cell, then jump forward in time.
4. Hover over the same track. You should see a popup message saying what frames objects will be selected on.
5. Click. The whole range between the previously clicked time point and the current one should be selected.
6. Keep shift held down and click the same cell again to toggle the whole range.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/31c7f0b9-b1f1-4fec-87f2-878b08bea8ee

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -79,6 +79,7 @@ export default class Dataset {
   private timesFile?: string;
   public trackIds?: Uint32Array | null;
   public times?: Uint32Array | null;
+  private cachedTracks: Map<number, Track>;
 
   public centroidsFile?: string;
   public centroids?: Uint16Array | null;
@@ -115,6 +116,7 @@ export default class Dataset {
     this.cleanupArrayLoaderOnDispose = !arrayLoader;
     this.arrayLoader = arrayLoader || new UrlArrayLoader();
     this.features = new Map();
+    this.cachedTracks = new Map();
     this.metadata = defaultMetadata;
   }
 
@@ -550,6 +552,11 @@ export default class Dataset {
   }
 
   public buildTrack(trackId: number): Track {
+    const cachedTrack = this.cachedTracks.get(trackId);
+    if (cachedTrack) {
+      return cachedTrack;
+    }
+
     // trackIds contains a track id for every cell id in order.
     // get all cell ids for given track
     const ids = this.trackIds ? this.getIdsOfTrack(trackId) : [];
@@ -573,7 +580,9 @@ export default class Dataset {
       }, [] as number[]);
     }
 
-    return new Track(trackId, times, ids, centroids, bounds);
+    const track = new Track(trackId, times, ids, centroids, bounds);
+    this.cachedTracks.set(trackId, track);
+    return track;
   }
 
   /*
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -519,6 +519,7 @@ export default class Dataset {
     if (this.cleanupArrayLoaderOnDispose) {
       this.arrayLoader.dispose();
     }
+    this.cachedTracks.clear();
   }
 
   /** get frame index of a given cell id */
@@ -551,7 +552,7 @@ export default class Dataset {
     }, []) as number[];
   }
 
-  public buildTrack(trackId: number): Track {
+  public getTrack(trackId: number): Track {
     const cachedTrack = this.cachedTracks.get(trackId);
     if (cachedTrack) {
       return cachedTrack;
--- a/src/colorizer/utils/react_utils.ts
+++ b/src/colorizer/utils/react_utils.ts
@@ -318,13 +318,15 @@ export type AnnotationState =  {
   setVisibility: (visible: boolean) => void;
   selectionMode: AnnotationSelectionMode;
   setSelectionMode: (mode: AnnotationSelectionMode) => void;
+  handleAnnotationClick: (dataset: Dataset, id: number, selectRange?: boolean) => void;
   /**
    * Contains annotation data getters. Use this object directly as a dependency
    * in `useMemo` or `useCallback` to trigger updates when the underlying data
    * changes.
    */
   data: IAnnotationDataGetters;
-} & IAnnotationDataSetters;
+} & IAnnotationDataSetters & {
+};
 
 export const useAnnotations = (): AnnotationState => {
   const annotationData = useConstructor(() => new AnnotationData());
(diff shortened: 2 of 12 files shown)


src/colorizer/utils/react_utils.ts
```diff
@@ -382,16 +400,89 @@ export const useAnnotations = (): AnnotationState => {
...
+      if (isSelectRangeHotkeyPressed && idRange !== null) {
+        setLastEditedRange(idRange);
+        annotationData.setLabelOnIds(currentLabelIdx, idRange, !isLabeled);
+      } else {
+        annotationData.setLabelOnIds(currentLabelIdx, [id], !isLabeled);
+      }
+    }
+    setLastClickedId(id);
+    setDataUpdateCounter((value) => value + 1);
+  };
```
ShrimpCryptid, line 464: Moved logic from `Viewer.tsx` and updated with range selection behavior

src/colorizer/Dataset.ts
```diff
@@ -550,6 +552,11 @@ export default class Dataset {
   }
 
   public buildTrack(trackId: number): Track {
+    const cachedTrack = this.cachedTracks.get(trackId);
```
toloudis, line 555: this is a cool idea. Were you needing to dynamically compute tracks frequently?

src/colorizer/Dataset.ts
```diff
@@ -115,6 +116,7 @@ export default class Dataset {
     this.cleanupArrayLoaderOnDispose = !arrayLoader;
     this.arrayLoader = arrayLoader || new UrlArrayLoader();
     this.features = new Map();
+    this.cachedTracks = new Map();
```
toloudis, line 119: I'm guessing this is not explicitly cleared because you just drop the whole Dataset?

src/colorizer/utils/react_utils.ts
```diff
@@ -318,6 +319,17 @@ export type AnnotationState =  {
...
+   * For a given ID, returns the range of IDs that would be selected if the ID
+   * is clicked with range selection mode turned on.
+   * @param dataset The dataset to look up track information from.
+   * @param id The ID of an object to select with range rules applied.
+   * @returns One of the following:
+   * - Returns `null` if no range would be selected.
+   * - A list of IDs to select if this ID is clicked with range selection mode turned on.
+   */
+  getSelectRangeFromId: (dataset: Dataset, id: number) => number[] | null;
+  handleAnnotationClick: (dataset: Dataset, id: number, selectRange?: boolean) => void;
```
toloudis, line 332: Nit: Because this is a "State" class, (actually just a type? any reason why it's not a class? I'm not really sure what this thing is... in fact maybe it should be in its own file...), it might be nice to have the actual data elements of the object (like currentLabelIx, selectionMode,...) collected together and separate from the function methods.

src/colorizer/Dataset.ts
```diff
@@ -573,7 +580,9 @@ export default class Dataset {
       }, [] as number[]);
     }
 
-    return new Track(trackId, times, ids, centroids, bounds);
+    const track = new Track(trackId, times, ids, centroids, bounds);
+    this.cachedTracks.set(trackId, track);
+    return track;
```
toloudis, line 585: this function could now be called getTrack since it doesn't always have to build one

toloudis (approved):
I see a couple of simple naming things and I was a little confused about AnnotationState but generally this LGTM

src/colorizer/Dataset.ts
ShrimpCryptid: Yes! This replaces an optimization I was making in the Annotation List View, where I have to build out every track that's been annotated to create its thumbnail view. Building a track takes O(N) time where N is the number of objects in the dataset, and becomes inefficient if you need to build lots of tracks.

(Arguably we could do a single pass through the dataset on load to create all tracks at once. It would take O(Nlog(N)) since we also need to sort every track chronologically, and also would increase the memory footprint of a Dataset since we'd be storing copies of the time/track/ID arrays.)

src/colorizer/utils/react_utils.ts
ShrimpCryptid: It's just a type. It extends the actual `AnnotationData` class which holds the core annotation-related data, which knows nothing about the UI. The react hook here (`useAnnotationState`) wraps the methods of `AnnotationData` with UI-related methods and data.

ShrimpCryptid: I would like to move `useAnnotationState` into its own file at some point though... I think there's a TODO for that

src/colorizer/Dataset.ts
toloudis: I imagine that for large datasets, users rarely would touch every track, so I think the lazy version is pretty good

interim17 (approved):

PR #562, opened 2025-02-12
refactor: Made TooltipWithSubtitle into its own file

ShrimpCryptid:
Refactors a custom tooltip wrapper component into its own file, for potential use in future PRs.

*Estimated review size: small, 5 minutes*

### DIFF
--- a/src/components/Tooltips/TooltipWithSubtitle.tsx
+++ b/src/components/Tooltips/TooltipWithSubtitle.tsx
@@ -1,0 +1,36 @@
+import { Tooltip, TooltipProps } from "antd";
+import React, { ReactElement, ReactNode, useRef } from "react";
+
+/**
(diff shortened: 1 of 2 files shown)


src/components/Tooltips/TooltipWithSubtitle.tsx
```diff
@@ -0,0 +1,35 @@
...
+                {text}
+              </p>
+            ))}
+        </>
+      }
+      getPopupContainer={() => divRef.current ?? document.body}
+    >
+      {props.children}
+      {/* TODO: include invisible text here copying the tooltip? */}
+      <div ref={divRef}></div>
```
ShrimpCryptid, line 32: I did make one change from the original here where the `divRef` used to place the Tooltip in the document hierarchy is inside of the Tooltip component. This fixes a bug where radio buttons wrapped in this component would have incorrect rounding.

frasercl (approved):

Issue #492, opened 2024-12-06
Refactor state management


ShrimpCryptid:
## Use Case

State is currently a loose collection of `useState`s in the main app viewer, and it's gotten large enough that passing props around is becoming a chore (see `CanvasWrapper`). This would be a great opportunity to switch a state management framework like Zustand.

```typescript
                <CanvasWrapper
                  loading={isDatasetLoading}
                  loadingProgress={datasetLoadProgress}
                  canv={canv}
                  collection={collection || null}
                  vectorData={motionDeltas}
                  dataset={dataset}
                  datasetKey={datasetKey}
                  featureKey={featureKey}
                  selectedBackdropKey={selectedBackdropKey}
                  colorRamp={getColorMap(colorRampData, colorRampKey, colorRampReversed)}
                  colorRampMin={colorRampMin}
                  colorRampMax={colorRampMax}
                  isRecording={isRecording}
                  categoricalColors={categoricalPalette}
                  selectedTrack={selectedTrack}
                  config={config}
                  updateConfig={updateConfig}
                  onTrackClicked={onTrackClicked}
                  inRangeLUT={inRangeLUT}
                  onMouseHover={(id: number): void => {
                    const isObject = id !== BACKGROUND_ID;
                    setShowObjectHoverInfo(isObject);
                    if (isObject) {
                      setLastValidHoveredId(id);
                    }
                  }}
                  onMouseLeave={() => setShowObjectHoverInfo(false)}
                  showAlert={isInitialDatasetLoaded ? showAlert : undefined}
                />
```

## Acceptance Criteria

- [x] Store Dataset and Collection as a Slice
- [x] Backdrop data as a slice
- [x] Color ramp slice
- [x] Feature key slice
- [x] Thresholds slice
- [x] Vector data slice
- [x] Track slice
- [x] Recording/playback state as slice?
- [x] Convert remaining ViewerConfig variables into slice state (scatter plot variables, draw modes, track path, scale bar, timestamp enabled)
- [x] Update URL parsing + sync to be part of slice
- [ ] Annotation data slice?
- [x] Move initial URL parsing/state handling into store
- [ ] Move `replaceDataset` logic into store

## Details
- This will likely need to be split up into multiple PRs.

ShrimpCryptid:
Lowest-lift version of this is even to make a shared viewer state object.

PR #572, opened 2025-02-18
refactor: Initial shared viewer state (state pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of ??? for #492, "Refactor state management."

This change adds Zustand, a hook-based state management library. The long-term plan is to move a lot of the state management behavior that's currently in `Viewer.tsx` into separate files that will be easier to parse and understand.

**This change is a little demonstration of how the new store system works, and I'm looking for feedback on the pattern.** 

This will be part of a LONG chain of PRs that slowly move more and more code logic into the store. There is also an immediate follow-up PR to this one that adds **unit testing**, #573!

Solution
========
- Adds Zustand as a dependency to the project.
- Adds example slices for the dataset, collection, and backdrop-related data.
  - The selected backdrop key can only ever be a valid key in `dataset`, and otherwise is set to a default key or `null`.
- Added the new `useViewerStateStore()` hook and used it in a few places (`CanvasWrapper.tsx`, `SettingsTab.tsx`) to reduce the required props. 

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-572/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Medium&feature=volume&t=15&bg-key=max_intensity_zprojection_of_lamin_b1&scatter-range=all
3. Switch between datasets in the collection
4. Click the Load button and load in a new dataset: https://raw.githubusercontent.com/allen-cell-animated/colorizer-data/refs/heads/main/documentation/example_dataset/manifest.json
5. The backdrop key should switch to the new dataset's default. From the settings tab you should be able to switch between backdrops.

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -62,7 +62,6 @@ import Collection from "./colorizer/Collection";
 import { BACKGROUND_ID } from "./colorizer/ColorizeCanvas";
 import { FeatureType } from "./colorizer/Dataset";
 import UrlArrayLoader from "./colorizer/loaders/UrlArrayLoader";
-import { useViewerStateStore } from "./colorizer/state/ViewerState";
 import TimeControls from "./colorizer/TimeControls";
 import SharedWorkerPool from "./colorizer/workers/SharedWorkerPool";
 import { AppThemeContext } from "./components/AppStyle";
@@ -91,6 +90,7 @@ import {
   SettingsTab,
 } from "./components/Tabs";
 import CanvasHoverTooltip from "./components/Tooltips/CanvasHoverTooltip";
+import { useViewerStateStore } from "./state/ViewerState";
 
 // TODO: Refactor with styled-components
 import styles from "./Viewer.module.css";
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -62,6 +62,7 @@ import Collection from "./colorizer/Collection";
 import { BACKGROUND_ID } from "./colorizer/ColorizeCanvas";
 import { FeatureType } from "./colorizer/Dataset";
 import UrlArrayLoader from "./colorizer/loaders/UrlArrayLoader";
+import { useViewerStateStore } from "./colorizer/state/ViewerState";
 import TimeControls from "./colorizer/TimeControls";
 import SharedWorkerPool from "./colorizer/workers/SharedWorkerPool";
 import { AppThemeContext } from "./components/AppStyle";
@@ -107,9 +108,18 @@ function Viewer(): ReactElement {
     return canvas;
   });
 
-  const [collection, setCollection] = useState<Collection | undefined>();
-  const [dataset, setDataset] = useState<Dataset | null>(null);
-  const [datasetKey, setDatasetKey] = useState("");
+  // TODO: Merge these into a single state slice call?
+  const collection = useViewerStateStore((state) => state.collection);
+  const setCollection = useViewerStateStore((state) => state.setCollection);
+  const dataset = useViewerStateStore((state) => state.dataset);
+  const datasetKey = useViewerStateStore((state) => state.datasetKey);
+  const setDataset = useViewerStateStore((state) => state.setDataset);
+  // TODO: Remove these when URL parameter initialization and updating is moved
+  // into a helper method/out of the component
+  /** Backdrop key is null if the dataset has no backdrops, or during initialization. */
+  const selectedBackdropKey = useViewerStateStore((state) => state.backdropKey);
+  const setSelectedBackdropKey = useViewerStateStore((state) => state.setBackdropKey);
+
   const [, addRecentCollection] = useRecentCollections();
 
   // Shared worker pool for background operations (e.g. loading data)
@@ -119,21 +129,7 @@ function Viewer(): ReactElement {
   const [featureKey, setFeatureKey] = useState("");
   const [selectedTrack, setSelectedTrack] = useState<Track | null>(null);
   const [currentFrame, setCurrentFrame] = useState<number>(0);
-  /** Backdrop key is null if the dataset has no backdrops, or during initialization. */
-  const [selectedBackdropKey, setSelectedBackdropKey] = useState<string | null>(null);
   const annotationState = useAnnotations();
-
-  useEffect(() => {
-    // Switch to default backdrop if the dataset has one and none is currently selected.
-    // If the dataset has no backdrops, hide the backdrop.
-    if (dataset && (selectedBackdropKey === null || !dataset.hasBackdrop(selectedBackdropKey))) {
-      const defaultBackdropKey = dataset.getDefaultBackdropKey();
-      setSelectedBackdropKey(defaultBackdropKey);
-      if (!defaultBackdropKey) {
-        updateConfig({ backdropVisible: false });
-      }
-    }
-  }, [dataset, selectedBackdropKey]);
 
   // TODO: Save these settings in local storage
   // Use reducer here in case multiple updates happen simultaneously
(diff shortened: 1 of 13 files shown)


src/colorizer/state/slices/backdrop_slice.ts
```diff
@@ -0,0 +1,61 @@
...
+   * is no Dataset loaded or if the dataset does not have backdrops. */
+  backdropKey: string | null;
+  backdropVisible: boolean;
+  /** Brightness, as a percentage in the `[0, 200]` range. 100 by default. */
+  backdropBrightness: number;
+  /** Saturation, as a percentage in the `[0, 100]` range. */
+  backdropSaturation: number;
+  /** Object opacity when backdrop is visible, as a percentage in the `[0, 100]`
+   * range. 50 by default. */
+  objectOpacity: number;
```
ShrimpCryptid, line 17: Note that these are not wired up to the UI yet.

src/Viewer.tsx
```diff
@@ -107,9 +108,18 @@ function Viewer(): ReactElement {
     return canvas;
   });
 
-  const [collection, setCollection] = useState<Collection | undefined>();
-  const [dataset, setDataset] = useState<Dataset | null>(null);
-  const [datasetKey, setDatasetKey] = useState("");
+  // TODO: Merge these into a single state slice call?
+  const collection = useViewerStateStore((state) => state.collection);
+  const setCollection = useViewerStateStore((state) => state.setCollection);
```
toloudis, line 113: why is the setter a separate state thing? is it an action and if so, do all state variables and actions get treated the same in this way?

src/colorizer/state/slices/backdrop_slice.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const createBackdropSlice: StateCreator<BackdropSlice & DatasetSlice, [], [], BackdropSlice> = (set, get) => ({
+  backdropKey: null,
+  backdropVisible: false,
+  backdropBrightness: 100,
+  backdropSaturation: 100,
+  objectOpacity: 50,
+
+  setBackdropKey: (key: string | null) => {
+    const dataset = get().dataset;
```
toloudis, line 46: Maybe my lack of understanding, but it feels weird to me that backdropslice somehow includes datasetslice.  This feels like a dependency that shouldn't be there -- like maybe setBackdropKey should take a dataset as an arg?  how does this work when one state slice has to bring in another for the actions?  From a code organization standpoint, this looks like a good reason to separate the actions (which might depend on other bits of state) from the data members.

toloudis (approved):
Are you making sure that funcitonality is not impacted on each of these PRs?  There is a lot of functionality to to test.  If you intend to at all, I suggest bringing in unit testing sooner than later for the state slices, actions etc.

src/colorizer/state/slices/backdrop_slice.ts
ShrimpCryptid: That's a really good suggestion, thank you! Yeah, requiring dependencies to be passed in makes a lot of sense for the simpler state (e.g. stuff that's not the dataset/collection) and makes it easier to test separately.

> how does this work when one state slice has to bring in another for the actions?
In the creator function for the slice, I can specify other slices (data members + actions) that the slice is dependent on. The `DatasetSlice` has an example of this where the creator function is consuming the `BackdropSlice`. I agree that it would be ideal to make `BackdropSlice` not have a dependency on `DatasetSlice` though.

> From a code organization standpoint, this looks like a good reason to separate the actions (which might depend on other bits of state) from the data members.
I agree, though for low-level setters I want to keep them near the data members they're changing. There will be a couple higher-level selectors/actions that will probably live in their own slices? E.g. initializing from URL and saving to URL

ShrimpCryptid:
> Are you making sure that funcitonality is not impacted on each of these PRs? There is a lot of functionality to to test. If you intend to at all, I suggest bringing in unit testing sooner than later for the state slices, actions etc.

Yup! Unit testing is coming soon in the next PR.

src/Viewer.tsx
ShrimpCryptid: Yes, it's an action. All actions and state variables are pulled out of the store with this pattern currently!

src/colorizer/state/slices/dataset_slice.ts
```diff
@@ -0,0 +1,40 @@
...
+  dataset: null,
+
+  setDataset: (key: string, dataset: Dataset) => {
+    ///// Validate dataset-dependent state values /////
+
+    // Switch to new dataset's default backdrop if the current one is not in the new dataset.
+    let backdropKey = get().backdropKey;
+    if (backdropKey === null || !dataset.hasBackdrop(backdropKey)) {
+      backdropKey = dataset.getDefaultBackdropKey();
+    }
```
frasercl, line 34: So this is reaching across slices? I remember this being a huge problem for redux, nice that it's easy here

src/colorizer/state/slices/collection_slice.ts
```diff
@@ -0,0 +1,27 @@
...
+type CollectionSliceActions = {
+  setCollection: (collection: Collection) => void;
+  // TODO: Add action for loading and setting a dataset here
+};
+
+export type CollectionSlice = CollectionSliceState & CollectionSliceActions;
+
+export const createCollectionSlice: StateCreator<CollectionSlice & DatasetSlice, [], [], CollectionSlice> = (
+  set,
+  _get
```
frasercl, line 20: Do you need to include the explicit unused argument here or can you just leave it off?

frasercl (approved):
I like it! Many clarifying questions and much thinking aloud below - the only gripe I might have is the placement of state organization under the "colorizer" directory, but that doesn't need to block things.

src/colorizer/state/ViewerState.ts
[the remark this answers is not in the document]
ShrimpCryptid: Oh hm, that's fair! I've been treating the `colorizer` folder as a `lib` folder at this point, but it's easy enough to move these.

src/Viewer.tsx
```diff
@@ -107,9 +108,18 @@ function Viewer(): ReactElement {
     return canvas;
   });
 
-  const [collection, setCollection] = useState<Collection | undefined>();
-  const [dataset, setDataset] = useState<Dataset | null>(null);
-  const [datasetKey, setDatasetKey] = useState("");
+  // TODO: Merge these into a single state slice call?
+  const collection = useViewerStateStore((state) => state.collection);
```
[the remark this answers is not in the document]
ShrimpCryptid: Yes! I think it'll be useful to have these all split out for now, since it'll let me know what variables are unused once I start removing data logic.

src/colorizer/state/slices/dataset_slice.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+type DatasetSliceActions = {
+  setDataset: (key: string, dataset: Dataset) => void;
+  clearDataset: () => void;
+};
+
+export type DatasetSlice = DatasetSliceState & DatasetSliceActions;
+
+export const createDatasetSlice: StateCreator<CollectionSlice & DatasetSlice & BackdropSlice, [], [], DatasetSlice> = (
```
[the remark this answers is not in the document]
ShrimpCryptid: I could certainly use `ViewerState`, yes! It might be cleaner that way too if any of the other slices change.

Right now this is a minimal version of the dependent slices.

src/colorizer/state/slices/dataset_slice.ts
ShrimpCryptid: Yup, this is the DatasetSlice accessing values defined by BackdropSlice.

src/colorizer/state/slices/collection_slice.ts
ShrimpCryptid: I can leave it off!

ShrimpCryptid: Mm, decided to leave it because I expect this slice to need it in the near future

PR #573, opened 2025-02-18
feat: Added unit testing for useViewerStateStore hook (state pt. 2)

ShrimpCryptid:
Problem
=======
Part 2 of ???? for #492, "Refactor state management." Builds off of #572 by adding unit testing for `useViewerStateStore`! 

*Estimated review size: medium, 20-30 minutes (most of the line count is from package-lock)*

Solution
========
- Adds a Zustand mock which resets the store between each unit test.
- Added unit tests for the dataset slice and backdrop slice.
- Fixed a bug where backdrop visibility would remain enabled even when a new dataset had no backdrops.

## Type of change


src/colorizer/state/slices/dataset_slice.ts
```diff
@@ -32,9 +32,10 @@ export const createDatasetSlice: StateCreator<CollectionSlice & DatasetSlice & B
...
       backdropKey = dataset.getDefaultBackdropKey();
     }
+    let backdropVisible = get().backdropVisible && backdropKey !== null;
 
     // TODO: Dispose of old dataset?
-    set({ datasetKey: key, dataset, backdropKey });
+    set({ datasetKey: key, dataset, backdropKey, backdropVisible });
   },
-  clearDataset: () => set({ datasetKey: null, dataset: null, backdropKey: null }),
+  clearDataset: () => set({ datasetKey: null, dataset: null, backdropKey: null, backdropVisible: false }),
```
ShrimpCryptid, line 40: Fix. This disables the backdrop visibility if the incoming dataset has no backdrops.

ShrimpCryptid: - [ ] TODO: Test for this in the dataset_slice.ts unit tests

tests/colorizer/ViewerStore/backdrop_slice.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      expect(result.current.backdropBrightness).toBe(75);
+      act(() => {
+        result.current.setBackdropBrightness(220);
+      });
+      expect(result.current.backdropBrightness).toBe(200);
+      act(() => {
+        result.current.setBackdropBrightness(-10);
+      });
+      expect(result.current.backdropBrightness).toBe(0);
+    });
```
ShrimpCryptid, line 97: Are these overkill?

src/colorizer/state/ViewerState.ts
ShrimpCryptid, line 1: Linter disliked double-underscore

tests/state/ViewerState/backdrop_slice.test.ts
```diff
@@ -0,0 +1,123 @@
+import { act, renderHook } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import { Dataset } from "../../../src/colorizer";
+import { ANY_ERROR } from "../../test_utils";
+import { MOCK_DATASET_WITHOUT_BACKDROP } from "./constants";
+
+import { useViewerStateStore } from "../../../src/state/ViewerState";
```
toloudis, line 8: group imports from same dir with line above

tests/state/ViewerState/backdrop_slice.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+describe("useViewerStateStore: BackdropSlice", () => {
+  describe("setBackdropKey", () => {
+    it("does not allow backdrop slice to be set when provided Dataset has no backdrops", () => {
+      const { result } = renderHook(() => useViewerStateStore());
```
toloudis, line 13: for my understanding, does this line just get the entire app state store?

tests/state/ViewerState/backdrop_slice.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      // Initialized as null
+      expect(result.current.backdropKey).toBeNull();
+
+      act(() => {
+        result.current.setBackdropKey(MOCK_DATASET_WITHOUT_BACKDROP, "test");
```
toloudis, line 19: why does this need to be wrapped in "act"? (I mean would you get the same result by calling it directly?)

toloudis: also, it seems to silently fail and just leave things as null.  is that basically the intent?  I'm not totally sure what is being tested here.

tests/state/ViewerState/backdrop_slice.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    expect(result.current.backdropBrightness).toBe(75);
+    expect(result.current.backdropSaturation).toBe(50);
+    expect(result.current.objectOpacity).toBe(25);
+
+    act(() => {
+      result.current.setBackdropBrightness(220);
+      result.current.setBackdropSaturation(150);
+      result.current.setObjectOpacity(120);
+    });
+    expect(result.current.backdropBrightness).toBe(200);
```
toloudis, line 97: Nit: these min and max values should probably be constants somewhere, maybe?   This is part of that data validation thing we discussed where I'm not sure the actions should be validating the inputs.  I wonder if there is a industry consensus on this.

 I suppose maybe clamping is better than not clamping.  But then below I see you throw for NaNs, but clamp for out of range values?

src/state/slices/backdrop_slice.ts
```diff
@@ -27,13 +27,35 @@ type BackdropSliceActions = {
...
+   * Sets the opacity of objects when the backdrop is visible, clamped to a
+   * percentage in the `[0, 100]` range.
+   * @throws {Error} If the opacity is `NaN`.
+   */
   setObjectOpacity: (opacity: number) => void;
 };
 
 export type BackdropSlice = BackdropSliceState & BackdropSliceActions;
 
+const clampPercentage = (value: number, min: number, max: number): number => {
```
toloudis, line 52: maybe name this `clampWithNaNCheck` and move it into a shareable utility module?

src/state/slices/backdrop_slice.ts
```diff
@@ -27,13 +27,35 @@ type BackdropSliceActions = {
...
+  if (Number.isNaN(value)) {
+    throw new Error(`Invalid percentage value: ${value}`);
```
toloudis, line 54:
```suggestion
    throw new Error(`Invalid value to clamp: ${value}`);
```

toloudis (approved):
A few questions...basically looks ok.  Does it feel like the tests are adding good value?

ShrimpCryptid:
> A few questions...basically looks ok. Does it feel like the tests are adding good value?

Yes! The tests helped me find some bugs in the code OOPS

tests/state/ViewerState/backdrop_slice.test.ts
ShrimpCryptid: Mentioned this above! For the react testing library, you have to wrap state changes in `act` to allow renders to occur.

ShrimpCryptid: And yes, this currently ignores backdrop keys that aren't in the specified dataset. I guess I should throw an explicit error here instead, since that would indicate a frontend bug.

tests/state/ViewerState/backdrop_slice.test.ts
ShrimpCryptid: I read some discussions on this! The general sense I got was that you should be validating inputs at multiple levels. It's good to do it from the UI so that users can get immediate feedback, but bad data passed in your state model usually indicates a developer error and should throw errors/warnings to help you catch it.

Agree on making constants for the min/max! I could probably make out-of-range values invalid as well.

tests/state/ViewerState/backdrop_slice.test.ts
ShrimpCryptid: Fixed! Had to change a rule in `.prettierrc`.

ShrimpCryptid: Ah, turns out that they do this because they want to split out class/component imports. I should probably rename `ViewerState.ts`

frasercl (approved):
Looks great! A tiny bit confused by the mocks setup, but looks like it's coming pretty directly off zustand's docs anyways.

PR #576, opened 2025-02-25
refactor: Move color ramp data into shared viewer state (state pt. 3)

ShrimpCryptid:
Problem
=======
Part 3 of ??? for #492, "refactor state management."

This change moves the color ramp into the shared viewer store, and also introduces a new pattern for derived state values using Zustand's subscriber model.

Solution
========
- Moved color ramp state (key, categorical palette, min and max range, reversed) into the `useViewerStateStore` hook.
  - Added subscribers that update the calculated color ramp and current categorical palette key when their dependencies change.
- Removed props for color ramp-related data from some components.
- Added unit tests for color ramp behavior
- Bumped `typescript` to `v5.0.4` to add support for the `const` keyword in literal expressions

## Type of change

Steps to Verify:
----------------
1. Open https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-576/viewer and load a dataset.
2. Change the color ramp by moving the sliders, choosing a different ramp, or reversing it. 
3. Change to any categorical feature (like "trajectory termination annotation").
4. Open the scatter plot tab. It should update when the ramp or palette changes.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/f5ec85d2-254f-4cdb-acd0-a0a70bad14c6


src/colorizer/state/ViewerState.ts
```diff
@@ -47,7 +50,7 @@ export const viewerStateStoreCreator: StateCreator<ViewerState> = (...a) => ({
  * console.log(store.dataset);
  * ```
  */
-export const useViewerStateStore = create<ViewerState>()(viewerStateStoreCreator);
+export const useViewerStateStore = create<ViewerState>()(subscribeWithSelector(viewerStateStoreCreator));
```
ShrimpCryptid, line 53: This allows subscribers to changes in the store to add a selector function to select a subset of state. This allows those subscribers to only be called when those dependencies change.

src/colorizer/state/types.ts
```diff
@@ -0,0 +1,24 @@
+import { StoreApi, UseBoundStore } from "zustand";
+
+// Copied from Zustand's types
```
ShrimpCryptid, line 3: This is probably a Typescript crime, don't tell @frasercl

src/colorizer/state/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    <U>(
+      selector: (state: T) => U,
+      listener: (selectedState: U, previousSelectedState: U) => void,
+      options?: {
+        equalityFn?: (a: U, b: U) => boolean;
+        fireImmediately?: boolean;
+      }
+    ): () => void;
+  };
+};
```
ShrimpCryptid, line 16: This is probably a Typescript crime, don't tell @frasercl

src/colorizer/state/slices/color_ramp_slice.ts
```diff
@@ -0,0 +1,121 @@
...
+export const createColorRampSlice: StateCreator<ColorRampSlice> = (set, _get) => ({
+  // State
+  colorRampKey: DEFAULT_COLOR_RAMP_KEY,
+  isColorRampReversed: false,
+  colorRampRange: [0, 0],
+  categoricalPalette: KNOWN_CATEGORICAL_PALETTES.get(DEFAULT_CATEGORICAL_PALETTE_KEY)!.colors,
+
+  // Derived state
+  colorRamp: getColorMap(KNOWN_COLOR_RAMPS, DEFAULT_COLOR_RAMP_KEY, false),
+  categoricalPaletteKey: DEFAULT_CATEGORICAL_PALETTE_KEY,
```
ShrimpCryptid, line 58: I went through a few iterations on how to handle derived/calculated state.

1. The first approach was having the setters for the dependencies update the derived value (so when you call `setColorRampKey` it also changes `colorRamp`, but this seemed bug-prone.
2. I tried using a `computed` method suggested in the Zustand GitHub discussions, where the derived value is calculated using a selector that returns a new value if the dependencies change. This didn't work because there was no state update to trigger rerenders.
3. I tried some `subscribe` middleware but couldn't get the typing to work. It also would have made the slice declarations pretty confusing.

The final method I landed on still uses the `subscribe` pattern in Zustand. When the store is created, listeners are added that immediately update derived values when their dependencies change. This seems like the solution that required the least changes and would be easiest to read.

src/colorizer/state/types.ts
ShrimpCryptid: TODO: Open issue ticket in Zustand repo about this

package.json
```diff
@@ -42,7 +42,7 @@
     "jsdom": "^22.1.0",
     "prettier": "^2.8.7",
     "semver": "^7.6.0",
-    "typescript": "^4.9.5",
+    "typescript": "<5.1.0",
```
toloudis, line 45: why this pin? this seems bad at first glance

src/colorizer/utils/data_utils.ts
```diff
@@ -5,6 +5,19 @@ import { FeatureThreshold, isThresholdCategorical, isThresholdNumeric, Threshold
 import ColorRamp from "../ColorRamp";
 import Dataset, { FeatureType } from "../Dataset";
 
+/** Returns whether the two arrays are deeply equal, where arr1[i] === arr2[i] for all i. */
+export function arrayDeepEquals<T>(arr1: T[], arr2: T[]): boolean {
```
toloudis, line 9: nit: tiny misnomer since this is only one layer deep and not all-the-way deep.  IDK if I have a better name in mind, maybe arrayElementsAreEqual, or maybe this name is ok...

src/components/Tabs/ScatterPlotTab.tsx
```diff
@@ -86,6 +83,20 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
+  const store = useViewerStateStore(
+    useShallow((state) => ({
+      dataset: state.dataset,
+      colorRampRange: state.colorRampRange,
+      colorRamp: state.colorRamp,
+      categoricalPalette: state.categoricalPalette,
+    }))
+  );
+
+  const dataset = useDebounce(store.dataset, 500);
```
toloudis, line 95: I know this is not part of the changes, but why did you choose the different values for the debounces?

src/state/ViewerState.ts
```diff
@@ -45,9 +48,19 @@ export const viewerStateStoreCreator: StateCreator<ViewerState> = (...a) => ({
...
  * const store = useViewerStateStore();
  * console.log(store.dataset);
+ *
+ * // Subscribing to changes in the store:
+ * useViewerStateStore.subscribe(
+ *   (state) => [state.dataset, state.collection],
+ *   ([dataset, collection]): void => {
+ *    console.log("Dataset or collection changed:", dataset, collection);
+ *   }
+ * );
```
toloudis, line 58: very helpful comment showing how this is intended to be used

src/state/slices/color_ramp_slice.ts
toloudis, line 1: why does this file use a different naming convention?  Is this distinguishing react components vs not?

toloudis (approved):
it's starting to look more and more reduxish.  I like how the changes seem to be simplifying the main app components.

src/components/Tabs/ScatterPlotTab.tsx
ShrimpCryptid: It's a little arbitrary, but `Dataset` has a longer debounce timer so that the scatterplot isn't trying to re-render right as we're loading the new Dataset and the rest of the UI is updating too.

src/state/slices/color_ramp_slice.ts
ShrimpCryptid: I did this to match the convention we're using in the `colorizer` directory for library-like utility files... currently this is the way we name any file that's not a class or a component, though it's not 100% consistent.

src/colorizer/utils/data_utils.ts
ShrimpCryptid: `arrayElementsAreEqual` makes sense to me. I'll make the change 👍

package.json
ShrimpCryptid: I changed it because I needed Typescript 5.x.x but also was having a compatibility warning show up when running linting. Let me poke at this a bit.

ascibisz (approved):

package.json
ShrimpCryptid: Changed this, we'll now get a warning on linting but I think that's fine.

PR #577, opened 2025-02-26
fix: Fixed bug where time playback could not be stopped

### LINKED ISSUE #575: Cannot pause playback when frame loading is slow
## Description

Sometimes when backdrops are enabled or the network is slow, users are unable to pause a playing timelapse.
This likely broke as a consequence of a fix implemented in #426.

## Expected Behavior

- Playback should pause when the pause button is clicked.

## Reproduction

1. Open TFE in an incognito window.
2. Load a dataset and enable the backdrop.
3. Hit play.
4. Try to pause playback.

## Environment

ShrimpCryptid:
Problem
=======
Closes #575, "cannot stop playback when frame loading is slow."

Solution
========
- Adds a check to prevent the time slider from setting the current time during playback.
- Fixes a behavior where playback would start/resume anytime the mouse was clicked and released.
- Syncs the time slider and the current frame when time playback is paused.

## Type of change

Steps to Verify:
----------------
1. Open PR preview and load a dataset: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-577/
2. Turn on backdrops.
3. Start playback, pause playback, jump through time, click and drag the slider during time playback, etc.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/24bd71e3-14dd-4999-b424-5b2c882ea607

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -335,7 +335,7 @@ function Viewer(): ReactElement {
     }
   }, [timeControls.isPlaying(), isRecording, getUrlParams, isInitialDatasetLoaded]);
 
-  // Callback for setting time, to be used only with timeControls.
+  // Callback for setting time, to be used directly only with timeControls.
   const setFrameCallback = useCallback(
     async (frame: number) => {
       await canv.setFrame(frame);
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -220,6 +220,7 @@ function Viewer(): ReactElement {
     }
   }, [timeControls.isPlaying()]);
 
+  const timeSliderContainerRef = useRef<HTMLDivElement>(null);
   /** The frame selected by the time UI. Changes to frameInput are reflected in
    * canvas after a short delay.
    */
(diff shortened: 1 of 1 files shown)


interim17 (approved):

src/Viewer.tsx
```diff
@@ -832,17 +833,27 @@ function Viewer(): ReactElement {
   // the frame using a debounced value to prevent constant updates as it moves.
   const debouncedFrameInput = useDebounce(frameInput, 250);
   useEffect(() => {
-    setFrame(debouncedFrameInput);
+    if (!timeControls.isPlaying() && currentFrame !== debouncedFrameInput) {
+      setFrame(debouncedFrameInput);
+    }
```
frasercl, line 838: clarifying q: the time slider gaining focus will pause playback, but this `if` protects against this effect still firing once or twice before that happens?

frasercl (approved):

src/Viewer.tsx
ShrimpCryptid: I found it useful to prevent a stuttering I was seeing, where if playback is enabled and a frame takes more than 250 ms to load, `debouncedFrameInput` would update and attempt to set the frame to the current one, which could cause a race condition with the next frame being loaded. `setFrame` will temporarily pause playback and resume it once the frame is loaded, which led to a lot of fighting as `timeControls` was rapidly paused/unpaused.

PR #582, opened 2025-03-04
refactor: Moved vector config into viewer state (state pt. 6)

ShrimpCryptid:
Problem
=======
Part 6 of ??? for #492, "refactor state management". This change moves the vector configuration into the viewer state store.

Note that this will temporarily break syncing of the vector state with URLs (you can save configuration to the URL but not load it from the URL).

*Estimated review size: large, 30-40 minutes*

Solution
========
- Added a shared handle for the SharedWorkerPool via the new `WorkerPoolSlice`, allowing worker-related async tasks to be executed by other actions in the store.
- Added a new utility method, `makeDebouncedCallback`, to delay updating derived state/listener callbacks that may require expensive operations.
- Added the new `VectorSlice` for handling vector visualization-related state.
  - Calculation for motion deltas is now handled by the store.
- Updates components to retrieve vector state from the shared store.

## Type of change

Steps to Verify:
----------------
1. Open PR preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-582/
2. Turn on vectors. 
3. Change vector settings. The onscreen vectors should still update properly in response.
4. Change the time interval for motion deltas. The vectors should update after a short (debounced) delay.

Screenshots (optional):
-----------------------

**Video Talk-through (🔊):**

https://github.com/user-attachments/assets/07fb4aec-6dad-420d-aa8b-fc0117edc87b


src/colorizer/workers/worker.ts
```diff
@@ -34,10 +34,10 @@ async function getMotionDeltas(
   trackIds: Uint32Array,
   times: Uint32Array,
   centroids: Uint16Array,
-  config: VectorConfig
+  timeIntervals: number
```
ShrimpCryptid, line 37: The only value in the config that was being used was the time interval, so I ended up changing this argument to be the time interval only.

src/state/slices/workerpool_slice.ts
```diff
@@ -0,0 +1,11 @@
+import { StateCreator } from "zustand";
+
+import SharedWorkerPool from "../../colorizer/workers/SharedWorkerPool";
+
+export type WorkerPoolSlice = {
+  workerPool: SharedWorkerPool;
+};
+
+export const createWorkerPoolSlice: StateCreator<WorkerPoolSlice, [], [], WorkerPoolSlice> = (_set, _get) => ({
+  workerPool: new SharedWorkerPool(),
```
toloudis, line 10: I'm totally confused as to how this is "state".  
Is this going here because it's a shared module that a lot of other components need to access?  
How do you decide what to put in state slices?

toloudis: to offer an alternative maybe there could just be a single global sharedworkerpool instance that lives in the workerpool module itself.  You might have to control how/when it's initialized, of course, but I think it doesn't really depend on anything else at creation time...  Just a thought, I might be totally of base :)

tests/state/ViewerState/vector_slice.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        result.current.setVectorMotionTimeIntervals(-1);
+      });
+      expect(result.current.vectorMotionTimeIntervals).toBe(1);
+      act(() => {
+        result.current.setVectorMotionTimeIntervals(-100);
+      });
+      expect(result.current.vectorMotionTimeIntervals).toBe(1);
+    });
+
+    it("round to integers", () => {
```
ShrimpCryptid, line 68:
```suggestion
    it("rounds to integers", () => {
```

src/state/ViewerState.ts
```diff
@@ -8,26 +8,34 @@ import { CollectionSlice, createCollectionSlice } from "./slices/collection_slic
...
+  ...createDatasetSlice(...a),
   ...createThresholdSlice(...a),
+  ...createVectorSlice(...a),
 });
 
 /**
  * Hook for accessing the global viewer state store. If used with selectors,
  * components will only rerender when the selected state changes.
  *
+ * NOTE: If you are experiencing a re-render loop while selecting multiple'
```
ShrimpCryptid, line 35:
```suggestion
 * NOTE: If you are experiencing a re-render loop while selecting multiple
```

src/components/Tabs/Settings/VectorFieldSettings.tsx
```diff
@@ -17,15 +19,24 @@ const VECTOR_OPTION_MOTION = {
...
+      vectorMotionTimeIntervals: state.vectorMotionTimeIntervals,
+      vectorScaleFactor: state.vectorScaleFactor,
+      vectorTooltipMode: state.vectorTooltipMode,
+      vectorColor: state.vectorColor,
+      setVectorVisible: state.setVectorVisible,
+      setVectorKey: state.setVectorKey,
+      setVectorMotionTimeIntervals: state.setVectorMotionTimeIntervals,
+      setVectorScaleFactor: state.setVectorScaleFactor,
+      setVectorTooltipMode: state.setVectorTooltipMode,
+      setVectorColor: state.setVectorColor,
```
aswallace, line 37: style nit: possibly a good place to alphabetize

src/colorizer/workers/SharedWorkerPool.ts
```diff
@@ -72,19 +72,26 @@ export default class SharedWorkerPool {
...
   }
 
-  terminate(): void {
-    this.workerPool.terminate();
+  async terminate(): Promise<void> {
+    await this.workerPool.terminate();
   }
 }
+
+let workerPool = new SharedWorkerPool();
```
aswallace, line 92: github flag on `let`

src/components/Tooltips/CanvasHoverTooltip.tsx
```diff
@@ -1,27 +1,20 @@
...
-  VectorTooltipMode,
-  ViewerConfig,
-} from "../../colorizer";
+import { AnnotationSelectionMode, VECTOR_KEY_MOTION_DELTA, VectorTooltipMode } from "../../colorizer";
 import { numberToStringDecimal } from "../../colorizer/utils/math_utils";
 import { AnnotationState } from "../../colorizer/utils/react_utils";
+import { selectVectorConfigFromState } from "../../state/slices";
 import { FlexColumn, FlexRow } from "../../styles/utils";
 
+import { useViewerStateStore } from "../../state/ViewerState";
```
aswallace, line 12: I'm curious about the import organization (e.g., state-related import here vs in line 9)

aswallace (approved):
Looks good to me! Works and seems logically consistent

src/components/Tooltips/CanvasHoverTooltip.tsx
ShrimpCryptid: There's an import rule that's separating out imports by file capitalization, which I'm planning to disable once this series of PRs is through. Unfortunately it'll cause horrendous merge conflicts so I'm waiting to do it!

src/colorizer/workers/SharedWorkerPool.ts
ShrimpCryptid, line 92:
```suggestion
const workerPool = new SharedWorkerPool();
```

src/state/utils/store_utils.ts
```diff
@@ -32,3 +32,33 @@ export const addDerivedStateSubscriber = <T, const U>(
...
+/**
+ * Returns a new callback function that wraps the original callback, but debounces
+ * repeated calls by a specified number of milliseconds.
+ * @param callback The original callback function to debounce.
+ * @param debounceMs The number of milliseconds to wait before calling the
+ * debounced callback. Defaults to 250ms.
+ */
+export const makeDebouncedCallback = <T, U, CallbackFn extends (state: T) => Partial<U> | undefined | void>(
+  callback: CallbackFn,
+  debounceMs: number = 250
```
ascibisz, line 45: Out of curiosity, how do you choose an amount of time for a debounced callback?

ascibisz (approved):
Everything works as expected (thanks, as always, for the preview link), and the code looks good to me! I'm not the strongest TypeScript developer, but these changes look good and make sense to me :)

src/state/utils/store_utils.ts
ShrimpCryptid: It's arbitrary! I might actually tweak this to 100ms because that's what it was originally.
The goal is to make sure that we're not firing a ton of expensive computations while a user is rapidly flipping through values. Since the time interval is a slider, it'll fire callbacks on every value in a continuous range.

PR #584, opened 2025-03-07
refactor: Moved frame and TimeControls into viewer state (state pt. 7)

ShrimpCryptid:
Problem
=======
Part 7 of ??? for #492, "refactor state management."

This change moves the current frame and the TimeControls object into shared viewer state, and handles interactions with the dataset.

*Estimated review size: medium, 30 minutes (line count seems high because of a minor change in a bunch of unit tests)*

Solution
========
- Removed TimeControls' dependency on ColorizeCanvas/CanvasWithOverlay.
- Moved TimeControls, playback FPS, and current frame into shared viewer state, as well as the `setFrame` action.
- Added unit tests for time slice.

Misc:
- Updated setters for dataset to be async in existing unit tests, due to async subscribers to dataset changes.
- Added caching to backdrop frames in Dataset.
- Fixed loading in `ColorizeCanvas` so frame fetch requests from old Datasets are dropped.

## Type of change

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-584/
2. Experiment with playback! Pause, play, drag the time slider. Try using the plot tab or scatterplot tab to jump to different time points.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/c0715ea7-318f-4fd1-914e-f702293d477d


tests/state/ViewerState/constants.ts
```diff
@@ -1,7 +1,26 @@
...
+export const setDatasetAsync = async (
+  result: { current: ViewerState },
+  dataset: Dataset,
+  datasetKey: string = "some-dataset"
+): Promise<void> => {
+  act(() => {
+    result.current.setDataset(datasetKey, dataset);
+  });
+  await waitFor(() => {});
+};
```
ShrimpCryptid, line 22: Needed to add this because the subscribers to Dataset now cross an async barrier. Any state updates that happen (even if they're instantaneous) will be treated by react-testing-library as happening outside of the `act` hook and it'll spit out warnings.

src/colorizer/TimeControls.ts
```diff
@@ -1,26 +1,27 @@
 import { DEFAULT_PLAYBACK_FPS } from "../constants";
 
-import CanvasWithOverlay from "./CanvasWithOverlay";
-
```
ShrimpCryptid, line 4: Removed dependency on CanvasWithOverlay by passing in a `getFrameFn` on construction

tests/state/ViewerState/dataset_slice.test.ts
```diff
@@ -9,58 +9,54 @@ import {
   MOCK_DATASET_WITHOUT_BACKDROP,
   MockFeatureKeys,
 } from "./constants";
+import { setDatasetAsync } from "./utils";
```
ShrimpCryptid, line 12: Added this utility method in this and the other tests. This is because the listener for the dataset setter in `time_slice` calls an async function that sets the state, and react-testing-library will output a bunch of warnings about asynchronous state updates. The helper function manages this for Dataset.

src/Viewer.tsx
```diff
@@ -206,6 +213,19 @@ function Viewer(): ReactElement {
...
+    const unsubscribe = useViewerStateStore.subscribe(
+      (state) => [state.pendingFrame],
+      ([pendingFrame]) => {
+        if (isTimeSliderDraggedDuringPlayback) {
+          return;
+        }
+        setFrameInput(pendingFrame);
+      }
+    );
+    return unsubscribe;
```
frasercl, line 226: The power of pair programming!

```suggestion
    // When user is controlling time slider, do not sync frame input w/ playback
    if (!isTimeSliderDraggedDuringPlayback) {
      return useViewerStateStore.subscribe((state) => state.pendingFrame, setFrameInput);
    }
```

frasercl: This will likely still incur an unnecessary re-render, but if a better data flow exists we can work that out in the likely future PR which moves the time slider into its own component.

frasercl (approved):
Let the state refactors roll on!

src/Viewer.tsx
ShrimpCryptid: Opened as https://github.com/allen-cell-animated/timelapse-colorizer/issues/589!

src/state/slices/time_slice.ts
```diff
@@ -0,0 +1,101 @@
...
+    set({ playbackFps: clampWithNanCheck(fps, 0, Number.MAX_SAFE_INTEGER) });
+    get().timeControls.setPlaybackFps(fps);
+  },
+});
+
+export const addTimeDerivedStateSubscribers = (store: SubscribableStore<DatasetSlice & TimeSlice>): void => {
+  // When dataset changes:
+  addDerivedStateSubscriber(
+    store,
+    (state) => [state.dataset],
```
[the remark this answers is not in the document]
ShrimpCryptid: 1. Yes! Removed.
2. Updated the method comment, thanks for the note!
3. Oh interesting, this is a good point. From a quick look at the source you linked, I think directly comparing two `Dataset` objects should be fine, since it would use `Object.is` comparison on the data arrays and not an element-by-element comparison of their contents.

meganrm (approved):

PR #586, opened 2025-03-10
refactor: Moved track into shared viewer state (state pt. 8)

ShrimpCryptid:
Problem
=======
Part 8 of ??? for #492, "refactor state management."

This PR moves the selected track into the shared `useViewerStateStore`, and refactors props on components that use it.

Solution
========
- Added `track` field to the `DatasetSlice`, as well as the `setTrack` and `clearTrack` actions.
- Added unit tests
- Changed `Dataset.getTrack` to explicitly return `null` when a track could not be found.
- Updated components to use the new viewer state store.

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-586/
2. Select, deselect a track.
3. Open the plot or scatterplot tab and click around to jump between tracks and times.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/c2505d15-2a65-4562-8b84-c9396ec657fa


src/colorizer/Dataset.ts
```diff
@@ -591,7 +591,10 @@ export default class Dataset {
       }, [] as number[]);
     }
 
-    const track = new Track(trackId, times, ids, centroids, bounds);
+    let track = null;
+    if (ids.length > 0) {
+      track = new Track(trackId, times, ids, centroids, bounds);
+    }
```
ShrimpCryptid, line 597: Explicitly return `null` if the track is empty. Before, if a track didn't exist in the Dataset, a Track with no IDs would be returned. This required the client to know that this behavior was possible and handle it as an error case.

src/Viewer.tsx
```diff
@@ -815,17 +788,7 @@ function Viewer(): ReactElement {
       key: TabType.TRACK_PLOT,
       children: (
         <div className={styles.tabContent}>
-          <PlotTab
-            setFrame={setFrame}
-            findTrackInputText={findTrackInput}
-            setFindTrackInputText={setFindTrackInput}
```
ShrimpCryptid, line 821: The contents of the track searchbox got to move into the `PlotTab` component!

ShrimpCryptid:
> I'm assuming if this is still working than this is good to go. Overall seems like some better error handling but otherwise it's a lot to process. let me know if you there anything in particular that needs looking at

Definitely agree that this is a lot of changes. Anything you'd like me to do next time to make this easier to review?

meganrm:
> > I'm assuming if this is still working than this is good to go. Overall seems like some better error handling but otherwise it's a lot to process. let me know if you there anything in particular that needs looking at
> 
> Definitely agree that this is a lot of changes. Anything you'd like me to do next time to make this easier to review?

you've done in some places, but I like a lot of comments that say "no code changes just moved to [file]" or "this is the main change". but that's a personal preference

src/components/Tabs/PlotTab.tsx
```diff
@@ -26,36 +27,50 @@ const TrackSearch = styled(FlexRowAlignCenter)`
...
+        } else {
+          setFindTrackInput("");
+        }
+      }
+    );
+    return unsubscribe;
+  }, []);
+
+  const [findTrackInput, setFindTrackInput] = useState("");
+  const isLoading = store.currentFrame !== store.pendingFrame;
```
frasercl, line 62: bit confusing that this effect definition uses a state setter which is defined after it

frasercl (approved):

src/components/Tabs/PlotTab.tsx
ShrimpCryptid: Oh, good point. I'll move that.

PR #592, opened 2025-03-13
refactor: Serialization/deserialization pattern for dataset and color ramps (state pt. 11)

ShrimpCryptid:
Problem
=======
What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

Solution
========

with @pairperson1

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


src/state/slices/dataset_slice.ts
```diff
@@ -107,3 +109,47 @@ export const createDatasetSlice: StateCreator<CollectionSlice & DatasetSlice, []
...
+    ret[UrlParam.BACKDROP_KEY] = slice.backdropKey;
+  }
+  return ret;
+};
+
+export const loadDatasetSliceFromParams = (slice: DatasetSlice, params: URLSearchParams): void => {
+  const dataset = slice.dataset;
+  if (!dataset) {
+    // TODO: Throw error here?
+    return;
```
interim17, line 134: Is there any potential issue from a data set slice having feature keys or track set but the data set slice being null? I think your early return handles it and it probably shouldn't happen, just curious if that's where you would want to do some error handling?

src/state/utils/store_io.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Loads the viewer state from the given URL parameters. Note that this MUST be called after the dataset is loaded and set in the store.
+ */
+export const loadViewerStateFromParams = async (store: Store<ViewerState>, params: URLSearchParams): Promise<void> => {
+  // NOTE: Ordering is important here, because of slice dependencies.
+  loadDatasetSliceFromParams(store.getState(), params);
+  // TODO: Add other slices here
+  // Ramp MUST be loaded last because it updates whenever thresholds or feature changes
+  loadColorRampSliceFromParams(store.getState(), params);
```
interim17, line 38: Does anything enforce this required call order?

interim17 (approved):
Asked a couple questions mostly out of curiosity, LGTM

src/state/utils/store_io.ts
ShrimpCryptid: No mechanisms other than documentation. I update this in a later PR to explain why each step is split out specifically.

image

src/state/slices/dataset_slice.ts
ShrimpCryptid: Currently the setters for `featureKey`, `track`, and `backdropKey` will throw an error if there is no Dataset. Generally though, TFE disables most of the UI if `dataset` is null, so I don't expect users to have a way to try and set those values. Does that answer your question?

tests/state/ViewerState/time_slice.test.ts
```diff
@@ -104,11 +104,10 @@ describe("useViewerStateStore: TimeSlice", () => {
     await act(async () => {
       await result.current.setFrame(3);
     });
-    expect(result.current.currentFrame).toBe(3);
     expect(result.current.pendingFrame).toBe(3);
+    expect(result.current.currentFrame).toBe(3);
 
     await setDatasetAsync(result, MOCK_DATASET_WITH_TWO_FRAMES);
-    expect(result.current.currentFrame).toBe(1);
```
ascibisz, line 111: I may have missed it, but what changed in the code that necessitates this test change?

ShrimpCryptid: Ah! Sorry, should have included a comment on this. This test would fail sometimes due to a race condition, since `currentFrame` is set asynchronously. `pendingFrame` still represents the intended time value that the state is loading.

I could also fix this by adding `await sleep(10)` or some other  async timing mechanism.

ascibisz: Gotcha! Thanks for the explanation, I'll leave it up to you whether you want to add some sort of timing mechanism to fix it or not, based on how well tested you think the intended behavior is in other tests

ascibisz (approved):

tests/state/ViewerState/time_slice.test.ts
ShrimpCryptid: I'll leave it for now... there's another PR where I could probably insert some adjustments to the time slice tests.

PR #595, opened 2025-03-18
refactor: Reorganize slices, update serialization/deserialization (state pt. 14)

ShrimpCryptid:
Problem
=======
Part 14 of 15(???) for #492, "refactor state management."

OKAY WE'RE ALMOST DONE! This change does some cleanup in the organization of the slice files-- now, there's only one file that needs to be updated when new slices are added or removed!

I also added a bunch of new types to help scope what parts of the state can be serialized. This allows for the other use case of the state serialization/deserialization, where we want to take some parameters for the app and turn it into a URL, like we do on the landing page.

Solution
========
- Added versions of `encode{}` methods in `url_utils` that handle `undefined` values.
- Renamed `ViewerState` to `ViewerStore`.
- Moved imports from slices out of `ViewerState.tes` and into `state/slices/index.ts`.
- In all slice files:
  - Added a type (`{...}SliceSerializableState`) that defines what parts of the state can be serialized or deserialized.
    - This prevents users from trying to pass in params to the app that can't be loaded, and will also trigger useful linting errors if changes to serialized values aren't updated in the right places.
  - Added a selector for state that is serialized, for use with `subscribe`.
  - Changed all serialization logic to handle `Partial` types, so partial params can be turned into URLs.

### DIFF
--- a/src/colorizer/utils/url_utils.ts
+++ b/src/colorizer/utils/url_utils.ts
@@ -328,6 +328,10 @@ export function encodeColor(value: Color): string {
   return value.getHexString();
 }
 
+export function encodeMaybeColor(value: Color | undefined): string | undefined {
+  return value ? encodeColor(value) : undefined;
+}
+
 export function isHexColor(value: string | null): value is HexColorString {
   const hexRegex = /^#([0-9a-f]{3}){1,2}$/;
   return value !== null && hexRegex.test(value);
@@ -340,6 +344,10 @@ export function decodeHexColor(value: string | null): Color | undefined {
 
 export function encodeNumber(value: number): string {
   return numberToStringDecimal(value, 3);
+}
+
+export function encodeMaybeNumber(value: number | undefined): string | undefined {
+  return value !== undefined ? encodeNumber(value) : undefined;
 }
 
 export function decodeFloat(value: string | null): number | undefined {
@@ -365,6 +373,10 @@ export function parseDrawSettings(
 
 export function encodeBoolean(value: boolean): string {
   return value ? "1" : "0";
+}
+
+export function encodeMaybeBoolean(value: boolean | undefined): string | undefined {
+  return value !== undefined ? encodeBoolean(value) : undefined;
 }
 
 export function decodeBoolean(value: string | null): boolean | undefined {
--- a/src/colorizer/utils/url_utils.ts
+++ b/src/colorizer/utils/url_utils.ts
@@ -29,6 +29,7 @@ import { numberToStringDecimal } from "./math_utils";
 // TODO: This file needs to be split up for easier reading and unit testing.
 // This could also be a great opportunity to reconsider how we store and manage state.
 
+export const URL_COLOR_RAMP_REVERSED_SUFFIX = "!";
 export enum UrlParam {
   TRACK = "track",
   DATASET = "dataset",
@@ -38,7 +39,6 @@ export enum UrlParam {
   THRESHOLDS = "filters",
   RANGE = "range",
   COLOR_RAMP = "color",
-  COLOR_RAMP_REVERSED_SUFFIX = "!",
   PALETTE = "palette",
   PALETTE_KEY = "palette-key",
   SHOW_BACKDROP = "bg",
@@ -550,7 +550,7 @@ export function paramsToUrlQueryString(state: Partial<UrlParams>): string {
   if (state.colorRampKey) {
     if (state.colorRampReversed) {
       includedParameters.push(
-        `${UrlParam.COLOR_RAMP}=${encodeURIComponent(state.colorRampKey + UrlParam.COLOR_RAMP_REVERSED_SUFFIX)}`
+        `${UrlParam.COLOR_RAMP}=${encodeURIComponent(state.colorRampKey + URL_COLOR_RAMP_REVERSED_SUFFIX)}`
       );
     } else {
       includedParameters.push(`${UrlParam.COLOR_RAMP}=${encodeURIComponent(state.colorRampKey)}`);
@@ -703,10 +703,7 @@ export function loadFromUrlSearchParams(urlParams: URLSearchParams): Partial<Url
   let colorRampParam: string | undefined = colorRampRawParam || undefined;
   let colorRampReversedParam: boolean | undefined = undefined;
   //  Color ramps are marked as reversed by adding ! to the end of the key
-  if (
-    colorRampRawParam &&
-    colorRampRawParam.charAt(colorRampRawParam.length - 1) === UrlParam.COLOR_RAMP_REVERSED_SUFFIX
-  ) {
+  if (colorRampRawParam && colorRampRawParam.charAt(colorRampRawParam.length - 1) === URL_COLOR_RAMP_REVERSED_SUFFIX) {
     colorRampReversedParam = true;
     colorRampParam = colorRampRawParam.slice(0, -1);
   }
--- a/src/state/ViewerState.ts
+++ b/src/state/ViewerState.ts
@@ -1,50 +1,8 @@
 import { create } from "zustand";
-import { StateCreator } from "zustand";
 import { subscribeWithSelector } from "zustand/middleware";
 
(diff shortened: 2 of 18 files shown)


src/state/slices/index.ts
```diff
@@ -7,3 +66,72 @@ export * from "./scatterplot_slice";
...
+  TimeSliceSerializableState &
+  VectorSliceSerializableState;
+
+/**
+ * Combined state for the React app. The ViewerState is composed of many smaller
+ * **slices**, modules of related state, actions, and selectors. See
+ * https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md#slices-pattern
+ * for more details on the pattern.
+ */
+export type ViewerStore = ViewerStoreState & ViewerStoreActions;
```
ShrimpCryptid, line 110: This was renamed from `ViewerState` to `ViewerStore`.

src/state/slices/index.ts
ShrimpCryptid, line 1: Contents of this file mostly copied from `ViewerState.ts`. I'll note the changes below.

src/state/slices/index.ts
```diff
@@ -7,3 +66,72 @@ export * from "./scatterplot_slice";
...
+
+export type ViewerStoreActions = BackdropSliceActions &
+  CollectionSliceActions &
+  ColorRampSliceActions &
+  ConfigSliceActions &
+  DatasetSliceActions &
+  ScatterPlotSliceActions &
+  ThresholdSliceActions &
+  TimeSliceActions &
+  VectorSliceActions;
```
ShrimpCryptid, line 88: Explicitly separated state and actions, which are combined to form the final store.

src/state/slices/backdrop_slice.ts
```diff
@@ -27,7 +33,12 @@ type BackdropSliceState = {
   objectOpacity: number;
 };
 
-type BackdropSliceActions = {
+export type BackdropSliceSerializableState = Pick<
+  BackdropSliceState,
+  "backdropVisible" | "backdropBrightness" | "backdropSaturation" | "objectOpacity"
+>;
```
ShrimpCryptid, line 39: Every slice now has a type representation for the subset of state that's serializable! This is used in a few places, including in the serialization/deserialization logic below.

In this case, currently every value of `BackdropSliceState` is serializable. That's not the case for every slice, since some have internal state values or state that is expected to be transient, like the export settings in `ConfigSlice`. 

I use `Pick` here deliberately (rather than setting `BackdropSliceSerializableState = BackdropSliceState`) so that I don't forget to set up serialization for new state values.

src/state/slices/backdrop_slice.ts
```diff
@@ -80,15 +91,23 @@ export const addBackdropDerivedStateSubscribers = (store: SubscribableStore<Back
...
 };
 
+/** Selects state values that serialization depends on. */
+export const backdropSliceSerializationDependencies = (slice: BackdropSlice): BackdropSliceSerializableState => ({
+  backdropVisible: slice.backdropVisible,
+  backdropBrightness: slice.backdropBrightness,
+  backdropSaturation: slice.backdropSaturation,
+  objectOpacity: slice.objectOpacity,
+});
+
```
ShrimpCryptid, line 110: Eventually when URL sync is set up, I'll want the URL to update ONLY when serializable values change. 
I set up a selector in each slice file to specify which values URL syncing should be listening for changes to.

src/state/slices/color_ramp_slice.ts
```diff
@@ -215,32 +226,45 @@ export const addColorRampDerivedStateSubscribers = (
...
 
-  ret[UrlParam.KEEP_RANGE] = encodeBoolean(slice.keepColorRampRange);
+  ret[UrlParam.KEEP_RANGE] = encodeMaybeBoolean(slice.keepColorRampRange);
 
-  const range = slice.colorRampRange;
-  ret[UrlParam.RANGE] = range.map(encodeNumber).join(",");
+  if (slice.colorRampRange !== undefined) {
+    const range = slice.colorRampRange;
+    ret[UrlParam.RANGE] = range.map(encodeNumber).join(",");
+  }
```
ShrimpCryptid, line 242: I recommend turning off whitespace changes, since some of these changes are just me adding checks for `undefined` values.

src/state/slices/backdrop_slice.ts
```diff
@@ -80,15 +91,23 @@ export const addBackdropDerivedStateSubscribers = (store: SubscribableStore<Back
   );
 };
 
-export const serializeBackdropSlice = (state: BackdropSlice): SerializedStoreData => {
+export const serializeBackdropSlice = (state: Partial<BackdropSliceSerializableState>): SerializedStoreData => {
```
ShrimpCryptid, line 94: This changed to `Partial` because sometimes we turn a bit of state into a URL, without including definitions for every serializable value in state.

For example, on the landing page, I create URLs using partial states, containing just the collection URL and the starting frame number.

src/colorizer/utils/url_utils.ts
```diff
@@ -29,6 +29,7 @@ import { numberToStringDecimal } from "./math_utils";
 // TODO: This file needs to be split up for easier reading and unit testing.
 // This could also be a great opportunity to reconsider how we store and manage state.
 
+export const URL_COLOR_RAMP_REVERSED_SUFFIX = "!";
```
ShrimpCryptid, line 32: This constant got moved out of `UrlParam` because it's technically not its own param, just a single character used in one of the params.

src/state/ViewerState.ts
ShrimpCryptid, line 1: The imports here got moved into `slices/index.ts`.

src/state/utils/store_io.ts
ShrimpCryptid, line 1: Unit tests for serialization/deserialization here will be added in the next PR, when it replaces `url_utils.ts` and the corresponding tests are removed.

src/state/ViewerState.ts
ShrimpCryptid, line 1: The imports here got moved into `slices/index.ts`.

src/state/slices/dataset_slice.ts
```diff
@@ -7,27 +7,22 @@ import { CollectionSlice } from "./collection_slice";
...
+  track: Track | null;
+  /** The key of the backdrop image set in the current dataset. `null` if there
+   * is no Dataset loaded or if the dataset does not have backdrops. */
+  backdropKey: string | null;
+};
+
+export type DatasetSliceSerializableState = Pick<
+  DatasetSliceState,
+  "datasetKey" | "featureKey" | "track" | "backdropKey"
+>;
```
ascibisz, line 23: I just learned through this PR what Pick is 😅, so I have a basic question. What makes you choose Pick vs Omit in a situation like this? Is it because you more specifically want to identify what parts you want instead of what parts you don't want?

ShrimpCryptid: Yes, exactly! In this case, I want a developer (me) to have to manually add new properties that will be serialized by picking them, rather than having them be included in the type automatically. 

I _could_ get the same final type with `Omit`, but it would be a different development experience.

If I did `Omit<DatasetSliceState, "dataset">`, any new property would be marked as serializable automatically, which could potentially lead to bugs if I don't actually update the methods that serialize the slice.

ascibisz (approved):
Looks good to me!

frasercl (approved):
Looks great!

src/state/utils/store_io.ts
```diff
@@ -16,30 +26,90 @@ import {
...
+    ret[UrlParam.COLLECTION] = params.collectionParam;
+  }
+  if (params.datasetParam) {
+    ret[UrlParam.DATASET] = params.datasetParam;
+  }
+  return removeUndefinedProperties({
+    // Order collection + dataset first in params, but override the default
+    // serialized collection fields by destructuring a second time.
+    ...ret,
+    ...serializeViewerState({ ...params }),
```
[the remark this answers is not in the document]
ShrimpCryptid: Ah, good catch, it's not.

PR #600, opened 2025-03-20
feature: Show error banner when URL parameter parsing fails

ShrimpCryptid:
Problem
=======
While I think the URL parsing should be pretty robust against bad inputs, it's possible that there are edge cases I haven't caught. If that happens, I want users to be able to know that the error happened and quickly raise attention to it.

This PR adds handling for failed URL parameter parsing by showing an alert banner and prompting users to submit a GitHub issue ticket. 

Solution
========
- Errors during URL parsing are now shown in an onscreen alert banner.
  - Clicking the link in the alert banner will open a new tab where users can submit a GitHub issue ticket, which is auto-populated with diagnostic information.
- Moved some build-time constants

## Type of change

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/4c5b5cc7-24a9-4728-a62d-53e77629b0ad

image

image


src/main.tsx
```diff
@@ -21,11 +21,8 @@ if (locationUrl.hash !== "" || isEncodedPathUrl(locationUrl)) {
   window.history.replaceState(null, "", newRelativePath);
 }
 
-const version = import.meta.env.VITE_APP_VERSION;
-const basename = import.meta.env.BASE_URL;
-
-console.log(`Timelapse Feature Explorer - Version ${version}`);
-console.log(`Timelapse Feature Explorer - Basename ${basename}`);
```
ShrimpCryptid, line 28: Moved these into `src/constants` so other parts of the app can access them.

src/state/utils/store_io.ts
```diff
@@ -122,7 +122,7 @@ export const serializedDataToUrl = (data: SerializedStoreData): string => {
  * Loads the viewer state from the given URL parameters. Note that this MUST be
  * called after the collection and dataset are loaded and set in the store.
  */
-export const loadViewerStateFromParams = async (store: Store<ViewerStore>, params: URLSearchParams): Promise<void> => {
+export const loadViewerStateFromParams = (store: Store<ViewerStore>, params: URLSearchParams): void => {
```
ShrimpCryptid, line 125: This function wasn't doing anything that required it to be async, and making it synchronous simplified the error handling.

src/Viewer.tsx
```diff
@@ -418,7 +419,12 @@ function Viewer(): ReactElement {
       setIsDatasetLoading(false);
 
       // Load the viewer state from the URL after the dataset is loaded.
-      loadViewerStateFromParams(useViewerStateStore, searchParams);
+      try {
+        loadViewerStateFromParams(useViewerStateStore, searchParams);
+      } catch (error) {
+        console.error("Failed to load viewer state from URL:", error);
```
meganrm, line 425: remove?

src/components/Banner/alert_templates.tsx
```diff
@@ -0,0 +1,79 @@
...
+    \`\`\`
+
+    **Error:**
+    \`\`\`
+    ${error}
+    \`\`\`
+
+    **Stack trace:**
+    \`\`\`
+    ${error.stack}
```
rugeli, line 56: is `error.stack` always defined?

src/components/Banner/alert_templates.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * from the URL. Includes a link to report the issue on GitHub.
+ */
+export const showFailedUrlParseAlert = (url: string, error: Error): AlertBannerProps => {
+  let errorMessage = error.message;
+  if (errorMessage.length > MAX_ERROR_MESSAGE_LENGTH) {
+    errorMessage = errorMessage.slice(0, MAX_ERROR_MESSAGE_LENGTH) + "...";
+  }
+  // Break URL into lines for each param for better readability
+  url = url.replace("?", "?\n");
+  url = url.replaceAll("&", "\n&");
```
rugeli, line 30: Just curious: is this string replacement robust enough for all possible input url formats?

ShrimpCryptid: It shouldn't cause any problems if the URL does, I think, since we're only adding line breaks.
For example, if you copy the below into the address bar, it should still work as normal.
```
https://timelapse.allencell
.org
```

src/components/Banner/alert_templates.tsx
ShrimpCryptid: Hmm, good question. It looks like MDN has conflicting notes about it. It's not strictly standardized but is usually defined in most JS environments. I think it's fine to include, since if it's not defined it'll just return `undefined` as a string.

rugeli (approved):

PR #601, opened 2025-03-21
refactor: Make Canvas interface, simplify ColorizeCanvas API (canvas pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of 2ish for #560, "Refactor CanvasOverlay to be its own class, make generic Canvas."

This change makes the new `IRenderCanvas`, a new interface for the `ColorizeCanvas` class. Eventually, the plan is for the new 3D canvas renderer to also implement `IRenderCanvas`. This will also make it easier to refactor `CanvasWithOverlay` in the next PR.

*Estimated review size: large, 40 minutes*

Solution
========
- Creates the new `IRenderCanvas` interface.
- Refactors `ColorizeCanvas` with the new `setParams` method, which accepts a subset of the `ViewerStoreState`.
- Removes unused setters from `CanvasWrapper.tsx`.

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-601/
2. Change canvas settings like the currently selected track, color ramp, feature, time, etc. Everything should work as it did before.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/6c66c4d6-780f-4f40-a9e8-96bccd66e7b9

Second video with an overview of the changes:

https://github.com/user-attachments/assets/f7693882-e13d-4677-b03c-537bce3fb519

Keyfiles:
-----------------------
1. `IRenderCanvas.ts`
3. `ColorizeCanvas.ts`

PR #606, opened 2025-03-24
feat: Canvas now reports frame loading results (canvas pt. 2)

ShrimpCryptid:
Problem
=======
Part 2 of 3 for #601, refactoring the Canvas to make it more modular.

This change adds the ability for the canvas to report the results of frame loading, including potential error states and out-of-bounds requests. The most recent frame load result is also stored in state, for things like onscreen error messaging.

*Estimated review size: small, 20 minutes*

Solution
========
- Adds the `FrameLoadResult` type.
- Updates the `setFrame` function in `IRenderCanvas` to return a result type.
  - Also formalizes a callback used to report missing frame data in `IRenderCanvas`. 
- Stores the result of the last loaded frame in viewer state.
- Fixed unit tests.

Steps to Verify:
----------------
1. Open PR preview. The dataset should load in without any issues: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-606/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json
2. Click around on the timeline, then return to 0. Copy and paste in this new dataset, which is missing its first frame: https://raw.githubusercontent.com/allen-cell-animated/timelapse-colorizer/refs/heads/test-datasets/docs/example_dataset/manifest.json
3. An error message should appear onscreen.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/1c6e9254-5e03-4be3-9979-1f1fe1fe79b1

-----------------------
1. `types.ts`
2. `IRenderCanvas.ts`
3. `ColorizeCanvas.ts`
4. `time_slice.ts`


src/components/Banner/hooks.tsx
ShrimpCryptid, line 1: Fixes a bug where the `useCallback` would update despite not depending on `bannerProps` directly, which could cause repeated error messages

src/state/slices/time_slice.ts
```diff
@@ -48,17 +57,24 @@ export const createTimeSlice: StateCreator<TimeSlice & DatasetSlice, [], [], Tim
...
-      set({ currentFrame: frame });
+      const result = await get().frameLoadCallback(frame);
+      if (result !== null) {
+        set({ currentFrame: frame });
+      } else if (get().pendingFrame === frame) {
+        // Reset pendingFrame if it hasn't changed. (e.g. no other calls to
+        // setFrame were made while this one was loading)
+        // TODO: More robust handling for requests? Request IDs?
+        set({ pendingFrame: get().currentFrame });
+      }
```
ShrimpCryptid, line 68: Now that `frameLoadCallback` can report the status of the load, we can do more validation during this step! A `null` result means that the frame requested was out of bounds.

src/state/slices/time_slice.ts
```diff
@@ -48,17 +57,24 @@ export const createTimeSlice: StateCreator<TimeSlice & DatasetSlice, [], [], Tim
...
     }
   ),
-  loadFrameCallback: (_frame: number) => {
-    return Promise.resolve();
+  frameLoadCallback: (frame: number): Promise<FrameLoadResult> => {
+    return Promise.resolve({ frame, isFrameLoaded: false, isBackdropLoaded: false, backdropKey: null });
```
ShrimpCryptid, line 72: Fake placeholder, which is expected to be changed immediately on store init in the actual app. I had it return a FrameLoadResult to keep the unit tests working with minimal changes.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -553,18 +555,14 @@ export default class ColorizeCanvas implements IRenderCanvas {
     this.render();
   }
 
-  /**
-   * Sets the current frame of the canvas, loading the new frame data if the
-   * frame number changes.
-   * @param index Index of the new frame.
-   * @param forceUpdate Force a reload of the frame data, even if the frame
-   * is already loaded.
-   */
```
ShrimpCryptid, line 562: Removed since docstring is now in `IRenderCanvas.ts`

src/colorizer/types.ts
```diff
@@ -76,6 +76,18 @@ export const featureTypeSpecs: { [T in FeatureDataType]: FeatureTypeSpec<T> } =
...
+export type FrameLoadResult = {
+  frame: number;
+  /** False if frame loading encountered an error. */
+  isFrameLoaded: boolean;
+  backdropKey: string | null;
+  /**
+   * False if backdrop loading encountered an error. Note that this will be
+   * true if the backdrop is not visible.
+   */
+  isBackdropLoaded: boolean;
```
ShrimpCryptid, line 88: I guess a better name for this would be, `didBackdropLoadEncounterError` but it's not as concise. Any thoughts?

src/state/slices/time_slice.ts
```diff
@@ -76,9 +92,15 @@ export const createTimeSlice: StateCreator<TimeSlice & DatasetSlice, [], [], Tim
...
-        set({ currentFrame: frame });
+      .frameLoadCallback(frame)
+      .then((result) => {
+        if (result !== null) {
+          set({ currentFrame: result.frame });
+        } else if (get().pendingFrame === frame) {
+          // Reset pendingFrame if it hasn't changed. (e.g. no other calls to
+          // setFrame were made while this one was loading)
+          set({ pendingFrame: get().currentFrame });
+        }
```
ShrimpCryptid, line 103: Note also, I didn't set `frameLoadResult` in this same step. I ended up having all changes to it go through the callback set in `Viewer.tsx`, because there are situations where `ColorizeCanvas` calls `setFrame` on itself (like when Datasets are being changed). Because those `setFrame` calls aren't being called from here in the TimeSlice, the slice would have no way to know the result of the most recent frame load request.

This is why there's still a callback in `ColorizeCanvas/IRenderCanvas` for when frames are loaded-- I had a discussion with Cameron and we agree that this pattern makes sense because it's like an event handler/event listener.

src/Viewer.tsx
```diff
@@ -512,10 +511,10 @@ function Viewer(): ReactElement {
       if (target && timeSliderContainerRef.current?.contains(target as Node)) {
         // If the user clicked and released on the slider, update the
         // time immediately.
-        setFrame(frameInput);
+        await setFrame(frameInput);
       }
       if (isTimeSliderDraggedDuringPlayback) {
-        setFrame(frameInput);
+        await setFrame(frameInput);
```
ShrimpCryptid, line 517: Fixes a race condition bug where clicking on the time slider would get ignored

src/components/Banner/hooks.tsx
```diff
@@ -58,7 +58,7 @@ export const useAlertBanner = (
       ignoredBannerMessages.current.add(props.message);
       setBannerProps((previousBannerProps) => [...previousBannerProps, props]);
     },
-    [bannerProps, ignoredBannerMessages.current]
```
ShrimpCryptid, line 61: Tiny fix! I didn't end up using it in this PR, but this callback (`showAlert`) had an unused dependency on `bannerProps` that made the callback update more often than it needed to.

src/state/slices/time_slice.ts
ShrimpCryptid: See also a note below about why we don't set `frameLoadResult` here.

src/state/slices/time_slice.ts
```diff
@@ -48,17 +57,24 @@ export const createTimeSlice: StateCreator<TimeSlice & DatasetSlice, [], [], Tim
...
-      set({ currentFrame: frame });
+      const result = await get().frameLoadCallback(frame);
+      if (result !== null) {
+        set({ currentFrame: result.frame, frameLoadResult: result });
+      } else if (get().pendingFrame === frame) {
+        // Reset pendingFrame if it hasn't changed. (e.g. no other calls to
+        // setFrame were made while this one was loading)
+        // TODO: More robust handling for requests? Request IDs?
+        set({ pendingFrame: get().currentFrame });
+      }
```
ShrimpCryptid, line 68: Now that `frameLoadCallback` can report the status of the load, we can do more validation during this step! A `null` result means that the frame requested was out of bounds.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -553,18 +555,15 @@ export default class ColorizeCanvas implements IRenderCanvas {
...
-  public async setFrame(index: number, forceUpdate: boolean = false): Promise<void> {
+  public async setFrame(index: number, forceUpdate: boolean = false): Promise<FrameLoadResult | null> {
     const dataset = this.params?.dataset;
-    // Ignore same or bad frame indices
-    if (!dataset || (!forceUpdate && this.currentFrame === index) || !dataset.isValidFrameIndex(index)) {
-      return;
+    if (!dataset || !dataset.isValidFrameIndex(index)) {
+      // Out of bounds
+      return null;
```
toloudis, line 562: why not have this return a FrameLoadResult also?

src/colorizer/types.ts
toloudis: `backdropError`.  And maybe change the other one to `frameError`.   Then you can write code like `if (frameResult.backdropError || frameResult.frameError) ...`
You could even keep extended error info in here if needed.

src/state/slices/time_slice.ts
```diff
@@ -1,24 +1,32 @@
...
+   * The frame that is currently being loaded. If no load is happening,
    * `pendingFrame === currentFrame`.
    */
   pendingFrame: number;
   /** The currently loaded and displayed frame. */
   currentFrame: number;
   playbackFps: number;
   timeControls: TimeControls;
-  loadFrameCallback: (frame: number) => Promise<void>;
+  frameLoadCallback: IRenderCanvas["setFrame"];
```
toloudis, line 24: putting callbacks in state seems like a code smell to me. Are you doing this anywhere else? Is this ok to do?

src/colorizer/ColorizeCanvas.ts
ShrimpCryptid: I wanted to distinguish between three states, failed loading (no dataset or invalid frame), partial success (FrameLoadResult but some flags indicate errors) or full success. I guess I could make FrameLoadResult have representations for all three, but the first (failed loading) means that we didn't even try because the parameters were incorrect/invalid.

src/state/slices/time_slice.ts
ShrimpCryptid: We're not doing this anywhere else. It comes down to frame setting needing to be reflected in state since it's accessed in so many places, but also that it needs to interact with the `canvas` to load and render frame data, which exists outside of state.

toloudis: This feels so wrong to me, but I found one zustand discussion where they say it's ok.  It just violates my philosophical sense of what "state" should be and in particular that it should be serializable.

toloudis (approved):

src/state/slices/time_slice.ts
ShrimpCryptid: Something I _could_ do is put the `canvas` in a context, and have the `setFrame` take in an optional callback for the loading action. Every caller of `setFrame` would need to pull the `canvas` out of the context (or a function that wraps `setFrame` that links it to `canvas`) and pass in the canvas' `setFrame` method as a callback.

It gets a bit messier too though because TimeControls is a class that's currently in state too, and should probably be moved out of it in that case? But it needs immediate access to a lot of time-related state.

src/colorizer/ColorizeCanvas.ts
```diff
@@ -225,15 +226,16 @@ export default class ColorizeCanvas implements IRenderCanvas {
 
     this.canvasResolution = new Vector2(1, 1);
     this.categoricalPalette = new ColorRamp(["black"]);
-    this.currentFrame = 0;
+    this.currentFrame = -1;
```
rugeli, line 229: I like the change from 0(frame 0 loaded) to -1(no frame loaded) here! Creates a clearer distinction.

src/state/slices/time_slice.ts
```diff
@@ -48,17 +57,24 @@ export const createTimeSlice: StateCreator<TimeSlice & DatasetSlice, [], [], Tim
     () => get().currentFrame,
     async (frame) => {
       set({ pendingFrame: frame });
-      await get().loadFrameCallback(frame);
-      set({ currentFrame: frame });
+      const result = await get().frameLoadCallback(frame);
+      if (result !== null) {
+        set({ currentFrame: result.frame, frameLoadResult: result });
+      } else if (get().pendingFrame === frame) {
```
rugeli, line 63: Will there be cases that `result` is null but `get().pendingFrame !== frame`?

rugeli (approved):

src/state/slices/time_slice.ts
ShrimpCryptid: Yes. In that case, I expect that another `setFrame` call has happened elsewhere in the app after this load was initialized, so the `pendingFrame` has been set to whatever the other value is. If that's the case, we don't want to touch `pendingFrame` in this handler.

(I think this may be unlikely to happen though because `null` results are returned synchronously... I likely need to make a more robust frame request system in the future, but we haven't encountered any UI issues with this SO far.)

PR #608, opened 2025-03-25
refactor: `ColorizeCanvas` as module in `CanvasOverlay` (canvas pt. 4)

### LINKED ISSUE #560: Refactor CanvasOverlay to be its own class, make generic Canvas
## Use Case

With the new 3D viewer, CanvasOverlay should no longer extend ColorizeCanvas but instead hold a reference to an abstract canvas. 

## Acceptance Criteria

- [x] Create an abstract Canvas that specifies shared canvas functionality (set parameters, getFrame, render, return DOM object)
- [x] ColorizeCanvas should extend/implement the abstract Canvas.
- [x] Change CanvasOverlay to be its own class that has a reference to an abstract Canvas.

## Details

Would be easier if #502 is implemented

ShrimpCryptid:
Problem
=======
CLOSES #560, "refactor CanvasOverlay to be its own class"! 

This change allows `ColorizeCanvas` to be a separate class from `CanvasOverlay`, which will let us make a new 3D canvas class for TFE 3D!

*Estimated review size: medium, 30 minutes*

Solution
========
- Moved some scaling calculations out of `ColorizeCanvas` and into a separate `utils` method, with a new `scaleInfo` property.
- Renamed `CanvasWithOverlay` to `CanvasOverlay`, and `ColorizeCanvas` to `ColorizeCanvas2D`.
- Changed `CanvasOverlay` to hold a reference to an `innerCanvas`, which it syncs parameters and rendering with.

## Type of change

Steps to Verify:
----------------
1. Open PR preview:
2. https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-608/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&tab=settings&bg=1
3. All canvas interactions should work like they did before! Try clicking objects, panning or zooming, toggling UI overlays on/off, etc.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/53ff9414-e5c6-4e68-b9b1-7577cf15f161

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -39,10 +39,11 @@ import { FlexRow, FlexRowAlignCenter } from "./styles/utils";
 import { LocationState } from "./types";
 import { loadInitialCollectionAndDataset } from "./utils/dataset_load_utils";
 
-import CanvasWithOverlay from "./colorizer/CanvasWithOverlay";
+import CanvasOverlay from "./colorizer/CanvasOverlay";
 import Collection from "./colorizer/Collection";
-import { BACKGROUND_ID } from "./colorizer/ColorizeCanvas";
+import ColorizeCanvas, { BACKGROUND_ID } from "./colorizer/ColorizeCanvas";
 import { FeatureType } from "./colorizer/Dataset";
+import { renderCanvasStateParamsSelector } from "./colorizer/IRenderCanvas";
 import UrlArrayLoader from "./colorizer/loaders/UrlArrayLoader";
 import { getSharedWorkerPool } from "./colorizer/workers/SharedWorkerPool";
 import { AppThemeContext } from "./components/AppStyle";
@@ -84,7 +85,8 @@ function Viewer(): ReactElement {
   const [, startTransition] = React.useTransition();
 
   const canv = useConstructor(() => {
-    const canvas = new CanvasWithOverlay();
+    const stateDeps = renderCanvasStateParamsSelector(useViewerStateStore.getState());
+    const canvas = new CanvasOverlay(new ColorizeCanvas(), stateDeps);
     canvas.domElement.className = styles.colorizeCanvas;
     // Report frame load results to the store
     canvas.setOnFrameLoadCallback(useViewerStateStore.getState().setFrameLoadResult);
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -41,7 +41,7 @@ import { loadInitialCollectionAndDataset } from "./utils/dataset_load_utils";
 
 import CanvasOverlay from "./colorizer/CanvasOverlay";
 import Collection from "./colorizer/Collection";
-import ColorizeCanvas, { BACKGROUND_ID } from "./colorizer/ColorizeCanvas";
+import ColorizeCanvas2D, { BACKGROUND_ID } from "./colorizer/ColorizeCanvas2D";
 import { FeatureType } from "./colorizer/Dataset";
 import { renderCanvasStateParamsSelector } from "./colorizer/IRenderCanvas";
 import UrlArrayLoader from "./colorizer/loaders/UrlArrayLoader";
@@ -86,7 +86,7 @@ function Viewer(): ReactElement {
 
   const canv = useConstructor(() => {
     const stateDeps = renderCanvasStateParamsSelector(useViewerStateStore.getState());
-    const canvas = new CanvasOverlay(new ColorizeCanvas(), stateDeps);
+    const canvas = new CanvasOverlay(new ColorizeCanvas2D(), stateDeps);
     canvas.domElement.className = styles.colorizeCanvas;
     // Report frame load results to the store
     canvas.setOnFrameLoadCallback(useViewerStateStore.getState().setFrameLoadResult);
--- a/src/colorizer/CanvasOverlay.ts
+++ b/src/colorizer/CanvasOverlay.ts
@@ -28,7 +28,7 @@ import { getPixelRatio } from "./canvas/utils";
 import { CanvasScaleInfo, CanvasType, FrameLoadResult } from "./types";
 
 import { LabelData } from "./AnnotationData";
-import ColorizeCanvas from "./ColorizeCanvas";
+import ColorizeCanvas2D from "./ColorizeCanvas2D";
 import { IRenderCanvas, RenderCanvasStateParams } from "./IRenderCanvas";
 
 /**
@@ -38,7 +38,7 @@ import { IRenderCanvas, RenderCanvasStateParams } from "./IRenderCanvas";
  */
 export default class CanvasOverlay implements IRenderCanvas {
   private canvasElement: HTMLCanvasElement;
-  private canvas: IRenderCanvas;
+  private innerCanvas: IRenderCanvas;
   private ctx: CanvasRenderingContext2D;
 
   private currentFrame: number;
@@ -62,7 +62,7 @@ export default class CanvasOverlay implements IRenderCanvas {
   private annotationStyle: AnnotationStyle;
 
   /** Size of the inner colorized canvas, in pixels. */
-  private canvasResolution: Vector2;
+  private innerCanvasSize: Vector2;
   // Size of the header and footer as of the current render.
   private headerSize: Vector2;
   private footerSize: Vector2;
@@ -90,13 +90,13 @@ export default class CanvasOverlay implements IRenderCanvas {
       footer?: FooterStyle;
     }
   ) {
-    this.canvas = canvas;
+    this.innerCanvas = canvas;
     this.canvasElement = document.createElement("canvas");
     this.canvasElement.style.display = "block";
     this.onFrameLoadCallback = () => {};
 
     this.params = params;
-    this.currentFrame = 0;
+    this.currentFrame = -1;
     this.zoomMultiplier = 1;
     this.panOffset = new Vector2();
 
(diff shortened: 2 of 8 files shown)


src/Viewer.tsx
```diff
@@ -84,7 +85,8 @@ function Viewer(): ReactElement {
   const [, startTransition] = React.useTransition();
 
   const canv = useConstructor(() => {
-    const canvas = new CanvasWithOverlay();
+    const stateDeps = renderCanvasStateParamsSelector(useViewerStateStore.getState());
+    const canvas = new CanvasOverlay(new ColorizeCanvas(), stateDeps);
```
ShrimpCryptid, line 89: `CanvasOverlay` now accepts a `IRenderCanvas` in its constructor!

src/colorizer/ColorizeCanvas.ts
```diff
@@ -291,57 +299,33 @@ export default class ColorizeCanvas implements IRenderCanvas {
...
-      // Canvas has a wider aspect ratio than the frame, so proportional height is 1
-      // and we scale width accordingly.
-      unscaledFrameSizeInCanvasCoords.x = canvasAspect / frameAspect;
-    } else {
-      unscaledFrameSizeInCanvasCoords.y = frameAspect / canvasAspect;
-    }
-
-    // Get final size by applying the current zoom level, where `zoomMultiplier=2` means the frame is 2x
-    // larger than its base size. Save this to use when calculating units with the scale bar.
-    this.frameSizeInCanvasCoordinates = unscaledFrameSizeInCanvasCoords.clone().multiplyScalar(this.zoomMultiplier);
```
ShrimpCryptid, line 323: Scale calculations got moved into `utils`!

src/colorizer/types.ts
```diff
@@ -88,6 +91,40 @@ export type FrameLoadResult = {
...
+   * frame to the [0,1] space of the canvas, accounting for zoom.
+   */
+  frameToCanvasCoordinates: Vector2;
+};
+
+export type Canvas3DScaleInfo = {
+  type: CanvasType.CANVAS_3D;
+};
+
+export type CanvasScaleInfo = Canvas3DScaleInfo | Canvas2DScaleInfo;
```
ShrimpCryptid, line 126: A little unsure about whether this type is the best way to handle scaling info.

src/components/CanvasWrapper.tsx
```diff
@@ -235,14 +234,6 @@ export default function CanvasWrapper(inputProps: CanvasWrapperProps): ReactElem
...
   }, [showTimestamp]);
 
-  useMemo(() => {
-    canv.setCollection(collection);
-  }, [collection]);
-
-  useMemo(() => {
-    canv.setDatasetKey(datasetKey);
-  }, [datasetKey]);
-
```
ShrimpCryptid, line 245: Moved into IRenderCanvas dependencies

src/colorizer/ColorizeCanvas.ts
```diff
@@ -453,17 +437,18 @@ export default class ColorizeCanvas implements IRenderCanvas {
     this.onFrameLoadCallback = callback;
   }
 
-  public setParams(params: RenderCanvasStateParams): void {
+  public async setParams(params: RenderCanvasStateParams): Promise<void> {
```
ShrimpCryptid, line 440: Made async for the situation where a parent caller (`CanvasOverlay`) needs to wait for changes to be applied before re-rendering.

src/colorizer/CanvasOverlay.ts
```diff
@@ -25,23 +25,28 @@ import {
...
 /**
- * Extends the ColorizeCanvas class by overlaying and compositing additional
- * dynamic elements (like a scale bar, timestamp, etc.) on top of the
+ * Wraps an IRenderCanvas class by overlaying and compositing additional
+ * dynamic elements (like a scale bar, timestamp, etc.) on top of a
  * base colorized image.
  */
-export default class CanvasWithOverlay extends ColorizeCanvas {
-  private canvas: HTMLCanvasElement;
+export default class CanvasOverlay implements IRenderCanvas {
```
ShrimpCryptid, line 39: Refactored so it now wraps an `IRenderCanvas` instead of just extending one! I don't know if it's weird for it to also be an IRenderCanvas though?

src/colorizer/CanvasOverlay.ts
```diff
@@ -25,23 +25,28 @@ import {
...
   private ctx: CanvasRenderingContext2D;
 
-  private collection: Collection | null;
-  private datasetKey: string | null;
+  private currentFrame: number;
+  private params: RenderCanvasStateParams;
+  private onFrameLoadCallback: (result: FrameLoadResult) => void;
+
+  private zoomMultiplier: number;
+  private panOffset: Vector2;
```
ShrimpCryptid, line 49: `CanvasOverlay` now tracks these values and passes them to the `canvas` when it's set.

src/colorizer/CanvasOverlay.ts
```diff
@@ -123,62 +134,107 @@ export default class CanvasWithOverlay extends ColorizeCanvas {
...
+    this.canvas = canvas;
+    this.canvas.setResolution(this.canvasResolution.x, this.canvasResolution.y);
+    this.canvas.setOnFrameLoadCallback(this.onFrameLoadCallback);
+    await this.canvas.setParams(this.params);
+    await this.canvas.setFrame(this.currentFrame);
+    if (this.canvas instanceof ColorizeCanvas) {
+      this.canvas.setZoom(this.zoomMultiplier);
+      this.canvas.setPan(this.panOffset.x, this.panOffset.y);
+    }
+    this.render(false);
```
ShrimpCryptid, line 234: When replacing the inner canvas, copy all params + properties to it. This function is async because the inner canvas may need to load frame data before it can render for the first time.

src/colorizer/CanvasOverlay.ts
```diff
@@ -123,62 +134,107 @@ export default class CanvasWithOverlay extends ColorizeCanvas {
...
+    this.render();
+  }
+
+  public setPan(x: number, y: number): void {
+    this.panOffset.set(x, y);
+    if (this.innerCanvas instanceof ColorizeCanvas2D) {
+      this.innerCanvas.setPan(x, y);
+    }
+    this.render();
+  }
```
ShrimpCryptid, line 216: TODO: Zoom and Pan are specific to 2D canvas! This interaction logic will definitely need some work later once we see how 3D panning/zooming is handled.

src/colorizer/CanvasOverlay.ts
interim17: Weighing some of these same questions in Simularium right now and thinking that a general philosophy around composition/inheritance is good thing for our apps to align on (and functional/imperative) if we want them to align over time.

I want to make a class extend another in Simularium but @frasercl may have convinced me that the modular/composition approach is the way to make things interoperable as we move forward.

src/colorizer/canvas/utils.ts
```diff
@@ -79,3 +80,48 @@ export function renderCanvasText(
 
   return new Vector2(textWidth, textHeight);
 }
+
+export function get2DCanvasScaling(
+  frameResolution: Vector2,
+  canvasResolution: Vector2,
+  zoomMultiplier: number
+): Canvas2DScaleInfo {
```
interim17, line 88: Do you expect these params to be well validated before calls of this function?

interim17 (commented):
LGTM

interim17 (approved):

src/colorizer/CanvasOverlay.ts
```diff
@@ -123,62 +134,107 @@ export default class CanvasWithOverlay extends ColorizeCanvas {
 
   // Wrapped ColorizeCanvas functions ///////
 
+  public get resolution(): Vector2 {
+    return this.innerCanvasSize.clone();
+  }
+
+  public getScaleInfo(): CanvasScaleInfo {
+    return this.innerCanvas.getScaleInfo();
+  }
```
meganrm, line 143: any reason for one of these to be a getter and one of these to be a function?

src/colorizer/CanvasOverlay.ts
```diff
@@ -123,62 +134,107 @@ export default class CanvasWithOverlay extends ColorizeCanvas {
...
+
   get domElement(): HTMLCanvasElement {
     // Override base ColorizeCanvas getter with the composited canvas.
-    return this.canvas;
+    return this.canvasElement;
+  }
+
+  setOnFrameLoadCallback(callback: (result: FrameLoadResult) => void): void {
```
meganrm, line 150: `private`?

src/colorizer/CanvasOverlay.ts
ShrimpCryptid: Not particularly... I'll change `scaleInfo` to be a getter

src/colorizer/CanvasOverlay.ts
ShrimpCryptid: Oops, good catch! `public`, actually!

src/colorizer/canvas/utils.ts
ShrimpCryptid: Hmm, good question. I think so since I only ever call this internally.

meganrm (approved):

Issue #545, opened 2025-02-05
🚩 MVP 3D Viewport


ShrimpCryptid:
## Use Case

Minimum viable product for 3D viewing in TFE.

## Acceptance Criteria

- [x] Switch to 3D (Vol-E) viewer if dataset has 3D data
- [x] Vol-E viewer opens and visualizes ZARR segmentation files
- [x] Vol-E viewer has support for existing controls (rotate/zoom/pan)
- [x] Cells in Vol-E can be clicked
- [x] Cells in Vol-E are colorized

## Details

Assumptions:
- We are working with segmented volume data only
- Rendered cells are opaque

PR #610, opened 2025-03-29
refactor: Upgraded dependencies on vite and three, installed vole-core (3D MVP pt. 1)

ShrimpCryptid:
Installs `vole-core` and updates dependencies on `vite` and `three` as part of changes for #545.

Includes some fixes due to breaking changes in `three`.

*Estimated review size: small, 5 minutes*


src/colorizer/types.ts
```diff
@@ -14,6 +14,8 @@ import {
 
 // This file provides a bit of type trickery to allow data loading code to be generic over multiple numeric types.
 
+export type HexColorString = `#${string}`;
+
```
ShrimpCryptid, line 18: `HexColorString` type was removed from `three`, so I defined it locally for the project.

src/colorizer/ColorRamp.ts
```diff
@@ -1,10 +1,24 @@
...
 
 export enum ColorRampType {
   LINEAR,
   HARD_STOP,
 }
 
+// Fixes an issue in Three.js where the new default color management causes all
+// Color objects to be interpreted as sRGB colors, which breaks expected linear
+// interpolation between colors in the ramp.
+ColorManagement.enabled = false;
```
toloudis, line 20: ouch, this one is scary to me since afaik we don't do it in other apps. do you know what version of three introduced this?

toloudis (approved):

src/colorizer/ColorRamp.ts
toloudis: conversion to linear space is actually what we want, so that linear lookups of color textures do a more correct thing. (`texture.colorSpace`)

rugeli (approved):

src/colorizer/ColorRamp.ts
ShrimpCryptid: Removed this and ended up just changing the unit test. 

I didn't explicitly change every `new Color()` or `Color.getHexString()` call to explicitly define the color space, since `three` assumes hex strings to be in sRGB when creating new colors and `Color.getHexString` defaults to sRGB. AFAIK it'll do the right thing, but if you think it's wise to be explicit LMK and I can make that change.

PR #622, opened 2025-04-04
refactor: Added compatibility between `ColorizeCanvas3D` and `OverlayCanvas`.

ShrimpCryptid:
Problem
=======
This PR continues work on #545 (3D viewport MVP), and restores a lot of OverlayCanvas functionality that I had to previously disable!

Before, the OverlayCanvas would request its inner canvas to render, then take the resulting image and draw it into its own canvas before compositing additional elements on top of it. This was necessary so that, in Export mode, we could get the viewport image from a single canvas. This had some issues, however, namely requiring that OverlayCanvas re-render every time its inner canvas rendered or else we would get desyncs. Additionally, all canvas interaction needed to be forwarded to the inner canvas, which broke orbit controls in the 3D canvas.

This change makes the OverlayCanvas transparent, and mounts the inner canvas behind it. This means that the two of them can be on independent rendering timings and allows 3D trackball interactions to work. In Export mode, OverlayCanvas reverts to the old method of manually drawing the inner canvas into itself.

Solution
========

with @pairperson1

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/colorizer/CanvasOverlay.ts
+++ b/src/colorizer/CanvasOverlay.ts
@@ -101,6 +101,9 @@ export default class CanvasOverlay implements IRenderCanvas {
     this.canvasElement.style.display = "block";
     // Let mouse events pass through to the inner canvas.
     this.canvasElement.style.pointerEvents = "none";
+    // Ensure the canvas is rendered on top of the inner canvas.
+    this.canvasElement.style.position = "relative";
+    this.canvasElement.style.zIndex = "1";
 
     // Set up DOM elements, which are structured like:
     // canvasContainerDiv
@@ -111,7 +114,6 @@ export default class CanvasOverlay implements IRenderCanvas {
     this.canvasContainerDiv.style.position = "relative";
     this.canvasContainerDiv.style.width = "100%";
     this.canvasContainerDiv.style.height = "100%";
-    this.canvasContainerDiv.style.zIndex = "1";
 
     this.innerCanvasContainerDiv = document.createElement("div");
     this.innerCanvasContainerDiv.appendChild(this.innerCanvas.domElement);
@@ -123,8 +125,8 @@ export default class CanvasOverlay implements IRenderCanvas {
     this.canvasElement.style.top = "0px";
     this.canvasElement.style.left = "0px";
 
-    this.canvasContainerDiv.appendChild(this.canvasElement);
     this.canvasContainerDiv.appendChild(this.innerCanvasContainerDiv);
+    this.canvasContainerDiv.appendChild(this.canvasElement);
 
     this.onFrameLoadCallback = () => {};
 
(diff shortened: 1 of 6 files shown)


src/colorizer/ColorizeCanvas3D.ts
```diff
@@ -153,7 +158,7 @@ export class ColorizeCanvas3D implements IRenderCanvas {
       }
     }
 
-    // this.view3d.redraw();
+    this.render(false);
```
ShrimpCryptid, line 161: Re-render inner canvas asynchronously after params have been updated.

src/colorizer/ColorizeCanvas3D.ts
```diff
@@ -223,7 +228,8 @@ export class ColorizeCanvas3D implements IRenderCanvas {
     }
 
     const loadVolumeFrame = async (): Promise<FrameLoadResult> => {
-      if (!this.volume || !this.loader) {
+      if ((!this.volume || !this.loader) && !this.isLoadingInitialVolume) {
+        this.isLoadingInitialVolume = true;
```
ShrimpCryptid, line 232: Flag prevents initialization from being triggered twice

ShrimpCryptid:
> LGTM
> 
> Just a note that the preview video won't play for me (offsite, on VPN), but I can load the preview link and datasets. I figure you would want to know as they take you time and effort!

Oh weird, does it work if you refresh?

src/colorizer/CanvasOverlay.ts
```diff
@@ -93,6 +99,36 @@ export default class CanvasOverlay implements IRenderCanvas {
...
+    this.canvasContainerDiv.style.width = "100%";
+    this.canvasContainerDiv.style.height = "100%";
+    this.canvasContainerDiv.style.overflow = "hidden";
+
+    this.innerCanvasContainerDiv = document.createElement("div");
+    this.innerCanvasContainerDiv.appendChild(this.innerCanvas.domElement);
+    this.innerCanvasContainerDiv.style.position = "absolute";
+    this.innerCanvasContainerDiv.style.top = "0px";
+    this.innerCanvasContainerDiv.style.left = "0px";
+    this.innerCanvasContainerDiv.style.zIndex = "0";
```
frasercl, line 124: optional thought: could you assign a single object in place of these individual property declarations?
```typescript
this.innerCanvasContainerDiv.style = {
    position: "absolute";
    top: "0px";
    // ...
};
```

frasercl (approved):
LGTM!

src/colorizer/CanvasOverlay.ts
ShrimpCryptid: I can set it via the `cssText` property!

Issue #203, opened 2024-02-02
Vary starting frame number by dataset


ShrimpCryptid:
## Use Case

Requested by @gouthamr321! Some datasets in the EMT project start at different time offsets. While this can be captured by the metadata properties (`startTimeSeconds`), the slider and interface will always show the first frame as frame 0.

The imaging, regardless of a condition (2D/3D/etc.), starts at the same time. But, expression of markers in different conditions starts at different times. For example, maybe 2D starts expressing Sox2 at frame 100, while 3D starts at frame 245.

## Acceptance Criteria
- Match the slider's starting frame number to a parameter defined in the manifests.

- Show all time frame #'s as being absolute based on that starting time.
  - So if a dataset starts at time 5, advancing by 5 frames should show the frame # as being frame 10.
- Switching between datasets should try and preserve the absolute time if possible.

![image](https://github.com/allen-cell-animated/nucmorph-colorizer/assets/30200665/0f1ecce1-2665-4c2b-b6f0-105963cb2253)

toloudis:
It wasn't stated: are "2D" and "3D" datasets in a collection?
Assuming so, is it sufficient to just have a starting offset value?  
Do we need to somehow account for dropped frames in the middle of a dataset?

ShrimpCryptid:
Yes, 2D and 3D are just two example datasets in the collection.

Having an offset value seems to be sufficient, since frame numbers can only be continuous (contiguous?) ranges under the current dataset structure. Frame numbers are just indices into the array of frame paths.

ShrimpCryptid:
Also would support the colony time comparison for Nucmorph deliverable...

ShrimpCryptid:
Encountered by Gokhan too, bumping this as a priority

toloudis:
as discussed:
* data has a startTimeIndex.  startTimeIndex is the first (leftmost) slider range value. 
* frames and backdrops arrays must be same length.  Empty strings / null sentinels are allowed, for which we will draw "blank" in the viewport.
* Should startTimeIndex this correspond to the minimum value in the array of Times?  If startTimeIndex is 30 but there are no segmented cells at T=30... we still need to have an entry in the frames and backdrops arrays for t=30.)

PR #637, opened 2025-04-19
feat: Calculate lookup tables to support arbitrary segmentation IDs (3D MVP pt. 5)

### LINKED ISSUE #630: Display datasets with arbitrary segmentation IDs
## Use Case

TFE data is structured in big arrays, where each index is a globally-unique ID.
Before, image frames were remapped/preprocessed so that the pixel values could be used to directly index into the big arrays.

Image

However, with TFE 3D, we want to do as little image preprocessing as possible. Segmentation IDs in data may correspond with tracks or be totally arbitrary values. They are guaranteed to be unique within the frame only.

Image

## Acceptance Criteria

- [x] Store the segmentation IDs as a column in the dataset definitions (the unused `object_id` column in the input DataFrame in `colorizer-data?)
- [x] In TFE frontend, use the segmentation IDs to make a lookup table for each frame number, that maps from the segmentation ID on that frame to its global unique index.
  - [x] For each frame, create a texture that maps from the segmentation ID to the global index.
  - [x] Update the shaders so they support the new indirection lookup for IDs.

## Details

One possible approach would be to store the segmentation IDs of each object in the dataset. When the dataset is loaded, we can construct an indirection table for each timepoint that maps from a segmentation ID to the globally-unique index of that object.

Image

For each frame, the CPU can make an array that's passed to the GPU that maps from a segmentation ID as an index to the globally-unique index. One issue with this is that it's potentially a very sparse array, depending on the segmentation IDs used.

Image

There are potentially some optimizations we can make based on the min and max segmentation IDs.

ShrimpCryptid:
Problem
=======
Closes #630, "display datasets with arbitrary segmentation IDs". 

This change allows datasets to skip the preprocessing step for 3D volumes and virtually eliminates it for 2D images as well! It removes a lot of the assumptions we were previously making about the ordering of data, and handles edge cases caused by missing data that we were previously not addressing.

NOTE: This branch will fail unit tests until the corresponding PR in vole-core (https://github.com/allen-cell-animated/vole-core/pull/313) is merged and released.

*Estimated review size: medium, 20-30 minutes (most of line count is unit tests, comments, and whitespace)*

Solution
========
- On load, Datasets now create a lookup for every frame that maps from segmentation IDs to global indices into the feature arrays.
  - This is passed into the 3D and 2D viewports.
- Updated Canvas `getIdAtPixel` to report additional information (both seg ID + global ID) to handle the case where segmentations do not have a corresponding global ID (e.g. data is missing).
- Updated canvas hover tooltip to show segmentation ID for debugging.
- Added unit tests.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-637/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fpeyton.lee%2Ftfe-emt-nuclear-segmentations%2Fdata%2Ftfe%2Fcollection.json&dataset=EMT+3D&feature=volume&t=47&color=matplotlib-cool&keep-range=1&range=1999%2C90963
2. Play through time. Coloring by volume should look consistent (besides occasional flashing)

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/f1c2db34-1571-405e-9e9c-e25ceb4b9730

-----------------------
1. `types.ts`
4. `data_utils.ts`
5. `IRenderCanvas.ts`

### DIFF
--- a/src/colorizer/ColorizeCanvas3D.ts
+++ b/src/colorizer/ColorizeCanvas3D.ts
@@ -124,7 +124,7 @@ export class ColorizeCanvas3D implements IRenderCanvas {
           outlierDrawMode: this.params.outlierDrawSettings.mode,
           outOfRangeDrawMode: this.params.outOfRangeDrawSettings.mode,
           hideOutOfRange: this.params.outOfRangeDrawSettings.mode === DrawMode.HIDE,
-          frameToGlobalIdLookup: dataset.frameToGlobalIdLookup!,
+          frameToGlobalIdLookup: dataset.frameToGlobalIdLookup ?? new Map(),
         };
         this.view3d.setChannelColorizeFeature(volume, channelIndex, feature);
       }
@@ -308,8 +308,7 @@ export class ColorizeCanvas3D implements IRenderCanvas {
     // during colorizer setup.
     if (this.volume?.isLoaded() && dataset) {
       const segId = this.view3d.hitTest(x, y);
-      const globalIdInfo = dataset.frameToGlobalIdLookup?.get(this.currentFrame);
-      if (!globalIdInfo || segId === -1) {
+      if (segId === -1) {
         // Background hit
         return null;
       }
--- a/src/colorizer/ColorizeCanvas3D.ts
+++ b/src/colorizer/ColorizeCanvas3D.ts
@@ -15,8 +15,8 @@ import {
 import { Vector2, Vector3 } from "three";
 
 import { MAX_FEATURE_CATEGORIES } from "../constants";
-import { CanvasScaleInfo, CanvasType, DrawMode, FeatureDataType, FrameLoadResult } from "./types";
-import { hasPropertyChanged } from "./utils/data_utils";
+import { CanvasScaleInfo, CanvasType, DrawMode, FeatureDataType, FrameLoadResult, PixelIdInfo } from "./types";
+import { getGlobalIdFromSegId, hasPropertyChanged } from "./utils/data_utils";
 import { packDataTexture } from "./utils/texture_utils";
 
 import { IRenderCanvas, RenderCanvasStateParams } from "./IRenderCanvas";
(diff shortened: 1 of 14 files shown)


src/components/Tooltips/CanvasHoverTooltip.tsx
ShrimpCryptid, line 1: Please turn off whitespace changes when reviewing this file!

src/colorizer/Dataset.ts
```diff
@@ -601,11 +568,13 @@ export default class Dataset {
       // Construct default segIds array (0, 1, 2, ...)
       this.segIds = new Uint32Array(this.numObjects);
       for (let i = 0; i < this.numObjects; i++) {
-        this.segIds[i] = i;
+        this.segIds[i] = i + 1;
```
ShrimpCryptid, line 571: +1 since 0 is background

src/colorizer/ColorizeCanvas3D.ts
```diff
@@ -289,7 +289,7 @@ export class ColorizeCanvas3D implements IRenderCanvas {
       return;
     }
     const id = this.params.track ? this.params.track.getIdAtTime(this.currentFrame) : -1;
-    this.view3d.setSelectedID(this.volume, this.params.dataset.segmentationChannel ?? 0, id + 1);
```
ShrimpCryptid, line 292: This was a bug caused by highlighted ids being off by one in the vole-core shader.

src/components/Tooltips/CanvasHoverTooltip.tsx
```diff
@@ -3,7 +3,7 @@ import React, { PropsWithChildren, ReactElement, useCallback } from "react";
 import styled from "styled-components";
 import { useShallow } from "zustand/shallow";
 
-import { AnnotationSelectionMode, VECTOR_KEY_MOTION_DELTA, VectorTooltipMode } from "../../colorizer";
+import { AnnotationSelectionMode, PixelIdInfo, VECTOR_KEY_MOTION_DELTA, VectorTooltipMode } from "../../colorizer";
```
ShrimpCryptid, line 6: I recommend turning off whitespace when reviewing this file, since I had to add a few conditionals that caused a lot of indenting.

ascibisz (approved):
LGTM!

src/colorizer/types.ts
```diff
@@ -230,6 +231,66 @@ export enum AnnotationSelectionMode {
...
+   */
+  minSegId: number;
+};
+
+export type PixelIdInfo = {
+  /** Segmentation ID of the pixel.*/
+  segId: number;
+  /** Global ID derived from the segmentation ID, used to index into data
+   * arrays. `undefined` if the segmentation ID is missing from the dataset.
+   */
```
frasercl, line 290: tiny thing for formatting consistency with the rest of the file


```suggestion
  /**
   * Global ID derived from the segmentation ID, used to index into data
   * arrays. `undefined` if the segmentation ID is missing from the dataset.
   */
```

src/colorizer/Dataset.ts
```diff
@@ -76,17 +77,24 @@ export default class Dataset {
...
    *
-   * For any object with segmentation ID `segId` at time `t`, the global
-   * index of that object is given by `segId + frameToIdOffsets[t]`.
+   * The global ID is `NaN` or `-1` if there is no global ID that matches that
+   * segmentation ID, such as when rows are missing in the dataset.
+   *
+   * See `GlobalIdLookupInfo` for more details.
    */
-  public frameToIdOffset: Uint32Array | null;
+  public frameToGlobalIdLookup: Map<number, GlobalIdLookupInfo> | null;
```
frasercl, line 97: Why a `Map` and not an array? If we expect frame numbers to be dense and begin at 0, an array would store and retrieve them more efficiently.

frasercl (approved):
Looks good overall, though I think switching to an array is worth serious consideration.

src/colorizer/Dataset.ts
ShrimpCryptid: There *is* an issue ticket in the backlog for datasets that have frame numbers that don't start at 0, I believe. #203 
I think switching this to an array is fine with me, but it would require a refactor on the vole-core side. I'm going to merge as-is but I'll open an issue ticket about refactoring this later.

ShrimpCryptid: Created as #655

PR #643, opened 2025-04-22
feat: Support 3D centroids (3D pt. 6)

### LINKED ISSUE #556: 3D pt. 2 - Support 3D centroids (+bounding box coords)
## Use Case

Allow centroid data to be provided as 3D coordinates instead of 2D only.

## Acceptance Criteria

- [x] Add a flag in the dataset (either manifest or elsewhere) that causes centroids + bounding box coordinates to be parsed as 3D instead of 2D.
- [x] Update all bounding-box- and centroid-related logic to return/handle 3D coordinates
  - [ ] For backwards compatibility with existing datasets, set the Z-value to 0 if coordinates are in 2D.

## Details

- See related data ticket: https://github.com/allen-cell-animated/colorizer-data/issues/77

ShrimpCryptid:
Problem
=======
Closes #556, "support 3D centroids."

This change updates Dataset to store centroids with 3D coordinates instead of just 2D, and updates all of the internal logic dealing with centroids data to match.

Solution
========
- Updated `Dataset` to convert from 2D centroids to 3D centroids on load.
- Updated logic for track paths, annotation, and vectors to use 3D centroid coordinates.
- Fixed a bug in `Dataset.getCentroid` where it would return `undefined` for vectors with `0` as a component.
- Added unit tests.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-643/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Small&feature=volume&track=74440&t=118&path=1&vc=1&vc-key=_motion_&vc-color=000000&vc-scale=8&vc-tooltip=m&vc-time-int=5
2. Click on a track and play through time. The motion vectors and track path should update correctly.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/95a348de-ddc1-41db-b613-1b4cf1d624c6

### DIFF
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -697,7 +697,7 @@ export default class Dataset {
     const centroidsData = this.centroids;
     if (centroidsData) {
       centroids = ids.reduce((result, i) => {
-        result.push(centroidsData[3 * i], centroidsData[3 * i + 1], centroidsData[3 * i + 2]);
+        result.push(...this.getCentroid(i)!);
         return result;
       }, [] as number[]);
     }
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -667,7 +667,7 @@ export default class Dataset {
     const x = this.centroids?.[index];
     const y = this.centroids?.[index + 1];
     const z = this.centroids?.[index + 2];
-    if (x && y && z) {
+    if (x !== undefined && y !== undefined && z !== undefined) {
       return [x, y, z];
     }
     return undefined;
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -11,6 +11,7 @@ import {
 import { AnalyticsEvent, triggerAnalyticsEvent } from "./utils/analytics";
 import { formatAsBulletList, getKeyFromName } from "./utils/data_utils";
 import { ManifestFile, ManifestFileMetadata, updateManifestVersion } from "./utils/dataset_utils";
+import { padCentroidsTo3d } from "./utils/math_utils";
 import * as urlUtils from "./utils/url_utils";
 
 import DataCache from "./DataCache";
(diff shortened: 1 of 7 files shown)


src/colorizer/Dataset.ts
```diff
@@ -625,12 +631,13 @@ export default class Dataset {
...
-    const index = objectId * 2;
+  public getCentroid(objectId: number): [number, number, number] | undefined {
+    const index = objectId * 3;
     const x = this.centroids?.[index];
     const y = this.centroids?.[index + 1];
-    if (x && y) {
-      return [x, y];
+    const z = this.centroids?.[index + 2];
+    if (x !== undefined && y !== undefined && z !== undefined) {
+      return [x, y, z];
```
frasercl, line 640: nit: with three of these here now, might want to just

```typescript
if (this.centroids === undefined || this.centroids === null) {
  return undefined;
}
```

the `undefined` checks for every individual component might still be good for checking if `objectId` was out of bounds (can't 100% tell whether that was or wasn't the original intent), but I might just check against `this.centroids.length` if so.

frasercl (approved):
looks good!

src/colorizer/Dataset.ts
```diff
@@ -625,12 +631,13 @@ export default class Dataset {
   /**
    * Returns the 2D centroid of a given object id.
```
rugeli, line 632: nit: update 2D to 3D in the comment as well?

rugeli (approved):
The handling of 3D vectors looks clean and straightforward! One non-blocking thought is that we're seeing repetitive code patterns when dealing with 3D vector components. We could either DRY them up for easier maintenance or leave as they are for readability.

PR #645, opened 2025-04-24
feat: Match 2D coloring with 3D coloring (3D pt. 7)

### LINKED ISSUE #636: 3D - Refactor 2D canvas to use same ID lookup rules as 3D
## Use Case

Currently, the 2D canvas requires image frames to have globally unique segmentation IDs.

#551 and #630 move to using a layer of indirection in 3D mode to remove this restriction. The 2D canvas should also use the same rules for consistency.

## Acceptance Criteria

- [x] Update shaders in the 2D canvas to take in the segmentation ID to global index lookup defined in #630.
- [x] Validate with a dataset that does not have remapped segmentation IDs (e.g. example nucmorph dataset)

## Details

ShrimpCryptid:
Problem
=======
Closes #636, "Refactor 2D canvas to use same ID lookup rules as 3D."

This change updates the 2D canvas shader to use the same lookup as the 3D canvas. It's almost identical to changes made here: https://github.com/allen-cell-animated/vole-core/pull/313/files#diff-feea930451be62e0d340bd6b10ac8c97cd4b5c542c9222e4b72ebdcc05bdbd5f

This allows both 2D and 3D datasets to use arbitrary segmentation IDs, which eliminates a lot of the preprocessing for 2D and 3D image data!

Solution
========
- Passed the global ID lookup texture data into the 2D colorize shader so it can map from local segmentation IDs to global IDs.
- Updated the 2D colorize shader to lookup global IDs with a helper method, `getId`.
- Fixed a bug in the ColorizeCanvas2D and ColorizeCanvas3D where colors were being passed as Linear-SRGB instead of SRGB.

## Type of change

Steps to Verify:
----------------
1. Open preview link (**Also, you'll need to be on VPN.**): https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-645/viewer?collection=https://dev-aics-dtp-001.int.allencell.org/peyton.lee/tfe-emt-nuclear-segmentations/data/tfe/collection.json&dataset=EMT+2D&feature=volume&keep-range=1&range=1999%2C70752&
2. Switch between the 2D and 3D datasets.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/7c3a4561-8602-48df-a09e-00b4d2905caa


src/colorizer/ColorizeCanvas2D.ts
```diff
@@ -362,7 +368,7 @@ export default class ColorizeCanvas2D implements IRenderCanvas {
   }
 
   private setOutlineColor(color: Color): void {
-    this.setUniform("outlineColor", color);
+    this.setUniform("outlineColor", color.clone().convertLinearToSRGB());
```
ShrimpCryptid, line 371: Fixes a bug where certain colors would look incorrect in the 2D and 3D viewports.

src/colorizer/shaders/colorize_RGBA8U.frag
```diff
@@ -84,6 +96,22 @@ vec4 getFloatFromTex(sampler2D tex, int index) {
...
 }
 
+uint getId(vec2 sUv) {
+  uint rawId = combineColor(texture(frame, sUv));
+  if (rawId == 0u) {
+    return BACKGROUND_ID;
+  }
+  uvec4 c = getUintFromTex(segIdToGlobalId, int(rawId - segIdOffset));
+  // Note: IDs are offset by `1` to reserve `0` for segmentations that don't
+  // have associated data. `1` MUST be subtracted from the ID when accessing
```
rugeli, line 106: Just curious: would it make sense to define -1 and 0 as constants rather than using literal values? Similar to `segIdOffset`

rugeli (approved):
LGTM!

ascibisz (approved):
I'm approving because it all looks good to me, and as you pointed out the shader changes are pretty much the same as some you have previously merged elsewhere!

But I do want to flag that I truly have no experience with shaders 😳 so if you have concerns about the shader code specifically, it may be good to get another set of eyes on it

src/colorizer/shaders/colorize_RGBA8U.frag
ShrimpCryptid: Nevermind! Changed it to a constant.

PR #648, opened 2025-04-29
refactor: Multi-value data for labels (new annotations pt. 3)

### LINKED ISSUE #459: Annotations - Multi-value labels for objects IDs / tracks
## Use Case

Requested by @chantelleleveille from a design discussion we had!

In addition to #439 (annotation labels for object IDs/tracks), it would be useful to capture custom inputs for different cell IDs and/or tracks. It could work with the existing annotation tab, only instead of binary label for cells, a user would be prompted with a custom text input.

It could likewise be exported into a CSV format.

Edit: Updated using input from @mfs4rd, who will also need this for upcoming EMT track validation.

## Acceptance Criteria

- [x] During creation, labels can be marked as a multi-value label via a checkbox in a pop-up menu. (TBD: Needs UX input)
- [x] When clicking/annotating, a text input will appear over the object with the last value auto-filled. Users can type to edit the contents of the textbox.
  - [x] Pressing enter or clicking off of the text input will update all objects in the last selection (track and/or range)
  - [ ] Consider having previously-used values appear below the text input, as with a dropdown/autocomplete.
- [x] When the CSV is exported, the values should appear in the column of the label.
- [ ] The annotation tab layout should change when a multi-value label is selected. (see below drawing)

## Details

Image

- Should all labels be multi-value? Is there a use case for just flagging a cell?

ShrimpCryptid:
Problem
=======
This change refactors the internals of `AnnotationData` to include values for labels! This will eventually close #459, "multi-value labels", once more UI is implemented.

Before, labels were only boolean/binary values, where an ID could be labeled or unlabeled only. Now, labeled IDs are associated with an additional string **value** which is specified when applying the labels.

I've introduced the concept of different label types, but for now only boolean labels can be chosen from the UI. Label types just inform how user interaction with the label works, but all labels behave the same internally.

*Estimated review size: medium, 20 minutes*

Solution
========
- Updated `AnnotationData` label setters to include a string `value`. An ID can only be associated with one value at a time, and can be reassigned.
- Added API for removing IDs from labels.
- Added additional `LabelData` options for types and behavior.
- Updated unit tests.

## Type of change

Steps to Verify:
----------------
1. Open the preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-648/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Small&feature=volume&track=77138&bg-key=max_intensity_zprojection_of_lamin_b1&t=15&tab=annotation
2. Create and select/deselect objects with an annotation label. All behavior should be the same as before.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/9415945d-db5d-4a83-ade0-d05c70eb16cc

### DIFF
--- a/src/colorizer/AnnotationData.ts
+++ b/src/colorizer/AnnotationData.ts
@@ -10,18 +10,32 @@ const CSV_COL_ID = "ID";
 const CSV_COL_TIME = "Frame";
 const CSV_COL_TRACK = "Track";
 
+export const BOOLEAN_VALUE = "TRUE";
+
 export const DEFAULT_ANNOTATION_LABEL_COLORS = KNOWN_CATEGORICAL_PALETTES.get(
   DEFAULT_CATEGORICAL_PALETTE_KEY
 )!.colorStops;
 
+export enum LabelType {
+  BOOLEAN = "boolean",
+  INTEGER = "integer",
+  CUSTOM = "custom",
+}
+
 export type LabelOptions = {
+  type: LabelType;
   name: string;
   color: Color;
+  autoIncrement: boolean;
 };
 
 export type LabelData = {
   options: LabelOptions;
+  valueToIds: Map<string, Set<number>>;
   ids: Set<number>;
+  lastValue: string | null;
+  // TODO: Store recently used values? Save values even if all IDs with them
+  // have been removed?
 };
 
 export interface IAnnotationDataGetters {
@@ -98,6 +112,10 @@ export interface IAnnotationDataGetters {
    */
   getNextDefaultLabelSettings(): LabelOptions;
 
+  getNextDefaultLabelValue(labelIdx: number): string;
+
+  getValueFromId(id: number, labelIdx: number): string | null;
+
   /**
    * Converts the annotation data to a CSV string.
    *
@@ -133,8 +151,8 @@ export interface IAnnotationDataSetters {
   createNewLabel(options?: Partial<LabelOptions>): number;
   setLabelOptions(labelIdx: number, options: Partial<LabelOptions>): void;
   deleteLabel(labelIdx: number): void;
-
-  setLabelOnIds(labelIdx: number, ids: number[], value: boolean): void;
+  setLabelValueOnIds(labelIdx: number, ids: number[], value: string): void;
+  removeLabelOnIds(labelIdx: number, ids: number[]): void;
   clear(): void;
 }
 
@@ -163,8 +181,12 @@ export class AnnotationData implements IAnnotationData {
     this.createNewLabel = this.createNewLabel.bind(this);
     this.deleteLabel = this.deleteLabel.bind(this);
     this.isLabelOnId = this.isLabelOnId.bind(this);
-    this.setLabelOnId = this.setLabelOnId.bind(this);
-    this.setLabelOnIds = this.setLabelOnIds.bind(this);
+    this.setLabelValueOnId = this.setLabelValueOnId.bind(this);
+    this.setLabelValueOnIds = this.setLabelValueOnIds.bind(this);
+    this.removeLabelOnId = this.removeLabelOnId.bind(this);
+    this.removeLabelOnIds = this.removeLabelOnIds.bind(this);
+    this.getNextDefaultLabelValue = this.getNextDefaultLabelValue.bind(this);
+    this.getValueFromId = this.getValueFromId.bind(this);
     this.toCsv = this.toCsv.bind(this);
   }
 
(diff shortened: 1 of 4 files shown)


src/colorizer/AnnotationData.ts
```diff
@@ -10,18 +10,32 @@ const CSV_COL_ID = "ID";
...
+  BOOLEAN = "boolean",
+  INTEGER = "integer",
+  CUSTOM = "custom",
+}
+
 export type LabelOptions = {
+  type: LabelType;
   name: string;
   color: Color;
+  autoIncrement: boolean;
```
ShrimpCryptid, line 29: The label's type is part of its options, which affect how the label is displayed and interacted with. Internally, all labels use the same data model.

src/colorizer/AnnotationData.ts
```diff
@@ -10,18 +10,32 @@ const CSV_COL_ID = "ID";
...
 };
 
 export type LabelData = {
   options: LabelOptions;
+  valueToIds: Map<string, Set<number>>;
```
ShrimpCryptid, line 34: I've left a couple comments on this in the code, but this makes the lookup for what values exist super simple, but make the lookup of a value from an ID potentially O(N). If we're having to do that operation a lot (like during rendering?) I may update the data model to include an `id->value` lookup.

src/colorizer/AnnotationData.ts
```diff
@@ -10,18 +10,32 @@ const CSV_COL_ID = "ID";
...
   ids: Set<number>;
+  lastValue: string | null;
+  // TODO: Store recently used values? Save values even if all IDs with them
+  // have been removed?
```
ShrimpCryptid, line 38: Will revisit once more UI is implemented! This would be for autocomplete features.

src/colorizer/AnnotationData.ts
```diff
@@ -10,18 +10,32 @@ const CSV_COL_ID = "ID";
 const CSV_COL_TIME = "Frame";
 const CSV_COL_TRACK = "Track";
 
+export const BOOLEAN_VALUE = "TRUE";
```
toloudis, line 13: Nit: how is this value written to csv?  it would be good if it were something easily parseable by json or Python like `true/false`, `True/False`, or `1/0`
Also I'm not sure whether you need an explicit "false" state to write to csv.  I realize that for other value types you need an empty "not set" kind of value and empty string fits the bill.

src/colorizer/AnnotationData.ts
toloudis: I kinda feel like autoincrement should be a LabelType

src/colorizer/AnnotationData.ts
toloudis: when do you need to do this kind of lookup?

src/colorizer/AnnotationData.ts
```diff
@@ -235,12 +257,55 @@ export class AnnotationData implements IAnnotationData {
...
+    }
+  }
+
   getNextDefaultLabelSettings(): LabelOptions {
     const color = new Color(
       DEFAULT_ANNOTATION_LABEL_COLORS[this.numLabelsCreated % DEFAULT_ANNOTATION_LABEL_COLORS.length]
     );
     const name = `Label ${this.numLabelsCreated + 1}`;
-    return { name, color };
+    return { type: LabelType.BOOLEAN, name, color, autoIncrement: true };
```
toloudis, line 287: Interesting how you have autoincrement true for this boolean but in the code right below here, you don't check the autoincrement flag in the boolean case!  (Booleans TRULY don't care about autoincrement!)

toloudis (approved):
LGTM.  Consider whether autoincrement could be its own type....

src/colorizer/AnnotationData.ts
ShrimpCryptid: Ah, yeah, I can do lowercase `true`. I'll ask Chantelle/Serge if an empty string is okay for boolean values in place of `false`, but it's easy to do either.

src/colorizer/AnnotationData.ts
ShrimpCryptid: The `id -> value` lookup is used in rendering, like if I want to show an annotation label onscreen and have it render the value (for example, an integer), I need to do a lookup from the ID to the value that's assigned to it. The inverse, `value -> id`, is more useful for deciding whether a value is still assigned to IDs.

src/colorizer/AnnotationData.ts
ShrimpCryptid: Updating here since we talked about this! I'm hesitant to make autoincrement a type for the specific case where we're importing a CSV as annotations back into the app. I can tell if a column is an integer or not, but I can't tell if it's intended to auto-increment.

rugeli (approved):
Looks good! I like how you separate helper functions and utils, with nice validation checks for id index. I did get a bit tripped up by the default values (`-1` for autoIncrement, `0` for non-autoIncrement, etc ) in a few places, but that might just be me.

PR #660, opened 2025-05-08
feat: Allow more internal paths

ShrimpCryptid:
Problem
=======
Scientists on new projects need to load internal-only datasets on TFE. This allows more internal paths to be loaded, while also updating error messaging in case a path cannot be loaded.

Solution
========
- Attempts to load any `/allen/aics` file path from the internal server.
- Updates error messaging to include reporting instructions for paths that may not be supported.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-660/
2. From the load menu, paste in this fake path. You should see additional error messaging guiding users to report the issue: `//allen/aics/not-a-real-path/manifest.json`.
3. Paste this new path instead. It should load without issues (if you are on the internal network): `//allen/aics/endothelial/morphological_features/timelapse_feature_explorer/20241120_20X_P0/cell_seg/manifest.json`

Screenshots (optional):
-----------------------

New error messaging:
image

### DIFF
--- a/src/colorizer/utils/url_utils.ts
+++ b/src/colorizer/utils/url_utils.ts
@@ -61,9 +61,7 @@ export enum UrlParam {
 const ALLEN_FILE_PREFIX = "/allen/";
 const ALLEN_PREFIX_TO_HTTPS: Record<string, string> = {
   // eslint-disable-next-line @typescript-eslint/naming-convention
-  "/allen/aics/assay-dev": "https://dev-aics-dtp-001.int.allencell.org/assay-dev",
-  // eslint-disable-next-line @typescript-eslint/naming-convention
-  "/allen/aics/microscopy": "https://dev-aics-dtp-001.int.allencell.org/microscopy",
+  "/allen/aics/": "https://dev-aics-dtp-001.int.allencell.org/",
 };
 
 export const DEFAULT_FETCH_TIMEOUT_MS = 2000;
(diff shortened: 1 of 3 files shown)


interim17 (approved):

src/colorizer/utils/url_utils.ts
```diff
@@ -61,9 +61,7 @@ export enum UrlParam {
 const ALLEN_FILE_PREFIX = "/allen/";
 const ALLEN_PREFIX_TO_HTTPS: Record<string, string> = {
   // eslint-disable-next-line @typescript-eslint/naming-convention
-  "/allen/aics/assay-dev": "https://dev-aics-dtp-001.int.allencell.org/assay-dev",
-  // eslint-disable-next-line @typescript-eslint/naming-convention
-  "/allen/aics/microscopy": "https://dev-aics-dtp-001.int.allencell.org/microscopy",
+  "/allen/aics/": "https://dev-aics-dtp-001.int.allencell.org/",
```
toloudis, line 64: is it possible that windows users will put in unc paths with backslashes?    Also sometimes I have seen a doubled forward slash before allen.
`//allen/aics/`
`\\\\allen\\aics\\`

toloudis (approved):

src/colorizer/utils/url_utils.ts
ShrimpCryptid: UNC paths are handled by the method `normalizeFilePathSlashes`! I tried a couple different combinations and double leading slashes work just fine.

The only configuration that didn't work was the quadruple-leading-backslash:

image

ShrimpCryptid: Fixed!

PR #668, opened 2025-05-14
fix: Change tab immediately on load

ShrimpCryptid:
Problem
=======
TFE stores the currently visible tab (and other viewer configuration) to the URL. However, currently these settings are not applied until after a dataset is loaded, which causes the tabs to jump suddenly.

This fix makes it so that state settings that aren't dataset-dependent (e.g. viewer config) are loaded and applied right away. It also includes a few small bugfixes.

Solution
========

with @pairperson1

## Type of change

Change summary:
---------------
* Something else I/we did

Steps to Verify:
----------------
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/Viewer.tsx
+++ b/src/Viewer.tsx
@@ -33,6 +33,7 @@ import { SCATTERPLOT_TIME_FEATURE } from "./components/Tabs/scatter_plot_data_ut
 import { DEFAULT_PLAYBACK_FPS, INTERNAL_BUILD } from "./constants";
 import { getDifferingProperties } from "./state/utils/data_validation";
 import {
+  loadInitialViewerStateFromParams,
   loadViewerStateFromParams,
   selectSerializationDependencies,
   serializeViewerState,
@@ -396,6 +397,8 @@ function Viewer(): ReactElement {
       if (isLoadingInitialDataset.current || isInitialDatasetLoaded) {
         return;
       }
+
+      loadInitialViewerStateFromParams(useViewerStateStore, searchParams);
 
       setIsDatasetLoading(true);
       setDatasetLoadProgress(null);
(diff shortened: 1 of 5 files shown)


src/Viewer.tsx
```diff
@@ -397,6 +398,8 @@ function Viewer(): ReactElement {
         return;
       }
 
+      loadInitialViewerStateFromParams(useViewerStateStore, searchParams);
```
ShrimpCryptid, line 401: Initial viewer state for non-dependent params is done before dataset starts to load in

src/Viewer.tsx
```diff
@@ -663,6 +666,8 @@ function Viewer(): ReactElement {
     },
   ];
   const tabItems = allTabItems.filter((item) => item.visible !== false);
+  const visibleTabKeys = useMemo(() => new Set(tabItems.map((item) => item.key)), [INTERNAL_BUILD]);
+  const currentTab = visibleTabKeys.has(openTab) ? openTab : TabType.TRACK_PLOT;
```
ShrimpCryptid, line 670: Fixes a bug where a link could specify a tab that isn't currently visible (due to internal vs. public build modes)

src/components/PlotWrapper.tsx
```diff
@@ -76,7 +76,9 @@ export default function PlotWrapper(inputProps: PlotWrapperProps): ReactElement
     }
     const width = plotDivRef.current.clientWidth;
     const height = plotDivRef.current.clientHeight;
-    plot?.setSize(width, height);
+    if (width && height) {
+      plot?.setSize(width, height);
+    }
```
ShrimpCryptid, line 81: Fixes a bug where the plot would crash after being given 0/undefined dimensions

ascibisz (approved):

toloudis (approved):

PR #728, opened 2025-07-11
feat: Controls for export dimensions

### LINKED ISSUE #445: Allow users to control export dimensions
## Use Case

Allow users to specify a target resolution for the exported video/images. Useful if a user has a specific resolution they need to export at for a video or presentation format!

Edit: Requested also by @chantelleleveille for the Endo project, which uses square frames. Currently, there is dead grey padding on either side of the viewport due to the different aspect ratios.

Image

## Acceptance Criteria

- [ ] Include a numeric input box in the Export modal for width and height (by default, this would be the current onscreen dimensions of the canvas when the export view is first opened).
  - [ ] Persist the chosen export dimensions when the user closes + reopens the export modal.
- [ ] Add two buttons to set the size to either the viewport (current viewport dimensions) or the frame (frame dimensions in dataset).
  - [ ] Frame should be disabled in 3D datasets.
- [ ] Add a lock toggle button icon that locks the aspect ratio.
- [ ] The exported image / video should use the selected dimensions, and the canvas should be reset once exporting is completed.

## Details

- Consider making this part of the state parameters?

ShrimpCryptid:
Problem
=======
Closes #445, "allow users to control export dimensions."

*Estimated review size: large, 30-40 minutes*

Solution
========
- Added controls for the dimensions of the exported frame size to the Export menu.
  - Users can choose presets for the current onscreen viewport size, or for the dataset's original frame dimensions.
  - A toggle button allows users to lock the current aspect ratio.
- The default filename now includes the exported video dimensions.
- Updated `IRenderCanvas.render()` to take an options object for clarity.
- Fixed unit tests.

## Type of change

Steps to Verify:
----------------
1. Open PR preview link here: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-728/viewer?collection=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2Fendothelial%2Fmorphological_features%2Ftimelapse_feature_explorer%2Fcollection.json
2. Go to **Export** and switch to exporting just a single frame.
3. Try varying the export dimensions or clicking the two buttons to jump to a preset size.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/1b862f46-7a39-4fa1-a57e-462c2793ed4d

image

### DIFF
--- a/src/colorizer/CanvasOverlay.ts
+++ b/src/colorizer/CanvasOverlay.ts
@@ -42,6 +42,13 @@ type OverlayRenderOptions = RenderOptions & {
 export type ExportOptions = {
   /** If true, enforces even pixel dimensions for the exported canvas. */
   enforceEven: boolean;
+  /** If true, shows the dataset name as a header in the exported canvas. */
+  showHeader: boolean;
+  /**
+   * If true, shows the legend in the footer (and the scale bar and timestamp
+   * if enabled) in the exported canvas.
+   */
+  showFooter: boolean;
 };
 
 /**
@@ -95,10 +102,6 @@ export default class CanvasOverlay implements IRenderCanvas {
    */
   private isExporting: boolean;
   private exportOptions: ExportOptions;
-  // TODO: Move these flags into export options instead of keeping them around
-  // as separate flags?
-  public isHeaderVisibleOnExport: boolean;
-  public isFooterVisibleOnExport: boolean;
   public isScaleBarVisible: boolean;
   public isTimestampVisible: boolean;
   public isAnnotationVisible: boolean;
@@ -165,9 +168,7 @@ export default class CanvasOverlay implements IRenderCanvas {
     this.footerSize = new Vector2(0, 0);
 
     this.isExporting = false;
-    this.exportOptions = { enforceEven: false };
-    this.isHeaderVisibleOnExport = true;
-    this.isFooterVisibleOnExport = true;
+    this.exportOptions = { enforceEven: false, showFooter: true, showHeader: true };
     this.isScaleBarVisible = true;
     this.isTimestampVisible = true;
     this.isAnnotationVisible = true;
@@ -464,8 +465,8 @@ export default class CanvasOverlay implements IRenderCanvas {
    */
   render(options?: OverlayRenderOptions): void {
     // Expand size by header + footer, if rendering:
-    const headerRenderer = this.getHeaderRenderer(this.isHeaderVisibleOnExport && this.isExporting);
-    const footerRenderer = this.getFooterRenderer(this.isFooterVisibleOnExport && this.isExporting);
+    const headerRenderer = this.getHeaderRenderer(this.exportOptions.showHeader && this.isExporting);
+    const footerRenderer = this.getFooterRenderer(this.exportOptions.showFooter && this.isExporting);
     this.headerSize = headerRenderer.sizePx;
     this.footerSize = footerRenderer.sizePx;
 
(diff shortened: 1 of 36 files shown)


src/colorizer/CanvasOverlay.ts
```diff
@@ -242,7 +256,11 @@ export default class CanvasOverlay implements IRenderCanvas {
   // TODO: Move `isExporting` flag into state
   public setIsExporting(isExporting: boolean): void {
     this.isExporting = isExporting;
-    this.render(false);
+    this.render({ renderInnerCanvas: false });
```
ShrimpCryptid, line 259: I also updated `render()` to take an options object too because I found that just the plain boolean argument wasn't very intuitive (and it actually means something different in the inner canvases)!

src/colorizer/CanvasOverlay.ts
```diff
@@ -358,14 +376,14 @@ export default class CanvasOverlay implements IRenderCanvas {
...
   }
 
   // Rendering functions ////////////////////////////
 
-  private getBaseRendererParams(): BaseRenderParams {
+  private getBaseRendererParams(overrideResolution?: Vector2): BaseRenderParams {
     return {
-      canvasSize: this.innerCanvasSize,
+      canvasSize: overrideResolution ?? this.innerCanvasSize,
```
ShrimpCryptid, line 386: Used when calculating export dimensions.

src/colorizer/CanvasOverlay.ts
```diff
@@ -445,17 +463,22 @@ export default class CanvasOverlay implements IRenderCanvas {
...
     // Update canvas resolution + size.
     const devicePixelRatio = getPixelRatio();
-    const baseCanvasWidthPx = toEven(this.innerCanvasSize.x);
-    const baseCanvasHeightPx = toEven(this.innerCanvasSize.y + this.headerSize.y + this.footerSize.y);
+
+    let baseCanvasWidthPx = this.innerCanvasSize.x;
+    let baseCanvasHeightPx = this.innerCanvasSize.y + this.headerSize.y + this.footerSize.y;
+    if (this.isExporting && this.exportOptions.enforceEven) {
+      baseCanvasWidthPx = toEven(baseCanvasWidthPx * devicePixelRatio) / devicePixelRatio;
+      baseCanvasHeightPx = toEven(baseCanvasHeightPx * devicePixelRatio) / devicePixelRatio;
```
ShrimpCryptid, line 480: I try to explain this elsewhere, but internally the canvas's internal resolution is determined by `baseCanvasWidthPx * devicePixelRatio`. This step here ensures that the final resulting canvas resolution will have even dimensions if it's a requirement.

src/components/Export.tsx
```diff
@@ -597,6 +732,105 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
               </Space>
             </MaxWidthRadioGroup>
           </Card>
+          <Card size="small" title={"Dimensions"}>
```
ShrimpCryptid, line 735: Image

src/components/Export.tsx
```diff
@@ -165,14 +195,83 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
   const [rangeMode, setRangeMode] = useState(RangeMode.ALL);
   const [customMin, setCustomMin] = useState(0);
   const [customMax, setCustomMax] = useState(props.totalFrames - 1);
-  const [imagePrefix, setImagePrefix] = useState(props.defaultImagePrefix);
+
+  const [useCurrentViewportSize, setUseCurrentViewportSize] = useState(true);
+  const [dimensionsInput, setDimensionsInput] = useState([1, 1]);
+  const [aspectRatio, setAspectRatio] = useState<number | null>(dimensionsInput[0] / dimensionsInput[1]);
```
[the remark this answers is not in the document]
ShrimpCryptid: As-is this will introduce a new bug where aspect ratio is reenabled every time `dimensionsInput` is changed. The current code will not have a bug since aspect ratio is adjusted when the modal is opened.

src/constants/index.ts
```diff
@@ -18,6 +18,8 @@ export const BACKDROP_OBJECT_OPACITY_MAX = 100;
 export const BACKDROP_OBJECT_OPACITY_DEFAULT = 50;
 export const COLOR_RAMP_RANGE_DEFAULT: [number, number] = [0, 0];
 
+export const CANVAS_ASPECT_RATIO = 14.6 / 10;
```
toloudis, line 21: does this need to be kept in sync with other magic numbers elsewhere?

toloudis (approved):

src/constants/index.ts
ShrimpCryptid: This is the only magic number now!

PR #731, opened 2025-07-15
fix: Fixed use of `label` tag, accessibility improvements in settings

### LINKED ISSUE #730: `SettingsItem` should not put interactive elements inside of a `label` HTML tag.
## Description

Currently, the `SettingsItem` component places all items inside of a `label` tag (implicit labeling).

https://css-tricks.com/html-inputs-and-labels-a-love-story/#aa-how-to-pair-a-label-and-an-input

This is not recommended for accessibility reasons, since it can be unclear what elements are being labeled. Additionally, this can cause unexpected behavior when there are other elements (besides the input) inside of the labeled area, like increasing the click target unexpectedly.

https://css-tricks.com/html-inputs-and-labels-a-love-story/#aa-dont-put-interactive-elements-inside-labels

(for example here, clicking ".mp4" will pull focus on the input text area)

Image

## Expected Behavior

- Add a `for` argument in the `SettingsItem` components, which takes in the ID of the labeled element.
- Change `SettingsItem` to not wrap all elements in a `label` tag, only the text label.
- Update CSS selectors in `SettingsContainer`.

ShrimpCryptid:
Problem
=======
Closes #730, "SettingsItem should not put interactive elements inside of a label HTML tag". 

It also fixes a couple related bugs, such as:
- Click targets for inputs were wider than the inputs themselves
- ColorPicker components could not be accessed using tab navigation
- Labels were not attached to dropdowns correctly

*Estimated review size: large, ~40 minutes (please turn off whitespace changes!)*

Solution
========
- `SettingsItem` no longer wraps both label text and inputs in a `label` component (implicit labeling)
  - Instead, users are expected to pass in an `htmlFor` attribute (explicit labeling)
- Updated several components (SelectDropdown, DropdownWithColorPicker, SpinBox, etc.) to accept and set an `id` prop.
- Fixed bug where inputs would have incorrect/wide click targets by updating all settings to use the `htmlFor` + `id` pattern.
- Fixed bug where SelectDropdown components were not using `label` HTML tags.
- Made a new component, `WrappedColorPicker`, which reimplements the color picker's target styling to fix several accessibility-related bugs.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-731/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Small&feature=volume&bg-key=max_intensity_zprojection_of_lamin_b1&t=15&tab=settings
3. Clicking on the white space to the right of the inputs will no longer cause them to become focused.
4. Use tab to navigate to a color picker, and space to click it. Tabbing should allow you to edit fields on the color picker popup.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/34be5e30-f2c0-46fb-a29d-a8015ec5b53b

-----------------------
1. `WrappedColorPicker.tsx`
2. `SettingsContainer.tsx`

### DIFF
--- a/src/components/CategoricalColorPicker.tsx
+++ b/src/components/CategoricalColorPicker.tsx
@@ -1,10 +1,9 @@
 import { Tooltip } from "antd";
-import React, { ReactElement, useMemo } from "react";
+import React, { ReactElement, useMemo, useRef } from "react";
 import styled from "styled-components";
 import { Color, ColorRepresentation } from "three";
 
 import { FlexRow, FlexRowAlignCenter } from "../styles/utils";
-import { DEFAULT_LABEL_COLOR_PRESETS } from "../utils/color_utils";
 
 import { useViewerStateStore } from "../state/ViewerState";
 import WrappedColorPicker from "./Inputs/WrappedColorPicker";
@@ -40,7 +39,8 @@ const ColorPickerContainer = styled(FlexRow)<{
     height: fit-content;
   }
 
-  & > div > span {
+  & > div > label,
+  & > div > label > span {
     // Text label, hide overflow as ellipsis
     white-space: nowrap;
     overflow: hidden;
@@ -53,6 +53,7 @@ export default function CategoricalColorPicker(inputProps: CategoricalColorPicke
 
   const categoricalPalette = useViewerStateStore((state) => state.categoricalPalette);
   const setCategoricalPalette = useViewerStateStore((state) => state.setCategoricalPalette);
+  const colorPickerContainerRef = useRef<HTMLDivElement>(null);
 
   const colorPickers = useMemo(() => {
     const elements = [];
@@ -64,21 +65,25 @@ export default function CategoricalColorPicker(inputProps: CategoricalColorPicke
         newPalette[i] = new Color(hex as ColorRepresentation);
         setCategoricalPalette(newPalette);
       };
+      const colorPickerId = `categorical-color-picker-${i}`;
 
       // Make the color picker component
       elements.push(
         <FlexRowAlignCenter key={i}>
           <WrappedColorPicker
+            id={colorPickerId}
             value={color.getHexString()}
             onChange={onChange}
             size={"small"}
             disabledAlpha={true}
-            presets={DEFAULT_LABEL_COLOR_PRESETS}
-            placement="bottom"
-            disablePopupContainer={true}
+            // Necessary to prevent the color picker from going off the screen
+            // TODO: Fix this? This prevents tab navigation from working.
+            getPopupContainer={undefined}
           />
           <Tooltip title={label} placement="top">
-            <span>{label}</span>
+            <label htmlFor={colorPickerId}>
+              <span>{label}</span>
+            </label>
           </Tooltip>
         </FlexRowAlignCenter>
       );
@@ -87,7 +92,7 @@ export default function CategoricalColorPicker(inputProps: CategoricalColorPicke
   }, [props]);
 
   return (
-    <ColorPickerContainer $itemGap="8px" $maxItemsPerRow="6" $itemWidth="110px">
+    <ColorPickerContainer $itemGap="8px" $maxItemsPerRow="6" $itemWidth="110px" ref={colorPickerContainerRef}>
       {colorPickers}
     </ColorPickerContainer>
   );
--- a/src/components/CategoricalColorPicker.tsx
+++ b/src/components/CategoricalColorPicker.tsx
@@ -76,8 +76,7 @@ export default function CategoricalColorPicker(inputProps: CategoricalColorPicke
             size={"small"}
             disabledAlpha={true}
             // Necessary to prevent the color picker from going off the screen
-            // TODO: Fix this? This prevents tab navigation from working.
-            getPopupContainer={undefined}
+            placement="right"
           />
           <Tooltip title={label} placement="top">
             <label htmlFor={colorPickerId}>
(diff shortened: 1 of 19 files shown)


src/components/Dropdowns/SelectionDropdown.tsx
```diff
@@ -195,14 +183,28 @@ export default function SelectionDropdown(inputProps: SelectionDropdownProps): R
...
   return (
     <FlexRowAlignCenter $gap={6}>
-      {props.label && <h3 id={id ?? props.htmlLabelId}>{props.label}</h3>}
+      {props.label && (
+        <label htmlFor={selectId}>
+          <h3 id={labelId}>{props.label}</h3>
+        </label>
+      )}
       <StyledSelect
-        aria-labelledby={id}
```
ShrimpCryptid, line 205: I was previously using `aria-labelledby` to try and connect the select input with the label, but from what I understand, `aria` doesn't interact with the HTML `label` system. `aria` is metadata that can help screen readers, but using `label` for form elements has direct functional implications.

ShrimpCryptid: You can test this change by clicking on the label text of any of the dropdowns. This should now cause the dropdown to be focused.

Image

src/components/Export.tsx
```diff
@@ -628,15 +654,19 @@ export default function Export(inputProps: ExportButtonProps): ReactElement {
                 </Button>
               </FlexRow>
             </SettingsItem>
-            <SettingsItem label={"Show feature legend"}>
-              <Checkbox
-                checked={showLegendDuringExport}
-                onChange={(e) => setShowLegendDuringExport(e.target.checked)}
-              />
+            <SettingsItem label={"Show feature legend"} htmlFor={HtmlIDs.EXPORT_MODAL_SHOW_FEATURE_LEGEND_CHECKBOX}>
+              <div style={{ width: "fit-content" }}>
```
ShrimpCryptid, line 658: Prevents the checkbox target from taking up the entire horizontal real estate

src/components/Inputs/WrappedColorPicker.tsx
```diff
@@ -0,0 +1,116 @@
+import { Button, ColorPicker, ColorPickerProps } from "antd";
```
ShrimpCryptid, line 1: Convenience component that wraps Ant's `ColorPicker` component and fixes several accessibility issues with it. The change was motivated by me needing to assign an `id` to a `ColorPicker` for labeling, which Ant doesn't support without changing its triggering button component.

Image

src/components/SettingsContainer.tsx
```diff
@@ -32,15 +35,29 @@ export function SettingsItem(inputProps: PropsWithChildren<Partial<SettingsItemP
...
+    if (props.label && typeof props.label === "string" && !props.htmlFor) {
+      console.warn(
+        "SettingsItem: Please set the 'htmlFor' attribute to support screen readers in setting '" + props.label + "'."
+      );
+    }
+  }, [props.label, props.htmlFor]);
+
+  let labelElement = <div></div>;
+  if (props.label) {
+    const formattedLabel = props.labelFormatter ? props.labelFormatter(props.label) : props.label;
```
ShrimpCryptid, line 48: I don't think this is used anywhere, honestly. I was debating whether to remove it in this PR

src/components/SettingsContainer.tsx
```diff
@@ -32,15 +35,29 @@ export function SettingsItem(inputProps: PropsWithChildren<Partial<SettingsItemP
...
-    <label style={props.style}>
-      <span style={props.labelStyle} id={props.id}>
-        {formattedLabel}
-      </span>
+    <div style={props.style} className={SETTINGS_ITEM_CLASS}>
+      {labelElement}
       {props.children}
-    </label>
+    </div>
   );
```
ShrimpCryptid, line 44: Scribbly before/after explanation!

Image

src/components/CategoricalColorPicker.tsx
```diff
@@ -62,13 +64,25 @@ export default function CategoricalColorPicker(inputProps: CategoricalColorPicke
...
+          <WrappedColorPicker
+            id={colorPickerId}
+            value={color.getHexString()}
+            onChange={onChange}
+            size={"small"}
+            disabledAlpha={true}
+            // Necessary to prevent the color picker from going off the screen
+            // TODO: Fix this? This prevents tab navigation from working.
+            getPopupContainer={undefined}
+          />
```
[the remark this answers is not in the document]
ShrimpCryptid: Yeah, it just means that I wasn't able to fix this for this specific instance (the color pickers that appear when you have a categorical feature selected)

toloudis (approved):
LGTM but I really question whether clicking on the label of a switch control should actually flip the switch...  macos certainly doesn't behave that way AFAIK.    

There is a lot of good work on components here in TFE that makes me wonder if we could start a reusable component library from it.

ShrimpCryptid:
> As a user, I would not expect clicking on a Label to do anything at all. It looks like a form to me, and I'd expect tabbing to tab through all the input elements and skip the labels. Is my take controversial?

There's two parts to this! For the first, clicking on a label to focus/interact with a form input is generally standard. See MDN reference here: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label

The second isn't controversial at all, and the `label` is not focused during tab navigation.

tabbing

Issue #747, opened 2025-08-21
Load ZIP files


ShrimpCryptid:
## Use Case

Users following our data processing tutorial in the `colorizer-data` repo in Google Colab (and running locally) don't have a convenient way of viewing their datasets in TFE.

We previously explored #376, "view local datasets/files", but there were some potential UX issues with it due to slow speeds when loading files on network drives (see #379).

A compromise would be to allow users to upload .ZIP files, containing all of their dataset or collection files. Users could download the zipped folder from Colab (or zip it locally) and then open it in TFE. This will also hopefully fix the bandwidth issue too!

## Acceptance Criteria

- [ ] In the Load menu, add a button that opens a file picker and allows users to select a `.zip` file.
- [ ] TFE should look for and load a `collection.json` or `manifest.json` within the zip archive.
  - [ ]  If there are nested folders, it should search for the files in the first non-empty folder.
- [ ] If the user refreshes the page or shares the link, they should be shown a prompt to upload + open the file.
  - [ ] Store a URL parameter that reflects that a ZIP file is open.
- [ ] Update error messaging to reflect URL vs. files.
- [ ] Optional, show an icon next to the collection name that indicates if it's from a ZIP file.

## Details

_Please provide any helpful specifications_

jennifermccarty:
Per Peyton - ~50% complete.

PR #748, opened 2025-08-25
feat: Improve error messaging during ZIP loading (ZIP pt. 2)

ShrimpCryptid:
Problem
=======
Part 2 of 3(?) for #747, "Load ZIP files".

This change allows Collections to store information about the type of data they're loading from (ZIP file vs. url) and improves error messaging when loading ZIP files.

*Estimated review size: medium, 20-25 minutes*

Solution
========
- Refactored `Collection` constructor to take in a `CollectionConfig`, which includes metadata about the source path and type.
- Updated error messaging to reflect the type of loaded file (ZIP vs. URL).
  - Added specific error messaging for when expected `collection.json` or `manifest.json` files are missing.
- Fixed a bug where the recent datasets dropdown would have empty items caused by loading ZIP files.

## Type of change

Steps to Verify:
----------------
1. Download the test ZIP files here:
 - [malformed-json-dataset.zip](https://github.com/user-attachments/files/21976514/malformed-json-dataset.zip)
 - [malformed-collection-and-dataset.zip](https://github.com/user-attachments/files/21976515/malformed-collection-and-dataset.zip)
 - [missing-files.zip.zip](https://github.com/user-attachments/files/21976545/missing-files.zip.zip)

2. Open the preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-748/
3. Load each of the ZIP files. You should see different error messages for each, based on what files were missing or malformed.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/78b75455-6f47-4c88-af12-8b825eb302c9

image

image

-----------------------
1. `Collection.ts`

### DIFF
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -39,22 +39,41 @@ export const enum CollectionSourceType {
   ZIP_FILE,
 }
 
+/**
+ * Optional configuration for the Collection, including options for how paths should be
+ * resolved and metadata about the collection source.
+ */
 export type CollectionConfig = {
   /**
    * Source path of the collection file, if one was used to load this
    * Collection. This can be a URL or a file path. `null` if there was no source
    * collection file, and the collection was an auto-generated wrapper around a
    * dataset.
+   *
+   * When loaded from a local collection file, this should be `collection.json`.
    */
   sourcePath?: string | null;
   /**
-   * Filename that was used to load this Collection.
+   * Name of the file that the Collection was loaded from, if loaded from a file
+   * source instead of a URL.
+   *
+   * For example, a ZIP file `MyCollection.zip` containing a `collection.json`
+   * at the root directory should have:
+   * - `sourceFilename` set to `MyCollection.zip`
+   * - `sourcePath` set to `collection.json`.
    */
   sourceFilename?: string | null;
   /** The type of Collection source. This can be an uploaded ZIP file or a URL. */
   sourceType?: CollectionSourceType;
   pathResolver?: IPathResolver;
 };
+
+const makeDefaultCollectionConfig = (): Required<CollectionConfig> => ({
+  sourcePath: null,
+  sourceFilename: null,
+  sourceType: CollectionSourceType.URL,
+  pathResolver: new UrlPathResolver(),
+});
 
 export type CollectionLoadOptions = {
   pathResolver?: IPathResolver;
@@ -84,16 +103,25 @@ export default class Collection {
    * such as when generating dummy Collections for single datasets.
    */
   public readonly sourcePath: string | null;
-  /** Optional. Name of the file used to open this Collection, if loaded from a file source.*/
+  /**
+   * Name of the file containing this Collection, if loaded from a
+   * file source.
+   */
   public readonly sourceFilename: string | null;
   public readonly sourceType: CollectionSourceType;
 
   /**
    * Constructs a new Collection from a CollectionData map.
-   * @param entries A map from string keys to CollectionEntry objects. The `path` of all
-   * entries MUST be the absolute path to the manifest JSON file of the dataset.
-   * @param source the optional string url representing the source of the Collection. `null` by default.
-   * @throws an error if a `path` is not a URL to a JSON resource.
+   * @param entries A map from string keys to CollectionEntry objects. The
+   * `path` of all entries MUST be the absolute path to the manifest JSON file
+   * of the dataset.
+   * @param config Optional configuration for the Collection, including path
+   * resolution and metadata about the collection source. By default, the source
+   * will be assumed to be from a URL.
+   * @param metadata Optional metadata, as loaded from a collection JSON
+   * manifest.
+   * @throws an error if any `path` in `entries` is not a valid path to a JSON
+   * resource.
    */
   constructor(
     entries: CollectionData,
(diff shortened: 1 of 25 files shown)


src/colorizer/types.ts
```diff
@@ -319,12 +319,16 @@ export enum LoadTroubleshooting {
     " Please check your network access.",
   CHECK_FILE_EXISTS = "Please check if the file exists and if you have access to it, or see the developer console for more details.",
   CHECK_FILE_OR_NETWORK = "This may be because of an unsupported format, missing files, or server and network issues. Please see the developer console for more details.",
+  CHECK_ZIP_FORMAT = "Make sure a 'collection.json' or 'manifest.json' file exists in the root directory.",
+  CHECK_ZIP_FORMAT_COLLECTION = "A 'collection.json' should exist in the root directory.",
+  CHECK_ZIP_FORMAT_MANIFEST = "A 'manifest.json' should exist in the root directory.",
```
ShrimpCryptid, line 324: *TECHNICALLY* this file doesn't need to be in the root directory, it just has to be in the highest folder in the hierarchy that contains files (and not just another nested directory). I'm not sure if there's a good way to word this, so for now "root directory" seems fine.

src/colorizer/Collection.ts
```diff
@@ -419,9 +503,18 @@ export default class Collection {
...
+      collectionLoadError.message.includes(LoadTroubleshooting.CHECK_ZIP_FORMAT_COLLECTION) &&
+      datasetLoadError.message.includes(LoadTroubleshooting.CHECK_ZIP_FORMAT_MANIFEST)
+    ) {
+      return new Error(LoadErrorMessage.ZIP_BOTH_UNREACHABLE + " " + LoadTroubleshooting.CHECK_ZIP_FORMAT);
     } else {
       // Format and return a message containing both errors.
-      console.error(`URL '${url}' could not be loaded as a collection or dataset.`);
+      console.error(
+        `${
+          type === CollectionSourceType.URL ? `URL '${path}'` : `ZIP file`
```
ShrimpCryptid, line 515: Change all instances of "ZIP" to ".zip"

meganrm (approved):
Consider tests

ShrimpCryptid:
> Consider tests

What if I don't want to consider 😔

ShrimpCryptid:
> Consider tests

Okay I considered, it took a lot of mocking and ended up being a relatively large change. I've requested you over on #752.

src/colorizer/types.ts
frasercl: "base" maybe?

src/colorizer/types.ts
```diff
@@ -319,12 +319,16 @@ export enum LoadTroubleshooting {
...
 }
 
 export enum LoadErrorMessage {
```
frasercl, line 327: Could be a `const enum`? Benefits and drawbacks:

- benefit: enum is inlined rather than becoming a weird object in the final code
- drawback: these strings are all pretty large, if one or two are used twice the cost of inlining them probably overrides the weird object thing.

Probably best to keep these non-`const` for now, but I have a potentially unjustified vendetta against non-`const` enums. See next comment for how they might make sense anyways?

src/colorizer/Collection.ts
```diff
@@ -419,9 +503,18 @@ export default class Collection {
       datasetLoadError.message.includes("404 (Not Found)")
     ) {
       return new Error(LoadErrorMessage.BOTH_404);
+    } else if (
+      collectionLoadError.message.includes(LoadTroubleshooting.CHECK_ZIP_FORMAT_COLLECTION) &&
```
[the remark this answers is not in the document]
frasercl: Building on this...
1. I'd make a separate `Error`-extending class for loading failures like this (Or maybe two for `Collection` and `Dataset`?) Checking the error *message* to determine the type of error isn't awesome &mdash; with your own error type, you can give it your own `type` field.
2. In fact, I might consider making types like `LoadErrorMessage` `const enum`s with numeric values, using that for a `type` field, and creating an object and/or a function that maps them to their respective messages.
3. To what extent could errors with a `type` field let you error early over in `loadFromAmbiguousResource`, rather than trying again as a `Dataset` and having to juggle two errors? (Very possible you've investigated this and the answer is "not much at all.")
4. Totally optional, but I tend to think `try`/`catch`/`throw` should be used as close to the application level as possible, and that methods on these sorts of collection types should instead represent failure in their return types. Consider returning a `Collection | SomeErrorEnum` and `throw`ing lower down? Could potentially solve the error messaging encapsulation problem as well.

frasercl (approved):

src/colorizer/Collection.ts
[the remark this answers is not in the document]
ShrimpCryptid: Wrote this up as a new ticket! https://github.com/allen-cell-animated/timelapse-colorizer/issues/763

src/colorizer/Collection.ts
```diff
@@ -34,6 +34,36 @@ export type DatasetLoadResult =
...
+export type CollectionConfig = {
+  /**
+   * Source path of the collection file, if one was used to load this
+   * Collection. This can be a URL or a file path. `null` if there was no source
+   * collection file, and the collection was an auto-generated wrapper around a
+   * dataset.
+   *
+   * When loaded from a local collection file, this should be `collection.json`.
+   */
+  sourcePath?: string | null;
```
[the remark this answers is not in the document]
ShrimpCryptid: Tried this... I still want `sourcePath` to be an optional field in Config, but when set I want it to map directly to the final value of `sourcePath` in `Collection` (which is `null`) so I'm leaving it for now.

PR #750, opened 2025-08-27
refactor: Moved hooks out of `colorizer` (ZIP pt. 3a)

ShrimpCryptid:
Problem
=======
The `colorizer` folder was mainly intended for the data model side of TFE and was not intended to have any React components, but has had a steadily growing `react_utils.ts` file containing React hooks. This change moves all of the hooks into a new `src/hooks` directory, and places each hook into its own file.

I made no major changes besides moving file contents around, so hopefully an easy rubber stamp! (Edit: I did go back in and rename a few variables in hooks.)

Solution
========
- Moved hooks in `react_utils.ts` into individual files in `src/hooks`.
- Updated imports.

Steps to Verify:
----------------
1. Open preview link: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-750/
2. Load a dataset and click around.


src/hooks/useAnnotations.ts
```diff
@@ -1,270 +1,10 @@
...
-import { useShortcutKey } from "./hooks";
+import { AnnotationSelectionMode, Dataset } from "../colorizer";
+import { useConstructor } from "./useConstructor";
+import { useShortcutKey } from "./useShortcutKey";
 
-import { AnnotationData, AnnotationMergeMode, IAnnotationDataGetters, IAnnotationDataSetters, LabelType } from "../AnnotationData";
-import Dataset from "../Dataset";
-
-// TODO: Move this to a folder outside of `colorizer`.
-// TODO: Split this up into multiple files.
```
ShrimpCryptid, line 12: This change completes these two TODOS!

src/hooks/useConstructor.ts
```diff
@@ -0,0 +1,32 @@
+import { MutableRefObject, useRef } from "react";
+
+/**
+ * Returns a reference to a constructed value that will not be re-computed
+ * between renders.
```
toloudis, line 5: For a React novice like me, a nice additional comment would be "why would I want to use this hook?".  If the important points are that (1) it's not re-computed between renders, and (2) it guarantees non-null, why would I want those features?

toloudis (approved):
Nice!

Issue #617, opened 2025-04-02
Distinguish linear/diverging/cyclical color ramps


ShrimpCryptid:
## Use Case

Spin-off from #613.

![Image](https://github.com/user-attachments/assets/66131a9a-90cd-4104-9f0d-55b11191330a)

Currently, all of the color ramps are in one vertical list. Some users were confused about what the cyclical color ramps did, because from an initial glance it's not obvious that the endpoints of the ramp join.

UX discussed a couple of solutions, but one idea was to widen the dropdown area into three columns, labeled "linear," "diverging," and "cyclical".

## Acceptance Criteria

- [ ] Define a color ramp type on the existing color ramps.
- [ ] The ColorRampDropdown divides the color ramps by type into columns.

## Details

_Please provide any helpful specifications_

PR #766, opened 2025-09-12
feat: Added gradient to color ramp range slider (layout pt. 3)

### LINKED ISSUE #221: feature: Divergent map controls + color ramp preview
This is just a thought I had for improving the color map and colored feature slider.  This thought was inspired by the need to control a diverging color map, or potentially position a particular color at a particular value in the middle.

Render the color gradient directly below the slider, perfectly aligned with the min and max.
So when the slider looks like this, assuming color goes from red to green:
```
min                                 max
|-------o------------------o--------|
red    red    varying     green   green
```
And have this thing update properly so it is always in sync and consistent. 
Next phase would be to add a little dragable handle on the color map to push the middle interpolation color, like this:
```
[red-------------orange-------------yellow]
                   (^)
```
If you drag the handle left, the red-orange is still linearly interpolated but it reaches orange faster. And the orange-yellow is still linearly interpolated but it reaches yellow more gradually.  

The point here is not to allow custom gradient builders just yet, but to:
(1) clearly line the gradient up with the numeric values
(2) give some freedom to color ranges above and below a given value.

This would not need to change the colormap dropdown picker at all.  It would be an additional component closely aligned with the colored feature values slider.

ShrimpCryptid:
Problem
=======
Closes #221, "Divergent map controls + color ramp preview". This change moves the feature range slider into its own component (`ColorRampRangeSlider.tsx`), and shows the current color ramp as a gradient across it.

Solution
========
- Moved the feature range slider into its own component (`ColorRampRangeSlider.tsx`).
- Replaced the keep feature range checkbox with a toggleable button.
- Added a CSS gradient to the slider to show how the currently-selected color ramp will be applied.
- Added a midpoint marker when a diverging ramp is selected to show the middle (neutral) value.

## Type of change

Steps to Verify:
----------------
1.  Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-766/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Small&feature=volume&bg-key=max_intensity_zprojection_of_lamin_b1&t=15&color=matplotlib-cool&keep-range=0&range=563.818%2C1523.818
2. Play with the color ramp slider! Switch color ramps, reverse the current color ramp, and choose different (and diverging) ramps. Refresh the page to ensure that the changes are saved.
3. Toggle on the range lock and switch datasets.

Screenshots (optional):
-----------------------
image

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/2ab74ea3-f404-46c0-91b3-cc0cb8aede9e


src/colorizer/colors/color_ramps.ts
```diff
@@ -14,7 +14,7 @@ export type RawColorData = {
    * Most commonly used for glasbey color ramp visualizations, where lots
    * of colors are needed to distinguish between many different values.
    */
-  categorical?: boolean;
+  type?: ColorRampType;
```
ShrimpCryptid, line 17: Added types field so I can have the color ramp slider show the midpoint when a diverging ramp is selected. This may eventually also help support #617.

Image

src/colorizer/ColorRamp.ts
```diff
@@ -61,20 +62,21 @@ export default class ColorRamp {
...
         ctx.fillStyle = `#${color.getHexString()}`;
         ctx.fillRect(Math.floor(step * idx), 0, Math.ceil(step), height);
       });
+    } else {
+      // All other ramp types are linear gradients
+      const gradientWidth = vertical ? 0 : width;
+      const gradientHeight = vertical ? height : 0;
+      const gradient = ColorRamp.linearGradientFromColors(ctx, this.colorStops, gradientWidth, gradientHeight);
+      ctx.fillStyle = gradient;
+      ctx.fillRect(0, 0, width, height);
```
ShrimpCryptid, line 79: Flipped around this conditional so that the Linear case is in the else branch, but all code is unchanged.

meganrm (approved):

src/components/Controls/ColorizeControl/ColorRampRangeSlider.tsx
```diff
@@ -0,0 +1,189 @@
...
+ */
+function getTrackCssGradient(
+  min: number,
+  max: number,
+  minRange: number,
+  maxRange: number,
+  colorRamp: ColorRamp
+): string {
+  // Determine the step size between color stops, in percentage space.
+  const range = max - min;
```
rugeli, line 37: The step calculation below assumes a non-zero range, is there a guard for `max === min` case that i might've missed?

rugeli (approved):
LGTM! I jumped back and forth while reviewing and was able to find the answer I was wondering about. One small observation from playing around with the slider: when the min and max points overlap, dragging one over the other makes the value tooltip flicker in a funky way. (not urgent to fix since I don’t think users would normally do this)

Issue #619, opened 2025-04-03
3D pt. 2 - Show other volume channels in addition to segmentations.


ShrimpCryptid:
## Use Case

Scientists want to validate segmentations against the original 3D data. One way to do this is allowing other channels to be included in the dataset. 

The analogue with our 2D datasets is our backdrop images, which allow you to compare 2D segmentations against original data.

## Acceptance Criteria

- [x] Treat additional channels in the Zarr (not designated as the segmentation channel) as backdrop channels.
- [x] Allow backdrop channels to be toggled on/off from the onscreen backdrop button and the settings tab.
- [x] Remap existing backdrop controls per channel (brightness, saturation(?), segmentation opacity)
  - [x] Update settings UI 

## Details

Questions for UX:
- What controls do users need for additional channels?
  - Do users need more than one backdrop channel to be enabled at a time?
  - Global vs. specific to a channel?
  - Coloring/LUT/etc.
- What existing controls for backdrops can we reuse?
- Does "backdrops" still make sense for these additional channels?

ShrimpCryptid:
Potential spec for backdrop channels:
```
{
  source: "https://something-else.url/ome.zarr",
  segmentationChannel: 0,
  totalFrames: 0,
  backdrops: [
     {
        source: "https://something.url/ome.zarr",
        name: "My Channel",
        description: "Some description of channel content",
        channelIndex: 0,
     }
  ]
}
```

ShrimpCryptid:
Initial attempts at blending backdrop + segmentation channels are rendering but are not super useful. Consider writing a new volume renderer for handling segmentations.

jennifermccarty:
Per Peyton - 60-70% done.

PR #780, opened 2025-10-01
feat: Default channel palettes vary by channel quantity (channels pt. 4)

ShrimpCryptid:
Problem
=======
Part 4 of 5 for #619, "Show other channels in addition to segmentations."

Solution
========
- Added default colors for background channels, which vary based on the number of channels in the dataset.
  - 1 channel: white
  - 2 channels: magenta, green
  - 3 channels: magenta, yellow, cyan

## Type of change

Steps to Verify:
----------------
Testing URL saving
1. Open the preview link:
2. You should see the following settings in the channel settings:
3. 

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
4. other important file

Thanks for contributing!

### DIFF
--- a/src/state/slices/channel_slice.ts
+++ b/src/state/slices/channel_slice.ts
@@ -4,11 +4,47 @@ import { StateCreator } from "zustand";
 import { ChannelRangePreset } from "../../colorizer/types";
 import { SerializedStoreData, SubscribableStore } from "../types";
 import { addDerivedStateSubscriber } from "../utils/store_utils";
+import { CollectionSlice } from "./collection_slice";
 import { DatasetSlice } from "./dataset_slice";
+
+import { Backdrop3dData } from "../../colorizer/Dataset";
+
+const WHITE = new Color(1, 1, 1);
+const MAGENTA = new Color(1, 0, 1);
+const CYAN = new Color(0, 1, 1);
+const YELLOW = new Color(1, 1, 0);
+const GREEN = new Color(0, 1, 0);
+
+const ONE_CHANNEL_PALETTE = [WHITE];
+const TWO_CHANNEL_PALETTE = [MAGENTA, GREEN];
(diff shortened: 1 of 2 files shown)


src/state/slices/channel_slice.ts
```diff
@@ -4,11 +4,47 @@ import { StateCreator } from "zustand";
...
+  } else if (totalChannels === 2) {
+    return TWO_CHANNEL_PALETTE[index] ?? WHITE;
+  } else {
+    return THREE_CHANNEL_PALETTE[index] ?? WHITE;
+  }
+}
+
+function getDefaultChannelSetting(index: number, totalChannels: number, backdropData?: Backdrop3dData): ChannelSetting {
+  return {
+    visible: index < 3,
```
interim17: Three visible channels is the limit? Maybe put the int 3 in a constant, and just confirming that `getDefaultColorForChannel` doesn't need another case for `totalChannels > 3`

ShrimpCryptid: If there are more than 3 channels, they'll be assigned white as a default.

ShrimpCryptid: It would probably be interesting to have more colors we default to in the future? But for now behavior is only specified for the first three.

src/state/slices/channel_slice.ts
```diff
@@ -56,32 +92,39 @@ export const createChannelSlice: StateCreator<ChannelSlice, [], [], ChannelSlice
...
+  );
+
   // When the dataset updates, create a number of default channel settings equal
-  // to the number of backdrop channels in the dataset.
+  // to the number of backdrop channels in the dataset. Preserve existing settings
+  // for channels that currently exist.
   addDerivedStateSubscriber(
     store,
     (state) => ({ dataset: state.dataset }),
     ({ dataset }) => {
```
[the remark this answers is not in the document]
ShrimpCryptid: Probably not if only one parameter is being passed in!

PR #850, opened 2026-02-02
fix: Encode special characters in internal file paths

### LINKED ISSUE #412: `/allen` paths with spaces or special characters fail to load
## Description

Goutham ran into a bug where the viewer failed to load datasets submitted by a PairUp scientist. The names were something like: `feature_manifest_1 uM Capsaicin BTX_Stress_NTG F#1.1_aligned_maxproject.csv`

He was able to fix it by removing both the spaces and the `#` symbols.

## Expected Behavior

This is kind of arguable, because the `/allen` path is only in use internally. It seems like the server we're using knows how to unescape the characters (so `https://dev-aics-dtp-001.int.allencell.org/assay-dev/users/Goutham/pairup_chris_results/colorizer_outputs/feature_manifest_1%20uM%20Capsaicin%20BTX_Stress_NTG%20F%231.1_aligned_maxproject.csv%2Fmanifest.json`) worked fine.

We should attempt to escape the URL characters before loading the file.

## Reproduction

## Environment

ShrimpCryptid:
Problem
=======
Closes #412, "`/allen` paths with spaces or special characters fail to load."

Solution
========
- Fixed a bug where paths with special characters (spaces, hashes, etc.) could not be loaded.
- Added unit tests.

## Type of change

Steps to Verify:
----------------
1. Copy the Vast path of a test dataset: 
```txt
/allen/aics/users/peyton.lee/test-data/My Test Data/Special Folder #15934/manifest.json
```
2. Open the (bugged) public build: https://timelapse.allencell.org/
3. Click **Load**, paste in the test dataset path, and attempt to load it. The load will fail due to a 404 error.
4. Open the fixed preview build: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-850/
5. Repeat step 3. The dataset will load.

### DIFF
--- a/tests/url_utils.test.ts
+++ b/tests/url_utils.test.ts
@@ -90,7 +90,7 @@ describe("convertAllenPathToHttps", () => {
     );
   });
 
-  it("replaces special characters", () => {
+  it("Replaces special characters", () => {
     expect(convertAllenPathToHttps("/allen/aics/some_resource/&*#$@^~!.json")).to.equal(
       `${VAST_FILES_URL}some_resource/%26*%23%24%40%5E~!.json`
     );


frasercl (approved):

meganrm (approved):
looks great!

PR #875, opened 2026-02-12
refactor: Store 2D/3D view mode flag in state

ShrimpCryptid:
Problem
=======
Previously, several parts of the UI would respond based on whether the dataset was 3D or 2D. However, they all were performing versions of the same logic in many different places, which meant that very simple components (like a button toggle) needed to know the high-level logic of whether a dataset is considered 3D.

This PR adds a derived flag to state that updates based on whether the dataset is 3D or 2D. This also adds the possibility in the future of supporting switching between 2D and 3D versions of the same dataset in the future.

Solution
========
- Added `viewMode` enum to `dataset_slice`, which updates on Dataset change based on whether the Dataset has 3D data.
- Updated current checks for `dataset.has3dFrames` to use `viewMode` state value instead.

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-875/viewer?collection=https%3A%2F%2Fraw.githubusercontent.com%2Fallen-cell-animated%2Ftimelapse-colorizer%2Frefs%2Fheads%2Ftest-datasets%2Fdocs%2Fnucmorph-zarr%2Fcollection.json
2. Check that the image button shows the Channels UI when hovered:
image

3. Change the dataset to "Small 2D".
4. Check that the image button shows the backdrops UI when hovered:
image


ascibisz (approved):

toloudis (approved):
Its kind of hard to tell how big an improvement this is, as it looks about the same amount of code, just checking different things.   But it does seem a bit cleaner and more specific. LGTM!!

PR #902, opened 2026-04-06
fix: Render to pick buffer in render step (points pt. 1)

ShrimpCryptid:
Problem
=======
Fixes a bug where each call to `getIdAtPixel` would cause a re-render of the pick buffer (eek!). Certain features like annotations would call `getIdAtPixel` for each onscreen object, potentially causing hundreds of re-renders each frame.

Now, the pick buffer is rendered to once during the `render` step. (I also threw in one more fix and changed some state dependencies.)

Solution
========
- The pick buffer is now updated once on render instead of on every call to `getIdAtPixel()`.
- `Viewer` no longer re-renders on every frame change (removed dependency on `currentFrame`).

## Type of change

Steps to Verify:
----------------
1. Download this CSV: [Exploratory dataset-annotations (8).csv](https://github.com/user-attachments/files/26521387/Exploratory.dataset-annotations.8.csv)
2. Open https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-902/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fbaseline_colonies_dataset%2Fcollection.json&t=15&tab=annotation and import the CSV annotations.
3. Play through time. (It may be slightly choppy on first playback but will be much faster on subsequent runs.)

### DIFF
--- a/src/colorizer/viewport/ColorizeCanvas2D.ts
+++ b/src/colorizer/viewport/ColorizeCanvas2D.ts
@@ -801,7 +801,15 @@ export default class ColorizeCanvas2D implements IInnerRenderCanvas {
     this.checkPixelRatio();
     this.syncTrackPathLine();
 
+    // Render main scene
     this.renderer.render(this.scene, this.camera);
+
+    // Render to pick buffer
+    const previousRenderTarget = this.renderer.getRenderTarget();
+    this.renderer.setRenderTarget(this.pickRenderTarget);
+    this.renderer.render(this.pickScene, this.camera);
+    this.renderer.setRenderTarget(previousRenderTarget);
+
     this.onRenderCallback?.();
   }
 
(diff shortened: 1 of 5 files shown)


src/colorizer/viewport/overlays/elements/annotations.ts
```diff
@@ -154,11 +154,6 @@ function drawAnnotationMarker(
   const pos = new Vector2(pos3d.x, pos3d.y);
   const { scale, clipOpacity } = params.depthToScale(pos3d.z);
 
-  // TODO: getIdAtPixel is very expensive and can cause performance issues when
-  // annotations are being updated without the view moving (like when a user is
-  // editing a value or is adding annotations). Turn annotation rendering into a
-  // class and cache the getIdAtPixel result for the current frame, invalidating
-  // it whenever the transform matrix or current frame changes.
```
interim17, line 161: What a nice thing to delete.

src/colorizer/viewport/ColorizeCanvas2D.ts
```diff
@@ -801,7 +799,15 @@ export default class ColorizeCanvas2D implements IInnerRenderCanvas {
...
     this.syncTrackPathLine();
 
+    // Render main scene
     this.renderer.render(this.scene, this.camera);
+
+    // Render to pick buffer
+    const previousRenderTarget = this.renderer.getRenderTarget();
+    this.renderer.setRenderTarget(this.pickRenderTarget);
+    this.renderer.render(this.pickScene, this.camera);
+    this.renderer.setRenderTarget(previousRenderTarget);
```
meganrm: why do you call `this.renderer.render` and `this.renderer.setRenderTarget` each twice?

ShrimpCryptid: I'm rendering a different scene (`this.pickScene`) to a different render target (`this.pickRenderTarget`) and then restoring the previous render target. I can add an additional comment here to make it more clear.

interim17 (approved):

meganrm (approved):

Issue #758, opened 2025-09-04
Point rendering


ShrimpCryptid:
## Use Case

Requested by @alexkhang18! There are a few different possible use cases here with shared functionality.

1. Alex would like to be able to view simulated motion data in TFE, where he has no segmentations but does have tracks/IDs/centroids/feature data. 
2. In Napari, users can turn off segmentations and just visualize the centroids of objects. Alex says this is useful when  segmentations are very densely packed.

## Acceptance Criteria

- [ ] Extend dataset manifest spec to include an override for the frame resolution and scaling (for 3D volumetric data).
- [ ] Add a toggle for showing points instead of segmentations?
  - [ ] If no frame data is provided, default to showing points?
  - [ ] Add controls for radius in the settings tab?
- [ ] Make points clickable.

## Details

- Much of TFE currently relies on loaded frames/volumes to determine the resolution of the viewport. We would need to store the frame resolution in the dataset.
- In 2D, we could fake this during dataset processing by generating points.
- Users would need to be able to click on the points to select them... in 2D we may need to create the pick/render texture on the fly.
- Would need implementations in both 3D and 2D
- Possible similarities to https://intracktive.sf.czbiohub.org/

ShrimpCryptid:
Updated request from Alex in #795:
> It would be great to visualize the nuclei centroids as spheres of controllable diameter and tracks as tails of controllable length behind the spheres. The track tail could be controlled by specifying the number of time frames that show up behind the sphere. It would also be great to toggle and show only the sphere or only the tails. All of this effort is building toward a simplified but hopefully more digestible visualization of the data.

PR #904, opened 2026-04-07
feat: Show centroids in 2D view (points pt. 3)

ShrimpCryptid:
Problem
=======
Part 3 of 6ish for #758, "Point rendering!"

This is a BIG PR, but also does most of the legwork for this feature! The centroid of each cell is now drawn onscreen, and can be resized in the settings. Centroids are colored to match their corresponding cell, and can be clicked and interacted with just like the existing cells.

Future PRs in this series will:
- Allow the segmentations to be hidden and only the points to be shown
- Add coloring modes for centroids (solid color)
- Add centroids in 3D view

I apologize for the large PR and will be reaching out to the requested reviewers to schedule a code walkthrough. Thank you very much!

*Estimated review size: large, 60 minutes*

Solution
========
- Added a new `PointRenderer`, which renders centroids of objects into a texture.
  - The texture contains the object's label ID, encoded as an RGB value.
  - The texture is passed into `ColorizeCanvas2D`, where it is colored and rendered onscreen.
  - Rendered points are responsive to zoom and panning.
- Added new state fields for the centroid radius and visibility.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-904/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fbaseline_colonies_dataset%2Fcollection.json&dataset=Small&feature=volume&bg-key=max_intensity_zprojection_of_lamin_b1&t=15&centroids=1&centroid-radius=3.500&tab=settings
2. Adjust centroid radius, toggle on and off, and play through time.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/efecc174-e604-4f3f-88f8-3a2771684dec

image

-----------------------
1. `PointRenderer2D.ts`
3. Shaders:
  a. `point.vert`
  b. `point.frag`
  c. `colorize_RGBA8U.frag`

Thanks for contributing!


src/colorizer/Dataset.ts
```diff
@@ -539,15 +539,17 @@ export default class Dataset {
...
+      const data = new Float32Array(rawData);
       if (this.frameDimensions) {
         // If provided, normalize centroid coordinates to physical units.
         const physicalDim = physicalDims[i];
         const pixelDim = pixelDims[i];
-        rawData = rawData.map((value) => (value / pixelDim) * physicalDim);
+        for (let j = 0; j < data.length; j++) {
+          data[j] = (data[j] / pixelDim) * physicalDim;
+        }
       }
```
ShrimpCryptid, line 551: Fix for an unrelated bug where the centroid scaling was performed on Uint32 values, which caused them to become 0. I refactored this so they're converted to Float32 values first.

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
```diff
@@ -2,6 +2,7 @@ precision highp usampler2D;
 precision highp int;
```
ShrimpCryptid, line 2: Okay, big refactor happened in this file. Before, the `frame` image was passed in, which stores the cell IDs using RGB values. Now, there's also a `framePoints` texture that's also passed in that has the same encoded cell IDs, and it needs to use some of the same coloring rules as the original base cells. Centroid points are rendered on top of the existing cell segmentations.

High-level changes: 
- The original method `getObjectColor` has had some of its contents split out into a new method, `getFeatureColor`. The rest of the contents of `getObjectColor` are unchanged.
- A separate method, `getPointColor`, handles coloring the centroid points. It has some special rules so it can draw smooth edges.

src/colorizer/viewport/shaders/cellId_RGBA8U.frag
```diff
@@ -1,6 +1,7 @@
 precision highp usampler2D;
```
ShrimpCryptid, line 1: This shader writes cell IDs to a texture that's used for click interactions. The main change here is that the centroid points are also written out, and take higher priority over the existing cells.

src/colorizer/viewport/points/point.vert
```diff
@@ -0,0 +1,31 @@
+/**
+ * Vertex shader for rendering points; applies per-instance position and scaling
+ * and passes parameters (instanceId and radius) to the fragment shader.
+ */
+
+#include <common>
+#include <color_pars_vertex>
+#include <fog_pars_vertex>
+#include <logdepthbuf_pars_vertex>
+#include <clipping_planes_pars_vertex>
```
ShrimpCryptid: Dan: Try removing these

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
```diff
@@ -294,12 +282,100 @@ vec4 getObjectColor(vec2 sUv, float opacity) {
...
+    // - `t=1` when `labelAlpha>=0.33` (e.g. show the edge color slightly
+    //   earlier than the middle of the transition zone so it's more visible)
+    float t = clamp((1.0 - labelAlpha) * 3.0, 0.0, 1.0);
+    bool isEdgePixel = isEdge(framePoints, uv, labelId, EDGE_WIDTH_PX, false);
+    if (isEdgePixel) {
+      // Also apply edge color when this pixel is next to the point of a different
+      // segmentation ID.
+      t = max(t, 0.5);
+    }
+    vec4 outlineColor = alphaBlend(vec4(edgeColor, edgeColorAlpha), baseColor);
```
ShrimpCryptid, line 373: Dan: Use mix() instead

src/colorizer/viewport/points/point.vert
toloudis: Can remove these includes?  The only concern here is future proofing, and simplicity.  Those includes  What does threejs advise?  

See https://github.com/simularium/simularium-viewer/blob/main/src/visGeometry/rendering/PDBGBufferShaders.ts which doesn't use threejs boilerplate in its shader. 
It looks like all you need here is projectionMatrix and modelViewMatrix.

src/colorizer/viewport/points/PointMaterial.ts
```diff
@@ -0,0 +1,55 @@
...
+import fragmentShader from "./point.frag";
+import vertexShader from "./point.vert";
+
+export const enum PointMaterialInstanceAttributes {
+  POSITION = "instancePosition",
+  LABEL_ID = "instanceId",
+}
+
+/**
+ * Draws circular points, using the RGB color to encode the instance ID.
```
toloudis: Consider renaming this to PointMaterial2D, and add in the comment: This is intended to be the identical encoding of instance ID that we get from the image data, so that it can be blended and compatible with reading ids from the image.

src/colorizer/viewport/points/PointRenderer2D.ts
```diff
@@ -0,0 +1,292 @@
...
+
+function getPointBufferGeometry(): BufferGeometry {
+  const geometry = new BufferGeometry();
+  const vertices = new Float32Array([0, 0, 0]);
+  geometry.setAttribute("position", new BufferAttribute(vertices, 3));
+  return geometry;
+}
+
+/**
+ * Renders centroid data to an image texture that can be passed into the
```
toloudis: should we say that the contract here is that the centroid data has to be RGBA encoded instance ids? (Is that what this is supposed to do here?)

src/colorizer/viewport/points/PointRenderer2D.ts
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * Transforms from [0,1] space of the canvas to the [0,1] space of the frame,
+   * accounting for zoom.
+   *
+   * e.g. If frame has the same aspect ratio as the canvas and zoom is set to
+   * 2x, then, assuming that the [0, 0] position of the frame and the canvas are
+   * in the same position, the position [1, 1] on the canvas should map to [0.5,
+   * 0.5] on the frame.
+   */
+  private frameToCanvasCoordinates: Vector2;
```
toloudis: if this is only a pure scaling, then I would not call it frameToCanvasCoordinates but instead something like frameToCanvasScale

src/colorizer/viewport/points/PointRenderer2D.ts
```diff
@@ -0,0 +1,292 @@
...
+  return geometry;
+}
+
+/**
+ * Renders centroid data to an image texture that can be passed into the
+ * colorize shader for 2D rendering + colorization.
+ */
+class PointRenderer2D {
+  private params: PointRendererParams | null = null;
+  private timeToIds: Map<number, Uint32Array>;
```
toloudis: Nit: can't help but think this could be globally useful and could be cached in some kind of higher level state.  It's not necessarily specific to point rendering. I can imagine other modules doing this same thing and keeping redundant info.

src/colorizer/viewport/points/PointRenderer2D.ts
```diff
@@ -0,0 +1,292 @@
...
+      minFilter: NearestFilter,
+      magFilter: NearestFilter,
+    });
+
+    this.emptyTexture = makeEmptyRgbaUint8Texture();
+
+    this.renderFrame = this.renderFrame.bind(this);
+  }
+
+  private computeTimeToIds(dataset: Dataset): Map<number, Uint32Array> {
```
toloudis: this could be a method on dataset, and stored in Dataset.  Then afterward you can just call dataset.getIdsAtTime or whatever.

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
toloudis: Future concern: At some point you need to count up all the fullscreen frame sized or image sized buffers we are allocating because they all have a cost in gpu memory.  The more complicated the engine gets the more you need to track resources, and you end up adding some kind of abstraction layer over resource creation and disposal

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
```diff
@@ -294,12 +282,100 @@ vec4 getObjectColor(vec2 sUv, float opacity) {
...
   if (baseAlpha != 0.0 && isEdgePixel) {
     vec4 transparentEdgeColor = vec4(edgeColor, edgeColorAlpha);
-    color = alphaBlend(transparentEdgeColor, color);
+    baseColor = alphaBlend(transparentEdgeColor, baseColor);
   }
+  return baseColor;
+}
+
+vec4 getPointColor(vec2 uv) {
+    // Get the segmentation id at this pixel
```
toloudis:
```suggestion
  // Get the segmentation id at this pixel
```

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
toloudis, line 373: how is alphaBlend different from mix?

src/colorizer/viewport/ColorizeCanvas2D.ts
```diff
@@ -437,6 +450,12 @@ export default class ColorizeCanvas2D implements IInnerRenderCanvas {
     this.setUniform("canvasToFrameScale", canvasToFrameCoordinates);
 
     this.trackPaths.forEach((trackPath) => trackPath.setPositionAndScale(this.panOffset, frameToCanvasCoordinates));
+    this.pointRenderer.setPositionAndScale(
+      this.panOffset,
+      this.savedScaleInfo.frameToCanvasCoordinates,
+      canvasResolution,
+      this.zoomMultiplier
+    );
```
toloudis, line 460: for consistency, consider putting this code in the same order relative to the this.vectorfield calls similar to that around line 356-362 above.

src/colorizer/viewport/ColorizeCanvas2D.ts
```diff
@@ -801,6 +822,12 @@ export default class ColorizeCanvas2D implements IInnerRenderCanvas {
     this.checkPixelRatio();
     this.syncTrackPathLine();
 
+    // Render points overlay
+    const frameTex = this.pointRenderer.renderFrame(this.renderer, this.currentFrame);
```
toloudis: Nit: bear in mind that renderFrame is asynchronous with respect to the gpu actually drawing into the actual texture.  So a more literal thing here would be `submitRequestFrameTexture` or something like that.

src/colorizer/Dataset.ts
toloudis: ooh this sounds like a bad bug

toloudis (approved):
I left a bunch of comments and suggestions, some probably worth doing but no big fundamental changes suggested.
..also I am noticing you have merge conflicts...

src/colorizer/viewport/points/point.vert
ShrimpCryptid: Removed!

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
ShrimpCryptid: `alphaBlend` is performing an A over B operation, where color A is being blended over color B: https://en.wikipedia.org/wiki/Alpha_compositing 

image


`mix(a, b, t)` performs a linear interpolation between `a` and `b`, e.g.
```
a * t + b * (1 - t)
```

Directly using `mix` does not give the correct values, but it's possible to rewrite `alphaBlend` to use `mix` in the calculations:

```
  float outAlpha = mix(b.a, 1.0, a.a);
  vec3 outColor = mix(b.rgb * b.a, a.rgb, a.a) / outAlpha;
  return vec4(outColor, outAlpha);
```

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
frasercl: Could a tool like Spector help shortcut this?

src/colorizer/viewport/points/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { Color } from "three";
+
+export function getColorForId(id: number): Color {
```
frasercl: optional: could use a line or two of docs for context here

src/colorizer/viewport/points/PointRenderer2D.ts
```diff
@@ -0,0 +1,269 @@
...
+
+const DEFAULT_INSTANCE_COUNT = 128;
+const SCALE_WITH_ZOOM = 0.25;
+
+function getPointBufferGeometry(): BufferGeometry {
+  const geometry = new BufferGeometry();
+  const vertices = new Float32Array([0, 0, 0]);
+  geometry.setAttribute("position", new BufferAttribute(vertices, 3));
+  return geometry;
+}
```
frasercl: Everywhere this function is currently used, its result is immediately cast to `InstancedBufferGeometry`, then copied into another `InstancedBufferGeometry` and discarded. Could this function just return an `InstancedBufferGeometry`?

frasercl (approved):

src/colorizer/viewport/points/utils.ts
ShrimpCryptid: Turns out I am no longer using this function, so I removed it altogether!

src/colorizer/viewport/points/PointRenderer2D.ts
ShrimpCryptid: Done, thank you for the suggestion!

src/colorizer/viewport/shaders/colorize_RGBA8U.frag
ShrimpCryptid: Spector captures commands run in connection with an onscreen canvas, but textures that are rendered offscreen, such as `framePoints`, don't show up when I tested it.

PR #906, opened 2026-04-11
feat: Download Dataset as CSV

### LINKED ISSUE #819: Download dataset CSV
## Use Case

Requested by Susanne! It would be useful for students/scientists looking at TFE data to be able to download the current tracks + features as a CSV for additional data analysis or plotting.

Susanne also mentioned that, in NucMorph, TFE's dataset is a subset of the larger dataset (which includes things like spherical harmonics, for instance). Susanne says that it would be less useful to link to the full dataset, because it introduces additional features and complexity that users may not be expecting.

## Acceptance Criteria

- Add a download button in the header.
- When clicked, the button should start a download of the whole dataset as a CSV file. Columns should include time, track number, label, and all features.
  - Columns should use the human-readable display name, rather than the key.
  - Should centroid/bbox data be included?

## Details

Should the CSV be pre-generated by `colorizer-data`, or should the TFE frontend always handle it?

ShrimpCryptid:
Problem
=======
Closes #819, "Download dataset CSV."

Adds an onscreen button for downloading the current Dataset as a CSV, including metadata (time, seg ID, track, centroids, etc.), feature data, and filter status. The writing of the CSV string is handled in a worker thread to prevent blocking the main thread, since some datasets can have 10,000+ rows!

The key steps for the download process are as follows:
1. The user clicks on the `DownloadDatasetButton`.
2. This calls into `dataset.toCsvDataColumns`, which passes out an array of objects containing the name of a column and its array data.
3. The filtering status (whether an object is included or excluded from a filter) is added to the array of columns.
4. The columns are passed to a worker thread via `workerPool.getCsvString()` to be processed into a CSV string (using the `columnsToCsv` helper method).
5. The final CSV string is returned to the main thread and downloaded as a new file.

*Estimated review size: medium, 25 minutes*

Solution
========
- Added an onscreen button (`DownloadDatasetButton.tsx`) to download the currently selected dataset.
  - On click, an asynchronous worker starts processing feature data into a CSV, which is then downloaded as a local file.
  - Also added a loading spinner and error handling.
- Added unit tests.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-906/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json
6. Once the dataset loads, click the Download icon button next to the dataset dropdown. The CSV should start processing and download when complete.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/31e91b0a-3650-4c53-bf90-5093e67de768

image

image

-----------------------
1. `src/components/Buttons/DownloadDatasetButton.tsx`
2. `src/colorizer/utils/csv_utils.ts`


src/components/LoadingSpinner.tsx
```diff
@@ -110,11 +110,7 @@ export default function LoadingSpinner(inputProps: PropsWithChildren<LoadingSpin
           <Spin
             indicator={
               <LoadSpinnerIconContainer $fontSize={props.iconSize}>
-                {/* Make larger loading icon by multiple of them together */}
-                {/* TODO: Make custom loading spinner SVG? */}
-                <LoadingOutlined />
-                <LoadingOutlined rotate={90} />
-                <LoadingOutlined rotate={135} />
+                <Loading3QuartersOutlined spin={true} />
```
rugeli, line 113: The spinner looks good, but when I clicked, a "No" symbol(circle with a slash) briefly flashed under the pointer cursor on my end, which felt a little unnatural. Consider removing it if you also see it?

ShrimpCryptid: Oh! Yeah, the download button is currently disabled once it begins processing data, which is why you're seeing the 🚫 mouse cursor. I could change the cursor type manually, but I do want to keep the disabled status/styling so users don't try to click on it multiple times.

rugeli (approved):

src/colorizer/utils/csv_utils.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+/**
+ * Takes one or more columns of data and formats them as a CSV string.
+ * @param columns The columns of data to format.
+ * @param delimiter The delimiter to use in the CSV string. Defaults to ",".
+ * @throws Will throw an error if the columns have different numbers of rows or if there are no columns.
+ * @returns The formatted CSV string.
+ */
+export function columnsToCsv(columns: CsvDataColumn[], delimiter: string = ","): string {
```
meganrm, line 16: there are libraries that write to csv, was there a reason they didn't work for this application or a reason you didn't want to increase dependencies?

src/components/Buttons/DownloadDatasetButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+          backgroundColor: theme.color.alert.fill.error,
+          border: `1px solid ${theme.color.alert.border.error}`,
+        },
+      });
+    },
+    [props.notificationApi, theme]
+  );
+
+  const downloadDatasetCsv = useCallback(() => {
+    if (!dataset || !collection || !datasetKey) {
```
meganrm: nit but you have this check below too for disabling a button, could be a constant like `dataAvailable`

src/components/Buttons/DownloadDatasetButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+type DownloadDatasetButtonProps = {
+  notificationApi: NotificationInstance;
+};
+
+export default function DownloadDatasetButton(props: DownloadDatasetButtonProps): ReactElement {
+  const theme = useContext(AppThemeContext);
+
+  const dataset = useViewerStateStore((state) => state.dataset);
+  const datasetKey = useViewerStateStore((state) => state.datasetKey);
+  const collection = useViewerStateStore((state) => state.collection);
```
meganrm, line 23: just out of curiosity are these all independent? or do they all get set at once. I saw below that you have your callback dependent on each of these being changed, but just based on names, I would assume that dataset and datasetKey are pretty linked. Why are they separated out in state?

ShrimpCryptid: Ooh, mostly yes, they are independent. The collection is loaded first, and it contains a mapping from dataset keys to datasets.

`dataset` is agnostic to the `datasetKey`, since you can technically have multiple dataset keys point to the same dataset, but both are set in state at the same time.

src/colorizer/utils/csv_utils.ts
ShrimpCryptid: I am using the `papaparse` library for CSV writing! It expects data in a row-based format, so this method is really just doing all the preprocessing to go from columns (how data is stored in TFE) to rows, and then throwing all of the rows into `papaparse.unparse()` to get the CSV string.

meganrm (approved):

PR #912, opened 2026-05-01
feat: 3D centroids (points pt. 6)

### LINKED ISSUE #758: Point rendering

ShrimpCryptid:
Problem
=======
Part 6 of 6, closing #758, "Point rendering"!

This change adds support for 3D centroid rendering in 3D, and passes the relevant centroid data into `vole-core`.

Solution
========
- Added 3D centroid rendering using the `Spheres3d` drawable in `vole-core`.
  - Centroids are updated on changes to coloring rules and size parameters.
  - Centroid data is not recalculated/sent to `vole-core` unless centroids are visible, for optimization.
- Enabled controls for centroids UI in 3D mode.

Misc:
- Fixed a bug where the 3D segmentations would always be visible on load, regardless of the URL state.

## Type of change

Steps to Verify:
----------------
(PR preview will not work until vole-core is merged, sorry!)

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/d03135e6-d908-45e4-9674-d69ac92037b1

| | |
| --- | --- |
| image | Spheres interact with volume |
| image | Spheres can be shown without the volume |
| image | Spheres can occlude the volume and be picked/highlighted | 

-----------------------
1. `ColorizeCanvas3d.ts`


src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -747,6 +827,7 @@ export class ColorizeCanvas3D implements IInnerRenderCanvas {
 
   render(options?: RenderOptions): void {
     this.syncTrackPathLine();
+    this.syncCentroids();
```
ShrimpCryptid, line 845: TODO: Centroids should not be reprocessed every render

src/colorizer/utils/data_utils.ts
```diff
@@ -486,7 +487,7 @@ export function cloneLabelData(label: LabelData): LabelData {
  * @returns The color for the given object ID, using the rules in the main
  * viewport shader.
  */
-export function computeColorFromId(id: number, params: TrackPathParams): Color {
+export function computeColorFromId(id: number, params: ColorizeStateParams): Color {
```
ShrimpCryptid, line 484: Changed the `params` type here so it wasn't reliant on the `track` parameter, which isn't actually used for coloring.

src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -7,6 +7,7 @@ import {
   type RawArrayLoader,
   RENDERMODE_RAYMARCH,
   SKY_LIGHT,
+  Spheres3d,
```
ShrimpCryptid, line 10: Will be resolved once vole-core updates

src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -615,7 +703,7 @@ export class ColorizeCanvas3D implements IInnerRenderCanvas {
 
       this.view3d.onVolumeData(currentVol, [channelIndex]);
       if (channelIndex === segChannel) {
-        this.view3d.setVolumeChannelEnabled(currentVol, channelIndex, true);
+        this.view3d.setVolumeChannelEnabled(currentVol, channelIndex, this.params?.showSegmentations ?? true);
```
ShrimpCryptid, line 705: Fixes a bug where the segmentations would always be loaded initially, even if `showSegmentations` was set to false.

src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -760,12 +859,13 @@ export class ColorizeCanvas3D implements IInnerRenderCanvas {
...
     this.view3d.removeDrawableObject(this.vectorObject);
     this.vectorObject.cleanup();
+    this.centroidsObject.cleanup();
     this.view3d.removeAllVolumes();
   }
 
   getIdAtPixel(x: number, y: number): PixelIdInfo | null {
     const dataset = this.params?.dataset;
-    if (this.volume?.isLoaded() && dataset) {
+    if (dataset) {
```
ShrimpCryptid, line 866: Allow picking when volume/segmentations are hidden

src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -615,7 +702,7 @@ export class ColorizeCanvas3D implements IInnerRenderCanvas {
 
       this.view3d.onVolumeData(currentVol, [channelIndex]);
       if (channelIndex === segChannel) {
-        this.view3d.setVolumeChannelEnabled(currentVol, channelIndex, true);
+        this.view3d.setVolumeChannelEnabled(currentVol, channelIndex, this.params?.showSegmentations ?? true);
         this.configureColorizeFeature(currentVol, channelIndex);
```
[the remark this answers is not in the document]
ShrimpCryptid: Technically fine, since L705 is called inside of the `onChannelLoadedCallback`, so it'll be called after the below option configuration.

src/colorizer/utils/data_utils.ts
```diff
@@ -515,11 +516,15 @@ export function computeColorFromId(id: number, params: TrackPathParams): Color {
...
- * will match what is used in the main viewport shader.
+ * Returns a Float32Array of RGB (vertex) colors for the given object IDs.
+ * Colors are determined by the feature data, color ramp, and other parameters,
+ * and will match what is used in the main viewport shader.
  * @param ids An array of object IDs to compute colors for.
  * @param params Rendering parameters.
+ * @param useSrgb Whether to convert the colors from linear to sRGB color space,
+ *   `false` by default. Set to `true` when colors will be written or
+ *   manipulated directly in the frag shader, rather than being set as vertex
+ *   colors.
```
toloudis, line 527: Just wondering, why was the Srgb flag introduced?

src/colorizer/viewport/tracks/TrackPath3D.ts
```diff
@@ -45,7 +44,7 @@ export default class TrackPath3D {
     this.lineOverlayObject.setRenderAsOverlay(true);
   }
 
-  public getSceneObjects(): IDrawableObject[] {
+  public getSceneObjects(): Line3d[] {
```
toloudis, line 47: why did this have to get turned into the more specific type?

src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -445,6 +452,86 @@ export class ColorizeCanvas3D implements IInnerRenderCanvas {
     this.updateVectorThickness();
   }
 
+  private getGlobalIdsForCurrentFrame(): Uint32Array | null {
+    if (!this.params || !this.params.dataset) {
+      return null;
+    }
+    const globalIds = this.params.dataset.frameToGlobalIdLookup?.get(this.currentFrame)?.globalIds;
```
toloudis: this looks like it should be a member function of Dataset

src/colorizer/viewport/tracks/TrackPath3D.ts
ShrimpCryptid: The import for `IDrawableObject` got really ugly, and I thought it would be more explicit here to return the underlying object anyways!

toloudis: What does ugly mean?

src/colorizer/utils/data_utils.ts
ShrimpCryptid: We were previously using this function to set colors on the arrows, which sets them as vertex colors which are in linear color space. Spheres3d writes the pixel color out in the fragment shader, which expects sRGB.

src/colorizer/viewport/tracks/TrackPath3D.ts
ShrimpCryptid: I don't think `IDrawableObject` is exported from the top level? So it was reaching down into vole-core's types, something like:

```
import { IDrawableObject } from "@aics/vole-core/es/types/drawables/IDrawableObject";
```

src/colorizer/utils/data_utils.ts
toloudis: the concern here is that it can be confusing as to what color space colors are to be specified in, and you have to be careful about knowing what space are the inputs, outputs, and computations being done.  Everywhere that color is being handled.

src/colorizer/viewport/ColorizeCanvas3D.ts
```diff
@@ -84,6 +92,9 @@ export class ColorizeCanvas3D implements IInnerRenderCanvas {
 
   private timeToVectorData: Map<number, FrameVectorData>;
   private vectorObject: VectorArrows3d;
+  private centroidsObject: Spheres3d;
+  /** Last frame the centroid were updated on. */
```
toloudis:
```suggestion
  /** Last frame the centroids were updated on. */
```

toloudis (approved):
🚀

src/colorizer/viewport/tracks/TrackPath3D.ts
frasercl: Should this type be exported? I guess we expect it to become obsolete in the public interface after eventual refactors.

frasercl (approved):

Issue #861, opened 2026-02-04
Option to visualize scatter plot as heatmap or contours


alexkhang18:
## Use Case
Talk to Ehssan for more detail

## Acceptance Criteria

- [ ] Add a checkbox/toggle for enabling heatmap view.
- [ ] When heatmap view is enabled, change all points to a solid color (grey?)

Stretch:
- [ ] Add color ramp controls for the heatmap view.

ShrimpCryptid:
See https://plotly.com/javascript/2d-density-plots/

ShrimpCryptid:
Feedback from Ehssan:
- [x] Control for the number of contours
- [x] Points should be grey + smaller when contour mode is enabled
- [x] Legend is important because there are no numbers on the contours themselves
- [ ] Color ramp switching is nice but optional. Magma/Plasma might be good defaults?

PR #914, opened 2026-05-02
refactor: Moved helper functions out of `ScatterPlotTab` (scatterplot pt. 1)

ShrimpCryptid:
Problem
=======
Pt. 1 of 2 for #861, "option to visualize scatterplot as heatmap or contours."

Unfortunately this is one of those big tedious PRs that has a really huge line count, but largely consists of shuffling functions between files. I've had it on my todo list for a while to go back and refactor this file, so I took some of the work in #861 as an opportunity to do so!

*Estimated review size: medium, 20-30 minutes (less if mostly skipping/comparing moved lines)*

Solution
========
- Moves the majority of helper functions out of `ScatterPlotTab.tsx` and into `scatter_plot_data_utils.ts`, updating some parameters that were previously set from environment state.
- Added a custom hook for getting debounced state updates.

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-914/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Small&t=15&tab=scatter_plot&interpolate=1&scatter-x=_time_&scatter-bins=20&scatter-range=all
2. Interact with the scatterplot. It should behave as expected.


src/colorizer/viewport/types.ts
```diff
@@ -106,6 +106,20 @@ export type RenderCanvasStateParams = {
...
+  "featureKey",
+  "colorRamp",
+  "colorRampRange",
+  "categoricalPaletteRamp",
+  "outOfRangeDrawSettings",
+  "outlierDrawSettings",
+  "inRangeLUT",
+] satisfies (keyof RenderCanvasStateParams)[];
+
+export type ColorizeStateParams = Pick<RenderCanvasStateParams, (typeof COLORIZE_STATE_KEYS)[number]>;
```
ShrimpCryptid, line 125: I wanted to package these together so that functions that require access to them can just be passed an object. This slightly complicated some debounce-related logic in `ScatterPlotTab`, where elements of state are debounced to prevent scatterplot re-renders during rapid state updates. I made the `useViewerStateStoreDebounced` hook to contain some of that complexity.

src/utils/scatter_plot_data_utils.ts
```diff
@@ -1,19 +1,30 @@
 import { unparse } from "papaparse";
 import type { PlotData, PlotMarker, PlotMouseEvent, Shape } from "plotly.js-dist-min";
-import type { Color } from "three";
+import { Color, type ColorRepresentation } from "three";
```
ShrimpCryptid: Methods are moved, in order wherever possible, from `ScatterPlotTab.tsx`. The only changes I made were updating the function parameters. I'd recommend comparing `ScatterPlotTab` and this file side-by-side.

src/components/Tabs/ScatterPlot/ScatterPlotTab.tsx
```diff
@@ -1,14 +1,14 @@
 import { ExportOutlined } from "@ant-design/icons";
 import { Button, Tooltip } from "antd";
-import Plotly, { type PlotData, type PlotMarker } from "plotly.js-dist-min";
+import Plotly, { type PlotData } from "plotly.js-dist-min";
```
ShrimpCryptid, line 3: I'd recommend comparing this file with `useViewerStateStoreDebounced.ts`, which is where much of its helper methods went.

src/components/Tabs/ScatterPlot/ScatterPlotTab.tsx
```diff
@@ -112,20 +102,15 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
-  const rawColorRampRange = useViewerStateStore((state) => state.colorRampRange);
-  const dataset = useDebounce(rawDataset, 500);
-  const categoricalPalette = useDebounce(rawCategoricalPalette, 100);
-  const [colorRampMin, colorRampMax] = useDebounce(rawColorRampRange, 100);
+  // Debounce changes to the dataset and other frequently-changing values to
+  // prevent noticeably blocking the UI thread with a re-render
+  const [colorizeConfig, isDebouncePending] = useViewerStateStoreDebounced(colorizeStateSelector, 100, {
+    dataset: 500,
+  });
+  const { dataset } = colorizeConfig;
```
ShrimpCryptid, line 109: This replaces all the above state selectors and the removed state handling. I think it's a lot cleaner, though the hook it's associated with is pretty big.

src/components/Tabs/ScatterPlot/ScatterPlotTab.tsx
```diff
@@ -330,407 +315,20 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
-    }
-    const categories = dataset.getFeatureCategories(featureKey) || [];
-    return (
-      categories.reduce((_prev: any, val: string, acc: number) => {
-        return Math.max(val.length, acc);
-      }, 0) * 8
-    );
-  };
-
-  // TODO: Move to `scatter_plot_data_utils.ts`
```
ShrimpCryptid, line 532: :eyes:

src/hooks/useViewerStateStoreDebounced.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Custom hook that returns a debounced, selected subset of viewer state.
+ * @param selector A selector function that selects a subset of the viewer state
+ * to debounce. Automatically wrapped with `useShallow` for shallow comparison
+ * of selected state.
+ * @param delayMs The default debounce delay, in milliseconds, to apply to all
+ * selected properties. Set to 0 to disable debounce by default.
+ * @param propertyDelayMs Property-specific delay overrides. Use this to
+ * increase or disable the debounce for certain values.
```
meganrm: I'm curious how you envision this being used, since it seems like you are only calling this hook in one place. does each call really need to set a default delay that would be different from other calls? I guess I'm wondering why these (delayMs and propertyDelayMs) are dynamic and not just constants

ShrimpCryptid: Being perfectly honest, I'm not sure if I'll use it elsewhere? But I did want to generalize it in case I do.

I like having the ability to tune particular values and adjust the timing based on how often the value is changing or how expensive the re-render is. Changes to `dataset` incurs renders across the entire app, so the debounce in `ScatterPlotTab` is primarily so the plotly render isn't happening at the same time as the rest of the re-rendering. Some properties that are associated with sliders tend to change quickly and need a higher debounce timing than others.

There probably could be constants across the app for different debounce timings (MINOR = 100ms, MAJOR = 500ms) or something?

src/hooks/useViewerStateStoreDebounced.ts
```diff
@@ -0,0 +1,102 @@
...
+ * @returns An array tuple of:
+ *   1. The debounced state selected from the viewer store.
+ *   2. A boolean, indicating whether there are pending debounced updates.
+ */
+export const useViewerStateStoreDebounced = <T extends Partial<ViewerStore>>(
+  selector: (state: ViewerStore) => T,
+  delayMs: number,
+  propertyDelayMs: Partial<Record<keyof T, number>> = {}
```
frasercl: My first instinct seeing this argument is: can you just select the specific state fields for which you want a longer debounce (looks like just `dataset`, in practice) in a different invocation of the hook?

src/hooks/useViewerStateStoreDebounced.ts
```diff
@@ -0,0 +1,102 @@
...
+  const propertyTimeouts = useRef<Partial<Record<keyof T, TimeoutHandle>>>({});
+  const pendingValues = useRef<Partial<T>>({});
+  const propertyDelayMsRef = useRef(propertyDelayMs);
+  propertyDelayMsRef.current = propertyDelayMs;
+
+  const debouncedStateRef = useRef(rawState);
+
+  const getPropertyDelayMs = (key: keyof T): number => {
+    return Math.max(propertyDelayMsRef.current[key] ?? delayMs, 0);
+  };
```
frasercl: Looks like this function is broken out in part to hide the ref that prevents stale closure problems... but `delayMs` doesn't get its own ref and so could become stale

src/hooks/useViewerStateStoreDebounced.ts
```diff
@@ -0,0 +1,102 @@
...
+          if (propertyTimeouts.current[key]) {
+            clearTimeout(propertyTimeouts.current[key]);
+          }
+          propertyTimeouts.current[key] = setTimeout(() => {
+            const value = pendingValues.current[key];
+            debouncedStateRef.current = {
+              ...debouncedStateRef.current,
+              [key]: value,
+            };
+            triggerStateUpdate();
```
frasercl: Why not just put `pendingValues` in state, rather than saving it to a ref and immediately setting a different dummy value?

src/hooks/useViewerStateStoreDebounced.ts
```diff
@@ -0,0 +1,102 @@
...
+      if (propertyDelayMs === 0) {
+        debouncedStateRef.current = {
+          ...debouncedStateRef.current,
+          [key]: rawState[key],
+        };
+      } else {
+        isDebouncePending = true;
+      }
+    }
+  }
```
frasercl: Kind of feels like you can reasonably expect a "debounce" hook to receive a nonzero timeout...

If I'm understanding it right, the rest of the hook shouldn't entirely fall over if this happens, it just might trigger more renders than necessary?

ShrimpCryptid: Yeah, there would just be additional re-renders happening.

src/hooks/useViewerStateStoreDebounced.ts
ShrimpCryptid: Part of what's nice about this hook is that the object it returns is memoized (only returns a new object reference when a property updates) so it can be used in dependency arrays. I could have `dataset` be in a separate hook, but I'd also need to combine and memoize it....

src/hooks/useViewerStateStoreDebounced.ts
ShrimpCryptid: I did it to prevent having to remake the `pendingValues` map every time it's changed, since otherwise updating its values without recreating it would not trigger a state update.

ShrimpCryptid: I tinkered with this a bit, but it ends up being less clean? I need to have a `useState` and a `useRef`, but I only ever use the state setter from `useState` and totally ignore its value in favor of the ref.

src/hooks/useViewerStateStoreDebounced.ts
frasercl: It wouldn't work to break `dataset` out of the selector you're currently passing to `useViewerStateStoreDebounced`, debounce only it using the old `useDebounce` method, and depend on two things instead of one? Or it would work but degrade ergonomics?

Come to think of it, I wonder if it wouldn't be slightly better from a separation-of-concerns perspective to pull `useViewerStateStore` out of `useViewerStateStoreDebounced` and rework it as a beefed-up variant of the current `useDebounce` that works over an object.

src/hooks/useViewerStateStoreDebounced.ts
ShrimpCryptid: Ended up revising this with Cameron into the new `useDebouncedRecord` that can debounce individual properties of an object.

meganrm (approved):

src/hooks/useDebounceRecord.ts
```diff
@@ -0,0 +1,48 @@
...
+          setState((currentState) => ({
+            ...currentState,
+            [key]: value,
+          }));
+          delete timeoutRef.current[key];
+        }, delayMs);
+      }
+    }
+    prevRecordRef.current = record;
+  }, [record, delayMs]);
```
frasercl, line 35: Thinking through what happens on an update to `delayMs`... current timeouts stick around with the previous delay. Which is I think as close to right as we can expect to get without a really significant increase in complexity?

src/components/Tabs/ScatterPlot/ScatterPlotTab.tsx
```diff
@@ -112,20 +102,14 @@ export default memo(function ScatterPlotTab(props: ScatterPlotTabProps): ReactEl
...
-  const rawCategoricalPalette = useViewerStateStore((state) => state.categoricalPalette);
-  const rawColorRampRange = useViewerStateStore((state) => state.colorRampRange);
-  const dataset = useDebounce(rawDataset, 500);
-  const categoricalPalette = useDebounce(rawCategoricalPalette, 100);
-  const [colorRampMin, colorRampMax] = useDebounce(rawColorRampRange, 100);
+  // Debounce changes to the dataset and other frequently-changing values to
+  // prevent noticeably blocking the UI thread with a re-render
+  const rawColorizeState = useViewerStateStore(useShallow(colorizeStateSelector));
+  const [colorizeConfig, isDebouncePending] = useDebounceRecord(rawColorizeState, 100);
+  const { dataset } = colorizeConfig;
```
frasercl, line 109: `dataset` is no longer debounced on its own delay, right?

You're the expert on whether that's a problem that needs fixing at all, but two options if it is:
- Pull `dataset` out of `colorizeStateSelector`, select it on its own, give it its own `useDebounce`... ugh, but then it runs into exactly the thing you were talking about with not having a flag for when it's pending. `rawDataset !== dataset`?
- Hate to say it but... especially if you're okay with the behavior described above with delays, it doesn't look like that much additional code to put back the custom delays per field in the new hook. Wouldn't even need a ref. Would absolutely run afoul of exactly the "is this too complicated" criticism I brought up above though, and make a hypocrite of me besides.

frasercl (approved):

src/hooks/useDebounceRecord.ts
ShrimpCryptid: Yeah, I think that's probably correct behavior? I don't expect `delayMs` to be set dynamically.

src/components/Tabs/ScatterPlot/ScatterPlotTab.tsx
ShrimpCryptid: I think it's fine for `dataset` to not be on its own delay. Lots of components are dependent on changes to `dataset`, so the main reason for debouncing it was to prevent the React re-renders and the Plotly re-renders from occurring simultaneously-- more is helpful but not necessary in this case.

src/hooks/useDebounceRecord.ts
frasercl: I wouldn't expect it either, but it's good to know that nothing will fall over if it someday is.

PR #919, opened 2026-05-12
refactor: Ran `zizmor` on workflows

ShrimpCryptid:
Problem
=======
Added `zizmor` to the repository for static workflow analysis after reading about it after the Tanstack security incident: https://tanstack.com/blog/incident-followup

Zizmor scans for common security issues in GitHub action workflows, and has already flagged several issues in this PR that I've fixed! You can find more information about the tool here: https://docs.zizmor.sh/

Solution
========
- Updated existing workflows:
  - Workflows now have narrowly-scoped permissions per-job.
  - Added hash pinning to all actions.
  - Disabled credential persistence on all usage of `actions/checkout`: https://docs.zizmor.sh/audits/#artipacked
  - Updated `actions/checkout` to `v6.0.2`.
  - Bumped several other actions to their newest versions.
- Added `zizmorcore/zizmor-action` for static workflow analysis.

## Type of change

Steps to Verify:
----------------
1. See actions are passing.

### DIFF
--- a/.github/workflows/aws-deploy.yml
+++ b/.github/workflows/aws-deploy.yml
@@ -76,7 +76,7 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
 
       # Setup variables based on the staging or production environment
-      - name: Set ECS variables based on environment
+      - name: Set AWS (S3/CloudFront) variables based on environment
         run: |
           if [ "${GITHUB_EVENT_INPUTS_ENVIRONMENT}" == "production" ]; then
             echo "S3_BUCKET=${PRODUCTION_S3_BUCKET}" >> $GITHUB_ENV
--- a/.github/workflows/aws-deploy.yml
+++ b/.github/workflows/aws-deploy.yml
@@ -29,7 +29,8 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 #v3.6.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+
         with:
           persist-credentials: false
 
--- a/.github/workflows/aws-deploy.yml
+++ b/.github/workflows/aws-deploy.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 #v3.6.0
         with:
           persist-credentials: false
 
@@ -91,7 +91,7 @@ jobs:
           GITHUB_EVENT_INPUTS_ENVIRONMENT: ${{ github.event.inputs.environment }}
 
       - name: Download build artifacts
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c #v8.0.1
         with:
           name: aws-deploy-files
           path: ./dist
--- a/.github/workflows/aws-deploy.yml
+++ b/.github/workflows/aws-deploy.yml
@@ -13,6 +13,8 @@ on:
         required: true
         type: choice
 
+permissions: {}
+
 env:
   AWS_ACCOUNT_ID: ${{ vars.AWS_PUBLIC_DATA_RELEASES_ACCOUNT_ID }}
   AWS_REGION: ${{ vars.AWS_DEFAULT_REGION }}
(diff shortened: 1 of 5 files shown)


.github/workflows/aws-deploy.yml
```diff
@@ -13,6 +13,8 @@ on:
         required: true
         type: choice
 
+permissions: {}
```
ShrimpCryptid, line 16: This removes default permissions from an action, so permissions have to be set on the specific jobs that need them.

.github/workflows/aws-deploy.yml
```diff
@@ -21,17 +23,16 @@ env:
...
   build:
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+
+        with:
+          persist-credentials: false
```
ShrimpCryptid, line 36: See https://docs.zizmor.sh/audits/#artipacked

.github/workflows/aws-deploy.yml
```diff
@@ -74,27 +78,29 @@ jobs:
       # Setup variables based on the staging or production environment
       - name: Set ECS variables based on environment
         run: |
-          if [ "${{ github.event.inputs.environment }}" == "production" ]; then
-            echo "S3_BUCKET=${{ env.PRODUCTION_S3_BUCKET }}" >> $GITHUB_ENV
-            echo "CLOUDFRONT_DISTRIBUTION_ID=${{ env.PRODUCTION_CLOUDFRONT_DISTRIBUTION_ID }}" >> $GITHUB_ENV
-          elif [ "${{ github.event.inputs.environment }}" == "staging" ]; then
-            echo "S3_BUCKET=${{ env.STAGING_S3_BUCKET }}" >> $GITHUB_ENV
-            echo "CLOUDFRONT_DISTRIBUTION_ID=${{ env.STAGING_CLOUDFRONT_DISTRIBUTION_ID }}" >> $GITHUB_ENV
```
ShrimpCryptid, line 82: The `${{ }}` template expansion syntax can allow for shell injection. I think this was fine for our use case, since none of these environment variables were controlled by users, but I fixed it anyways.

See https://docs.zizmor.sh/audits/#template-injection

.github/workflows/deploy-public-build.yml
```diff
@@ -32,27 +35,31 @@ jobs:
...
 
       - name: Download artifact
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c #v8.0.1
         with:
           name: production-files
           path: ./dist
 
       - name: Deploy
-        uses: JamesIves/github-pages-deploy-action@v4
```
ShrimpCryptid, line 55: I don't think `v4` was actually a versioned release... it looks like an accidental tag that was committed to main.

.github/workflows/release.yml
```diff
@@ -40,13 +46,11 @@ jobs:
...
 
       - name: Zip build artifact
-        run: zip -r tfe-${{ steps.get_version.outputs.VERSION }}.zip viewer/*
+        run: zip -r tfe-${VERSION_TAG}.zip viewer/*
+        env:
+          VERSION_TAG: ${{ steps.get_version.outputs.VERSION }}
 
       - name: Create release
-        # v2.4.1 10/14/25
-        uses: softprops/action-gh-release@6da8fa9354ddfdc4aeace5fc48d7f679b5214090
```
ShrimpCryptid, line 47: Replaced this action with GitHub's native `gh` CLI tool.

.github/workflows/run-static-checks.yml
```diff
@@ -1,12 +1,20 @@
 name: run-static-checks
-run-name: ✅ Checking '${{'#'}}${{github.ref_name}}' on branch '${{github.head_ref}}'
```
ShrimpCryptid, line 2: lol

toloudis (approved):
Changes seem fine, although it's introducing patterns that future code changes may or may not follow.  (Also, it looks like copilot did some work here too... so do we trust zizmor or not?)
Did you do this through the vscode plugin?
Is there any long term plan for enforcement?

.github/workflows/run-static-checks.yml
```diff
@@ -1,12 +1,21 @@
...
+run-name: ✅ Checking '#${{github.ref_name}}' on branch '${{github.head_ref}}'
```
rugeli: 1. with `✅ Checking ...`, a failed run would show a red X next to the green check icon, which looks a little mixed. Maybe use a more neutral icon here?

2. `#${{ github.ref_name }}` shows as `#919/merge` for pr runs. would `github.event.number` work better here so the title shows just `#919`?

rugeli (approved):
Overall looks great! left comments around the `run-name` naming

ShrimpCryptid:
> Changes seem fine, although it's introducing patterns that future code changes may or may not follow. (Also, it looks like copilot did some work here too... so do we trust zizmor or not?) Did you do this through the vscode plugin? Is there any long term plan for enforcement?

I installed `zizmor` locally and was running it as my check, but there's also now a zizmor action that is run on every pull request that will also flag future issues! See `run-static-checks/static-workflow-checks`.

PR #924, opened 2026-05-21
feat: Apply gaussian smoothing kernel to points (3d plot pt. 2)

ShrimpCryptid:
Problem
=======
Part 2 of 4ish for the 3D plotting feature!

This change updates the math used to calculate the vector flow fields, and replicates the Gaussian smoothing described in the upcoming endo paper. Big thank you to @eeangelini for help understanding and validating the approach!

*Estimated review size: medium-large, 30-40 minutes (please turn off whitespace)*

Solution
========
- Added filtered value + outlier exclusion when calculating vector flow fields.
- Refactored vector flow field calculation into separate summing and averaging/smoothing step.
  - Added logic for creating and convolving with Gaussian kernels.
  - Added user control for Gaussian smoothing of vector bins.
- Added unit tests

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-924/viewer?collection=https%3A%2F%2Fvast-files.int.allencell.org%2Fendothelial%2Fmorphological_features%2Ftimelapse_feature_explorer_grid%2Fcollection.json&dataset=Low-to-high+shear+stress+20250224_20X_P0&tab=3d_plot
2. Select PC1, PC2, and PC3 as the feature axes. A Gaussian-smoothed plot will appear.
3. Click the checkbox to toggle Gaussian smoothing on and off.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/093578f5-2ca8-4469-b520-0c4c489c1cdd

| Before | After |
| --- | --- |
| image | image |

-----------------------
1. `math_utils.ts`


src/colorizer/utils/math_utils.ts
```diff
@@ -381,25 +380,27 @@ export function getBinValue(binIndex: number, range: [number, number], steps: nu
  * @param yRange Range of values for the Y dimension, as a tuple [min, max].
  * @param zRange Range of values for the Z dimension, as a tuple [min, max].
  * @param binsPerAxis Number of bins per axis, as a tuple [xBins, yBins, zBins].
- * @returns VectorFieldData containing the calculated vector flow field.
+ * @returns VectorSumData containing the summed vectors.
  */
-export function calculateVectorFlowField(
```
ShrimpCryptid, line 395: Before, this function first summed the vectors, then performed an averaging step by dividing by the count. This change scopes the function to just the first step of summing vectors.

The returned `VectorSumData` is then passed into new helper functions to either 1) divide by the count, or 2) perform a gaussian smoothing step using neighboring bins.

src/colorizer/utils/math_utils.ts
```diff
@@ -1,6 +1,6 @@
 import { Vector2 } from "three";
 
-import type { VectorFieldData } from "src/colorizer";
+import type { VectorFieldData, VectorSumData } from "src/colorizer";
```
ShrimpCryptid: I made a new type! `VectorSumData` is calculated first, and the final `VectorFieldData` can be derived from it in one of two ways based on user settings.

tests/math_utils.test.ts
```diff
@@ -3,8 +3,9 @@ import { describe, expect, it } from "vitest";
```
ShrimpCryptid, line 3: Easier to review with whitespace turned off

src/colorizer/utils/math_utils.ts
```diff
@@ -496,3 +532,149 @@ export function thresholdVectorFlowFieldByCount(
...
+  kernelZ: number[]
+): Float32Array {
+  let output: Float32Array;
+
+  output = convolve1dFilter(arr, arrDims, kernelX, "x");
+  output = convolve1dFilter(output, arrDims, kernelY, "y");
+  output = convolve1dFilter(output, arrDims, kernelZ, "z");
+
+  return output;
+}
```
ShrimpCryptid, line 651: We perform the convolution for X, Y, and Z because of a neat property of Gaussians-- it's a [separable filter](https://en.wikipedia.org/wiki/Separable_filter), so it can be split into multiple 1-dimensional filters!

This speeds up a calculation that's normally `O(N x M^3)` to `O(3 x N x M) = O(N x M)`, where N is the size of the array and M is the filter size. I didn't measure it, but the speedup is 5-6x faster at least.

src/colorizer/utils/math_utils.ts
```diff
@@ -496,3 +532,148 @@ export function thresholdVectorFlowFieldByCount(
...
+          const offset = k - kernelMid;
+          const weight = kernel[k];
+          const sampleX = direction === "x" ? x + offset : x;
+          const sampleY = direction === "y" ? y + offset : y;
+          const sampleZ = direction === "z" ? z + offset : z;
+          // Skip out-of-bounds samples
+          if (sampleX < 0 || sampleX >= xDim || sampleY < 0 || sampleY >= yDim || sampleZ < 0 || sampleZ >= zDim) {
+            continue;
+          }
+          const arrValue = arr[getIndex(sampleX, sampleY, sampleZ)] || 0;
```
rugeli, line 615: not a blocker, and I didn’t trace the upstream functions closely, but just wondering: could arr[index] ever be NaN here? If yes, this would silently turn it into 0

rugeli (approved):
nice work and walkthrough video! I gave this a surface-level pass and left one small non-blocking question

src/colorizer/types.ts
```diff
@@ -373,12 +373,22 @@ export type FeatureRangeData = {
   range: [number, number];
 };
 
+export type VectorSumData = {
```
toloudis, line 391: Readability: I would appreciate comments here on what the Pos variables are supposed to mean, and what the Sum variables are supposed to mean.  What are the array lengths? What's in them?

src/colorizer/types.ts
```diff
@@ -373,12 +373,22 @@ export type FeatureRangeData = {
...
+
 export type VectorFieldData = {
   xPos: Float32Array;
   yPos: Float32Array;
   zPos: Float32Array;
   xData: Float32Array;
   yData: Float32Array;
   zData: Float32Array;
-  count: Uint32Array;
+  count: Float32Array;
```
toloudis, line 422: Same as above, a comment about VectorFieldData and VectorSumData to say what they are filled with, would help.

src/colorizer/workers/SharedWorkerPool.ts
```diff
@@ -95,12 +95,38 @@ export default class SharedWorkerPool {
...
+   * @param inRangeLUT Lookup table for whether an object instance is in range
+   * of filters (`=1`) or not (`=0`), as a flat array. Values outside of the
+   * filter range will be ignored when computing the flow field.
+   * @param gaussianBandwidth The bandwidth (or standard deviation) of the
+   * Gaussian kernel to apply to the binned vectors, as a fraction of the number
+   * of bins. If provided, uses Gaussian smoothing; if not provided, uses simple
+   * averaging.
+   * @returns a `VectorFieldData` object containing the computed flow field data
+   * and metadata.
+   */
```
toloudis, line 113: Nit: it's interesting to me that SharedWorkerPool contains the specific tasks, instead of just being a worker pool that can be handed arbitrary tasks.  I would consider putting the tasks in a module that's more specific to them, if possible..   
After reading more, it looks like this is how the worker is designed.   I still would consider breaking out the individual tasks into their own ts files, for maintenance and readability reasons.  Maybe for a different PR.

src/colorizer/utils/math_utils.ts
```diff
@@ -496,3 +532,148 @@ export function thresholdVectorFlowFieldByCount(
     count: count.filter((_, i) => passedThreshold[i]),
   };
 }
+
+export function make1dGaussianKernel(size: number, bandwidth: number): number[] {
```
toloudis, line 554: Nit: I am more used to "sigma" or "standard deviation" instead of "bandwidth".  A brief comment to say this is the classic sigma of a gaussian might be helpful.   But it's pretty obvious when you see the Math.exp expression.

src/colorizer/workers/worker.ts
```diff
@@ -63,29 +66,55 @@ async function getVectorFlowField(
...
+      [xFeature.bins, yFeature.bins, zFeature.bins],
+      kernelX,
+      kernelY,
+      kernelZ
+    );
+  } else {
+    vectorFlowFieldData = averageVectorFlowField(vectorSumData);
+  }
+
+  vectorFlowFieldData = filterVectorFlowFieldData(vectorFlowFieldData);
```
toloudis, line 110: this special kind of binned averaging had a name ,which I already forgot.  But it would be great to name the thing here in a comment, so people can refer to the theoretical basis for it.

src/colorizer/utils/math_utils.ts
toloudis, line 1: file is getting long; maybe the specialized vector field handling functions could be in its own ts file.

toloudis (approved):
I have a couple recommendations for comments and putting code in single-purpose modules, but overall LGTM!  The math seems to be working, and separable filters are great.

src/colorizer/utils/math_utils.ts
ShrimpCryptid: Oh, hmm, that's a good question! It would be possible for NaN values to be passed in, and in that case I think the NaN => 0 behavior is correct.

src/colorizer/utils/math_utils.ts
ShrimpCryptid: I'm going to open up a new PR that's just for refactoring `math_utils`.

ShrimpCryptid: See https://github.com/allen-cell-animated/timelapse-colorizer/pull/931!

PR #925, opened 2026-05-22
refactor: Added 3D Plot state slice, replaced local state (3d plot pt. 3)

ShrimpCryptid:
Problem
=======
Part 3 of 4 for the 3D plot visualization!

This change takes all of the state variables that were previously stored in `Plot3dTab.tsx` and moves them into the global state store, with URL serialization, deserialization, and validation. This means that any parameters set in the 3D plot tab will persist when a page is refreshed or sent as a link.

*Estimated review size: medium-large, 30-40 minutes (much of it is boilerplate in `plot_3d_slice.ts`)*

Solution
========
- Added new state slice for the 3D plot (`plot_3d_slice.ts`).
  - Added serialization and deserialization.
- Updated Plot 3D components to derive state from global state, instead of declaring and passing state props.
- Updated unit tests.

## Type of change

Steps to Verify:
----------------
1. Open PR preview. The 3D plot view should match the screenshot below: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview-internal/pr-925/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&tab=3d_plot&p3d-x=volume&p3d-y=height&p3d-z=xy_aspect_ratio&p3d-vc-bins=20&p3d-vc-scale=1.500&p3d-vc-ramp=matplotlib-plasma&p3d-vc-thresh=1&p3d-avg-w=1.600&p3d-avg-n=1&p3d-gauss=1&p3d-gauss-bw=10
2. Adjust any settings in the 3D plot tab. Refresh the page; the changes will persist.

image

-----------------------
1. `plot_3d_slice.ts`


src/state/slices/plot_3d_slice.ts
```diff
@@ -0,0 +1,341 @@
...
+  plot3dVectorScale: number;
+  plot3dVectorColorRampKey: string;
+  plot3dVectorColorRampReversed: boolean;
+  /**
+   * Minimum number of deltas required for a bin to be included in the vector
+   * field.
+   */
+  plot3dVectorThreshold: number;
+
+  plot3dLineWidth: number;
```
ShrimpCryptid, line 38: Will have UI added in a follow-up PR, #926

src/state/slices/plot_3d_slice.ts
```diff
@@ -0,0 +1,353 @@
...
+  plot3dLineMovingAverageWindow: number;
+
+  plot3dUseGaussian: boolean;
+  plot3dGaussianBandwidthPct: number;
```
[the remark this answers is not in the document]
ShrimpCryptid: UI for this is added in #926.

src/components/Tabs/Plot3d/Plot3dConeControls.tsx
```diff
@@ -7,17 +7,11 @@ import {
...
-  coneSize: number;
-  setConeSize: (value: number) => void;
-  coneColorRampKey: string;
-  setConeColorRampKey: (value: string) => void;
-  coneColorRampReversed: boolean;
-  setConeColorRampReversed: (value: boolean) => void;
-  threshold: number;
-  setThreshold: (value: number) => void;
+  disabled?: boolean;
 };
```
ShrimpCryptid, line 21: All these components no longer require props, and can pull the relevant getters + setters from global state.

src/colorizer/colors/color_ramps.ts
```diff
@@ -421,20 +421,24 @@ export const DISPLAY_COLOR_RAMP_LINEAR_KEYS = [
   "seaborn-cubehelix_blue",
   "matplotlib-reds",
   "matplotlib-turbo",
-];
-export const DISPLAY_COLOR_RAMP_DIVERGING_KEYS = ["esri-blue_red_8", "esri-green_brown_1", "matplotlib-purple_orange"];
+] as const;
```
ShrimpCryptid, line 424: Made these into constants so I can do type validation against them.

src/colorizer/ColorRamp.ts
```diff
@@ -144,20 +144,14 @@ export default class ColorRamp {
     return canvas;
   }
 
-  public getPlotlyColorScale(reversed: boolean = false): [number, string][] {
```
ShrimpCryptid, line 147: Removed reverse flag, since reversed ColorRamp objects will already have their color stops reversed.

src/colorizer/colors/color_ramps.ts
toloudis: what does "as const" mean?

src/components/Tabs/Plot3d/Plot3dTab.tsx
```diff
@@ -28,47 +21,41 @@ export default function Plot3dTab(): ReactElement {
...
+  const rawBins = useViewerStateStore((state) => state.plot3dVectorBins);
+  const xAxisFeatureKey = useViewerStateStore((state) => state.plot3dXAxis);
+  const yAxisFeatureKey = useViewerStateStore((state) => state.plot3dYAxis);
+  const zAxisFeatureKey = useViewerStateStore((state) => state.plot3dZAxis);
+  const rawConeSize = useViewerStateStore((state) => state.plot3dVectorScale);
+  const applyGaussian = useViewerStateStore((state) => state.plot3dUseGaussian);
+  const gaussianBandwidthPct = useViewerStateStore((state) => state.plot3dGaussianBandwidthPct);
+  const coneColorRamp = useViewerStateStore((state) => state.plot3dColorRamp);
+  const rawThreshold = useViewerStateStore((state) => state.plot3dVectorThreshold);
+  const rawMovingAverageWindow = useViewerStateStore((state) => state.plot3dLineMovingAverageWindow);
```
toloudis, line 50: is it normal for these lines to be repeated in multiple files? (like PlotFeatureControls and here - and I realize it's not exactly the same lines but a lot of overlap)

src/state/slices/plot_3d_slice.ts
```diff
@@ -0,0 +1,334 @@
...
+import type { DatasetSlice } from "./dataset_slice";
+
+const DEFAULT_COLOR_RAMP_KEY = "matplotlib-turbo" satisfies (typeof DISPLAY_COLOR_RAMP_LINEAR_KEYS)[number];
+
+export type Plot3dSliceState = {
+  plot3dXAxis: string | null;
+  plot3dYAxis: string | null;
+  plot3dZAxis: string | null;
+
+  plot3dVectorBins: number;
```
toloudis, line 28: plot3dAxis makes sense to me, but what are "vector bins" in a 3d plot?  is that explained anywhere else?  For me, reading this code, plot3dVectorBins could use a comment.

src/state/slices/plot_3d_slice.ts
```diff
@@ -0,0 +1,334 @@
...
+   * field.
+   */
+  plot3dVectorThreshold: number;
+
+  plot3dLineWidth: number;
+  plot3dLineMovingAverageWindow: number;
+
+  plot3dUseGaussian: boolean;
+  /** The bandwidth or standard deviation for the gaussian smoothing. */
+  plot3dGaussianBandwidthPct: number;
```
toloudis, line 51: Pct?  What is it a percentage of?

src/state/slices/plot_3d_slice.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  // Derived state
+  plot3dColorRamp: ColorRamp;
```
toloudis: what's the difference between plot3dColorRamp and plot3dVectorColorRampKey? are they referring to the same thing?

src/state/slices/plot_3d_slice.ts
```diff
@@ -0,0 +1,334 @@
...
+  /**
+   * Minimum number of deltas required for a bin to be included in the vector
+   * field.
+   */
+  plot3dVectorThreshold: number;
+
+  plot3dLineWidth: number;
+  plot3dLineMovingAverageWindow: number;
+
+  plot3dUseGaussian: boolean;
```
toloudis, line 46: what happens if usegaussian is false?  does it still do binning/averaging?

src/colorizer/colors/color_ramps.ts
ShrimpCryptid: It marks for TypeScript that the list is immutable/readonly, and narrows its type definition!

```ts
// Before: var is a string[]
const var = ["a", "b", "c"];

// After: var is an array of the specific string values "a", "b", "c" in that order
const var = ["a", "b", "c"] as const;
```

In this case, this lets me validate my constants in `plot_3d_slice` and will flag any potential mismatches.

```ts
type LinearRampKey = (typeof DISPLAY_COLOR_RAMP_LINEAR_KEYS)[number];
const color1 = "matplotlib-turbo" satisfies LinearRampKey; // ✅
const color2 = "not a color ramp key" satisfies LinearRampKey; // 🚫 compilation error
```

src/components/Tabs/Plot3d/Plot3dTab.tsx
ShrimpCryptid: Yes! Having these as individual lines means that each component is only pulling in the state values that are relevant for its render, and unrelated state does not trigger unnecessary re-renders. 

Here in `Plot3dTab`,  for instance, we're accessing lots of state values but not their setters, while in `PlotFeatureControls` we need both.

I could reduce the line count here by having a single shared selector, but it comes with several drawbacks:
1. A component will re-render on any change to any value in that selector, even if it doesn't actually need it.
2. If a value is no longer needed in the future, there is no way to automatically flag that the value is unused.

src/state/slices/plot_3d_slice.ts
ShrimpCryptid: Oops, I updated this in pt. 4 and forgot to pull those changes in here. 

The updated comment says:
```ts
  /**
   * The bandwidth or standard deviation for the gaussian smoothing, as a
   * percentage of the number of bins in the (0, 100] range.
   */
  plot3dGaussianBandwidthPct: 15,
```

src/state/slices/plot_3d_slice.ts
ShrimpCryptid: The `plot3dColorRamp` is derived from the current key and the reversed flag. Each time either of those is updated, the color ramp is recalculated and the updated value is stored in state.

src/state/slices/plot_3d_slice.ts
ShrimpCryptid: Yes, it does a simple average (sum of bin / count)

src/state/slices/plot_3d_slice.ts
ShrimpCryptid: This is the number of bins. I'll update this!

src/state/slices/plot_3d_slice.ts
toloudis: Nit: in that case, I might have an enum of GAUSSIAN, MEAN instead of the bool.  would have helped readability for this PR.

toloudis: The new comment helps a lot!

toloudis (approved):
Thanks!

ascibisz (approved):

Issue #632, opened 2025-04-12
Single file datasets


ShrimpCryptid:
## Use Case

For sharing and for certain applications like BFF, it's more convenient to package datasets as a single file instead of a directory with many files.

## Acceptance Criteria

- [ ] Add an option in `colorizer-data` to package the dataset as a single file.
  - [ ] When selected, save data in a single Parquet file with columns for each of the features, time, track, and other data
- [ ] On the TFE frontend, add support for loading a `.parquet` file, with the assumption that it's a single file dataset.
  - [ ] Also ideally, show loading progress?

## Details

- Single-file datasets do not need to include image data (can be a link to an external resource)

PR #930, opened 2026-06-03
refactor: Separate dataset load function into class (single-file datasets pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of 2 for #632, "single file datasets."

Previously, the `Dataset` class in TFE acted as a representation of the currently loaded Dataset, but also served double-duty as its own data loader. This change moves loading logic out of `Dataset` and into a new `JsonDatasetLoader` class with several helper methods.

Doing this will allow us to make many different kinds of loaders in the future that support different source formats!

*Estimated review size: large, 40-50 minutes (mostly copied/moved code, faster if comparing)*

Solution
========
- Moved loading logic out of `Dataset` and into the new `JsonDatasetLoader`.
  - `JsonDatasetLoader.open()` returns a `Dataset` promise.
- Moved several functions into the new `dataset_loader_utils.ts` helper file.
- Removed unused fields from `Dataset`, changed default value for missing data arrays to `null` instead of `undefined`.
- Updated unit tests.

Misc:
- Fixed a bug in `channel_slice.test.ts` where, at low probabilities, an unserializable `-0` value would be returned.
- Added fallbacks for determining the number of frames in a dataset using `times` data.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-930/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&dataset=Medium
2. The dataset should load and be interactive. Feel free to click around the UI.

-----------------------
1. `JsonDatasetLoader.ts` (compare with `Dataset.ts`)
2. `dataset_loader_utils.ts` (compare with `Dataset.ts`)

### DIFF
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -1,7 +1,7 @@
 import { DEFAULT_COLLECTION_FILENAME, DEFAULT_DATASET_FILENAME } from "src/colorizer/constants";
 import JsonDatasetLoader from "src/colorizer/dataset_loaders/JsonDatasetLoader";
 
-import Dataset from "./Dataset";
+import type Dataset from "./Dataset";
 import type { IArrayLoader, ITextureImageLoader } from "./loaders/ILoader";
 import { FilePathResolver, type IPathResolver, UrlPathResolver } from "./path_resolvers";
 import { LoadErrorMessage, LoadTroubleshooting, type ReportWarningCallback } from "./types";
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -223,32 +223,34 @@ export default class Collection {
     const path = this.getAbsoluteDatasetPath(datasetKey);
     console.log(`Fetching dataset from path '${path}'`);
 
+    let result: DatasetLoadResult;
+    const datasetLoader = new JsonDatasetLoader(path, {
+      frameLoader: options.frameLoader,
+      arrayLoader: options.arrayLoader,
+      reportProgress: options.onLoadProgress,
+      reportWarning: options.reportWarning,
+      manifestLoader: options.manifestLoader,
+      pathResolver: this.pathResolver,
+    });
     try {
-      const datasetLoader = new JsonDatasetLoader(path, {
-        frameLoader: options.frameLoader,
-        arrayLoader: options.arrayLoader,
-        reportProgress: options.onLoadProgress,
-        reportWarning: options.reportWarning,
-        manifestLoader: options.manifestLoader,
-        pathResolver: this.pathResolver,
-      });
       const dataset = await datasetLoader.open();
-      datasetLoader.dispose();
       console.timeEnd("loadDataset");
-      return { loaded: true, dataset: dataset };
+      result = { loaded: true, dataset: dataset };
     } catch (e) {
       console.timeEnd("loadDataset");
       console.error(e);
       if (e instanceof Error) {
-        return {
+        result = {
           loaded: false,
           dataset: null,
           errorMessage: e.message,
         };
       } else {
-        return { loaded: false, dataset: null };
+        result = { loaded: false, dataset: null };
       }
     }
+    datasetLoader.dispose();
+    return result;
   }
 
   public dispose(): void {
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -1,4 +1,5 @@
 import { DEFAULT_COLLECTION_FILENAME, DEFAULT_DATASET_FILENAME } from "src/colorizer/constants";
+import JsonDatasetLoader from "src/colorizer/dataset_loaders/JsonDatasetLoader";
 
 import Dataset from "./Dataset";
 import type { IArrayLoader, ITextureImageLoader } from "./loaders/ILoader";
@@ -222,28 +223,16 @@ export default class Collection {
     const path = this.getAbsoluteDatasetPath(datasetKey);
     console.log(`Fetching dataset from path '${path}'`);
 
-    let totalLoadItems = 0;
-    let completedLoadItems = 0;
-    const onLoadStart = (): void => {
-      totalLoadItems++;
-    };
-    const onLoadComplete = (): void => {
-      completedLoadItems++;
-      options.onLoadProgress?.(completedLoadItems, totalLoadItems);
-    };
-
     try {
-      const dataset = new Dataset(path, {
+      const datasetLoader = new JsonDatasetLoader(path, {
         frameLoader: options.frameLoader,
         arrayLoader: options.arrayLoader,
-        pathResolver: this.pathResolver,
-      });
-      await dataset.open({
-        onLoadStart,
-        onLoadComplete,
+        reportProgress: options.onLoadProgress,
         reportWarning: options.reportWarning,
         manifestLoader: options.manifestLoader,
+        pathResolver: this.pathResolver,
       });
+      const dataset = await datasetLoader.open();
       console.timeEnd("loadDataset");
       return { loaded: true, dataset: dataset };
     } catch (e) {
--- a/src/colorizer/Collection.ts
+++ b/src/colorizer/Collection.ts
@@ -233,6 +233,7 @@ export default class Collection {
         pathResolver: this.pathResolver,
       });
       const dataset = await datasetLoader.open();
+      datasetLoader.dispose();
       console.timeEnd("loadDataset");
       return { loaded: true, dataset: dataset };
     } catch (e) {
--- a/src/colorizer/Dataset.ts
+++ b/src/colorizer/Dataset.ts
@@ -304,10 +304,7 @@ export default class Dataset {
 
   public get numObjects(): number {
     const featureData = this.getFeatureData(this.featureKeys[0]);
-    if (!featureData) {
-      throw new Error("Dataset.numObjects: The first feature could not be loaded. Is the dataset manifest file valid?");
-    }
-    return featureData.data.length;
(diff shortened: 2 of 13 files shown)


src/state/slices/dataset_slice.ts
```diff
@@ -78,7 +78,7 @@ export const createDatasetSlice: StateCreator<CollectionSlice & DatasetSlice, []
     let featureKey = get().featureKey;
     if (featureKey === null || !dataset.hasFeatureKey(featureKey)) {
       // Set to default feature key of the new dataset
-      featureKey = dataset.featureKeys[0];
+      featureKey = dataset.featureKeys[0] ?? null;
```
ShrimpCryptid, line 81: I removed a hard check during Dataset loading that requires at least one feature to be present, so it's very very unlikely but possible that `dataset.featureKeys[0]` is `undefined`. This just casts it to `null`, which is the default `featureKey` value.

I did confirm that the UI still works as expected when no features are present!

tests/state/ViewerState/channel_slice.test.ts
```diff
@@ -346,7 +346,11 @@ describe("ChannelSlice", () => {
 
     function getRandomSerializableNumber(): number {
       // Must have no more than 3 digits of precision
-      return Math.round(Math.random() * 200_000 - 100_000) / 1000;
+      const value = Math.round(Math.random() * 200_000 - 100_000) / 1000;
```
ShrimpCryptid, line 349: I ran into a very unlikely bug here and tracked it down to this function. For any value `n` in the range `(-0.5, 0)`, `Math.round(n) = -0`. 

During serialization, `-0` is cast to `0` and the tests fail when comparing the two values. The change here casts all `-0` values to `0` early for testing.

src/colorizer/Dataset.ts
```diff
@@ -98,22 +79,18 @@ const defaultMetadata: ManifestFileMetadata = {
   startTimeSeconds: 0,
 };
 
-const MAX_CACHED_FRAME_BYTES = 500_000_000; // 500 MB
-const MAX_CACHED_BACKDROPS_BYTES = 500_000_000; // 500 MB
-
 export default class Dataset {
   private frameLoader: ITextureImageLoader;
-  private frameFiles?: string[];
```
ShrimpCryptid, line 106: All fields are now declared in the constructor, so I've adjusted the types so `undefined` is no longer a possible value.

src/colorizer/Collection.ts
```diff
@@ -222,28 +223,16 @@ export default class Collection {
...
 
-    let totalLoadItems = 0;
-    let completedLoadItems = 0;
-    const onLoadStart = (): void => {
-      totalLoadItems++;
-    };
-    const onLoadComplete = (): void => {
-      completedLoadItems++;
-      options.onLoadProgress?.(completedLoadItems, totalLoadItems);
-    };
```
ShrimpCryptid, line 233: Progress is now reported directly by the loader.

tests/constants/index.ts
```diff
@@ -157,6 +157,21 @@ export const MOCK_DATASET_ARRAY_LOADER_DEFAULT_SOURCE = makeMockDatasetArrayLoad
...
+  ["tracks.json", MOCK_DATASET_TRACKS],
+  ["feature1.json", MOCK_DATASET_FEATURE_1],
+  ["feature2.json", MOCK_DATASET_FEATURE_2],
+  ["feature3.json", MOCK_DATASET_FEATURE_3],
+  ["feature4.json", MOCK_DATASET_FEATURE_4],
+  ["centroids.json", MOCK_DATASET_CENTROIDS],
+];
+export const MOCK_DATASET_FILEMAP = Object.fromEntries(
+  fileToData.map(([filename, data]) => [filename, new File([JSON.stringify(data)], filename)])
+);
```
ShrimpCryptid, line 173: Moved the mock filemap from `Collection.test.ts` and made it more robust.

src/colorizer/dataset_loaders/dataset_loader_utils.ts
```diff
@@ -0,0 +1,137 @@
+import type { Vector2 } from "three";
```
ShrimpCryptid, line 1: Contents copied from `Dataset`

src/colorizer/dataset_loaders/dataset_loader_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+    const loadedFeatureNames = new Set(Array.from(loadedFeatures.values()).map((f) => f.name));
+    const missingFeatureNames = featureSpec.filter((f) => !loadedFeatureNames.has(f.name)).map((f) => f.name);
+
+    reportWarning?.("Some features failed to load.", [
+      "The following feature(s) could not be loaded and will not be shown: ",
+      ...formatAsBulletList(missingFeatureNames, 5),
+      LoadTroubleshooting.CHECK_FILE_OR_NETWORK,
+    ]);
+  }
+}
```
ShrimpCryptid, line 137: Adapted from logic in `Dataset`.

src/colorizer/dataset_loaders/JsonDatasetLoader.ts
```diff
@@ -0,0 +1,424 @@
+import { RGBAFormat, RGBAIntegerFormat, Vector2 } from "three";
```
ShrimpCryptid, line 1: Contents of this file are largely copied from `Dataset`!

src/colorizer/Dataset.ts
```diff
@@ -60,10 +48,11 @@ export type FeatureData = {
 
 type BackdropData = {
   name: string;
-  frames: string[];
+  frames: (string | null)[];
```
ShrimpCryptid, line 51: Backdrops and frames are now expected to have fully resolved paths (so instead of "./frame0.png" it should be "https://my-url.com/frame0.png", for example). If a path can't be resolved (usually only happens when loading files), the returned path is `null`.

`Dataset` already handles `null`, I just removed the step where it has to resolve the paths too.

src/colorizer/dataset_loaders/JsonDatasetLoader.ts
```diff
@@ -0,0 +1,440 @@
...
+    units: "",
+  },
+  frameDurationSeconds: 0,
+  startTimeSeconds: 0,
+};
+
+/**
+ * Loads a dataset from a manifest JSON file.
+ */
+export default class JsonDatasetLoader {
```
toloudis, line 52: should this extend some abstract interface so as to have all loaders implement the same basic API?

toloudis (approved):
LGTM, just curious to see the dataset loader api isolated somewhere.

rugeli:
### Code review

Found 1 issue:

1. GPU texture leak in `getFrameDims`: `this.frameLoader.load()` returns a Three.js `Texture` (a GPU resource) that is used only to read `image.width/height` and is then discarded — never `.dispose()`d and never inserted into the `DataCache`. This leaks one frame-sized GPU allocation per dataset load. Previously `Dataset.open()` obtained frame 0 via `loadFrame(0)`, which cached the texture in `this.frames` so it was disposed with the dataset. (Copilot flagged the `UrlArrayLoader` disposal leak, but not this one.) Fix: call `result.dispose()` after reading the dimensions (or seed the `DataCache` with frame 0 to also avoid re-decoding it on first render).

https://github.com/allen-cell-animated/timelapse-colorizer/blob/fdd67e1db6d45e82e90c6bde39ab565e29d95850/src/colorizer/dataset_loaders/JsonDatasetLoader.ts#L183-L186

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rugeli:
> ### Code review
> Found 1 issue:
> 
> 1. GPU texture leak in `getFrameDims`: `this.frameLoader.load()` returns a Three.js `Texture` (a GPU resource) that is used only to read `image.width/height` and is then discarded — never `.dispose()`d and never inserted into the `DataCache`. This leaks one frame-sized GPU allocation per dataset load. Previously `Dataset.open()` obtained frame 0 via `loadFrame(0)`, which cached the texture in `this.frames` so it was disposed with the dataset. (Copilot flagged the `UrlArrayLoader` disposal leak, but not this one.) Fix: call `result.dispose()` after reading the dimensions (or seed the `DataCache` with frame 0 to also avoid re-decoding it on first render).
> 
> https://github.com/allen-cell-animated/timelapse-colorizer/blob/fdd67e1db6d45e82e90c6bde39ab565e29d95850/src/colorizer/dataset_loaders/JsonDatasetLoader.ts#L183-L186
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

oops first time trying the Claude code review plugin and I didn't realize it would post this comment. I'm also unsure how important that "GPU texture leak" is lol, feel free to skip it or remove it from the thread.

rugeli (approved):
This looks like a clean split! and after reading through the linked ticket, the refactor makes sense as setup for the next single file datasets feature step. 

(I'll leave the issue claude flagged to your judgment)

src/colorizer/dataset_loaders/JsonDatasetLoader.ts
ShrimpCryptid: Oops, thought I responded to this! I add this in the next PR in this series :>

PR #931, opened 2026-06-04
refactor: Split math_utils into many smaller subfiles

ShrimpCryptid:
Problem
=======
Hi! This is a quick file move based on a suggestion by Dan in #924. 

Code is largely unchanged besides file import updates, so I'm just asking for a rubber-stamp here. Please let me know if the file names make sense to you based on their contents.

 (Update: I did update some method comments and method names based on Copilot feedback. Functionality is unchanged!)

Solution
========
- Split `math_utils.ts` into the `math_utils` folder with many smaller subfiles.
  - Added a barrel/index file so existing imports are largely unaffected.

Key Files
=======
- `src/colorizer/utils/math_utils/index.ts`


src/colorizer/utils/math_utils/correlation.ts
ShrimpCryptid, line 1: The correlation file was also moved into `math_utils` because it only had one math-related function inside.

src/colorizer/workers/worker.ts
```diff
@@ -2,18 +2,18 @@ import { Transfer, worker } from "workerpool";
...
+import { calculateMotionDeltas } from "src/colorizer/utils/math_utils/motion_deltas";
 import {
   averageVectorFlowField,
   binAndSumFeatureVectors,
-  calculateMotionDeltas,
-  constructAllTracksFromData,
   filterVectorFlowFieldData,
   kernelSmoothVectorFlowField,
   make1dGaussianKernel,
-} from "src/colorizer/utils/math_utils";
```
ShrimpCryptid, line 17: I split this import into several smaller imports from the corresponding submodules. This helps to scope imports the narrowest set needed, and avoids circular dependencies that can cause the worker build to fail.

rugeli (approved):
nice work splitting the big util file into focused sub-files!

toloudis (approved):
LGTM!

PR #935, opened 2026-06-11
fix: Added back behavior where flow field plot threshold scaled with bin

ShrimpCryptid:
Problem
=======
Fixes a behavior that was lost during merging for the 3D flow field plot.

Solution
========
- Added scaling of threshold with bin count.

## Type of change

### DIFF
--- a/src/components/Tabs/Plot3d/Plot3dTab.tsx
+++ b/src/components/Tabs/Plot3d/Plot3dTab.tsx
@@ -10,8 +10,8 @@ import { useInteractionListener } from "src/hooks";
 import { useViewerStateStore } from "src/state";
 import { FlexColumn } from "src/styles/utils";
 
-import Plot3d from "./Plot3d";
 import { make3dConeTrace } from "./plot_3d_utils";
+import Plot3d from "./Plot3d";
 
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Plot3d/Plot3dTab.tsx
```diff
@@ -182,11 +182,11 @@ export default function Plot3dTab(): ReactElement {
         make3dConeTrace(vectorFieldData, {
           coneSize,
           colorRamp: coneColorRamp,
-          threshold,
+          threshold: threshold * (10 / bins),
```
toloudis: why 10?

ShrimpCryptid: I'll add this as a comment:

10 is the minimum number of bins that's allowed, so at 10 bins, the value is directly thresholding the integer number of deltas that fell into this bin. At higher bin counts, dividing by the number of bins keeps the thresholding proportionally the same (this is more relevant when gaussian smoothing is applied).

toloudis (approved):
thanks!

ascibisz (approved):

Issue #947, opened 2026-07-01
Lineage: Comparing multiple tracking graphs


ShrimpCryptid:
## Use Case

When comparing tracking algorithms, Max wants to compare and see the results.
- Segmentations/objects would not change*, just the track assignment.
- Each tracking result contains an array of node edges, track edges, and track IDs.

There are also possible cases where segmentations/objects would change. This would either need to be a separate ticket or a user would need to make two different datasets and load them in windows side-by-side.

## Acceptance Criteria

- [x] Change TFE's Dataset format to support multiple alternate track results (especially track IDs). Fetching track ID values will need to be keyed.
- [ ] Add a control in the viewport to switch between track results.
- [ ] In the lineage view, add a button for switching to a split view.
  - [ ] Each split view would have a dropdown for selecting the track result to visualize.
  - [ ] Update the views based on current frame -> lookup ID and then corresponding track for each track result?

## Details

_Please provide any helpful specifications_

Issue #945, opened 2026-07-01
Lineage: Load + visualize track-level lineage data


ShrimpCryptid:
## Use Case

Lineage data in this case is the relationships between tracks (split + merge).

## Acceptance Criteria

- [x] Accept a parquet file in the TFE dataset that provides a `track_start` and `track_end` column, where each represents an edge between tracks.
- [x] Add a tab for lineage visualization.
  - [x] Within the tab, show each track as a node with edges drawn between them in a tree hierarchy. 
  - [x] Clicking on a track should select it and jump to the corresponding time in the TFE viewer if the track is not currently visible at the current frame.

## Details

_Please provide any helpful specifications_

PR #950, opened 2026-07-01
feat: Add lineage data to Dataset spec (lineage pt. 1)

ShrimpCryptid:
Problem
=======
Part 1 of 2ish for #945, "load + visualize track-level lineage data"! This change updates the TFE Dataset manifest spec. 

One of Max's potential use cases is to use TFE to compare the output of one or more tracking algorithms run on the same segmentations (see #947). Each output will contain track IDs, relationships between tracks (track edges), and relationships between nodes in the tracks (node edges). 

Before, `tracks` was provided as a single string path to a JSON or parquet file.

```json
{
  // ...
  "tracks": "tracks.parquet",
}
```

To support this, `tracks` now is an array of objects. The intention is to make it possible in the future to support multiple alternate track outputs or even allow for track editing.

```json
{
  // ...
  "tracks": [
       {
           "name": "Tracks 1",
           "key": "tracks_1",
           "description": "Output from tracking algorithm A",
           "tracks": "tracks.parquet",
           "trackEdges": "track_edges.parquet",
           "nodeEdges": "node_edges.parquet",
       }
   ]
}
```

Solution
========
- Changed `Dataset` to expect a map of `TrackData` objects instead of a `trackIds` array.
  - For compatibility for now, the `trackId` field is populated from the first element of the map. If we give users the ability to switch between multiple track views, we will need to store the current track view's key in state.
- Updated `dataset_utils.ts` with a fixup from the old format to the new format.
- Updated `JsonDatasetLoader.ts` to handle the new track data loading.
- Updated unit tests.

## Type of change

Steps to Verify:
----------------
1. Open PR preview, which links to a new dataset with the tracks object array. Tracks should load in correctly: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-950/viewer?dataset=https://vast-files.int.allencell.org/users/peyton.lee/liberali-data/tfe/manifest.json&t=270&track=21
3. Open PR preview, which links to our old datasets. The tracks will also still load in correctly: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-950/viewer?collection=https%3A%2F%2Fallencell.s3.amazonaws.com%2Faics%2Fnuc-morph-dataset%2Ftimelapse_feature_explorer_datasets%2Fexploratory_dataset%2Fcollection.json&t=195&track=71933


src/colorizer/dataset_loaders/JsonDatasetLoader.ts
```diff
@@ -112,6 +113,45 @@ export default class JsonDatasetLoader {
...
+    } else if (trackEdgesResult.status === "rejected") {
+      console.warn(`Track ${index}: Failed to load track edges: ${trackEdgesResult.reason}`);
+      return undefined;
+    } else if (nodeEdgesResult.status === "rejected") {
+      console.warn(`Track ${index}: Failed to load node edges: ${nodeEdgesResult.reason}`);
+      return undefined;
+    }
+    const tracks = tracksResult.status === "fulfilled" ? tracksResult.value : undefined;
+    const trackEdges = trackEdgesResult.status === "fulfilled" ? trackEdgesResult.value : undefined;
+    const nodeEdges = nodeEdgesResult.status === "fulfilled" ? nodeEdgesResult.value : undefined;
```
frasercl: Looks like TS is correctly narrowing the type of these results to `PromiseFulfilledResult` off of the `if`/`return`s above, so these ternaries might not be necessary. If the worry here is forwards-compatibility with other possible variants of `PromiseSettledResult.status`, the `if`s could also be changed to  `... !== "fulfilled"` (though I very seriously doubt the interface would be changed).

frasercl (approved):

src/colorizer/dataset_loaders/JsonDatasetLoader.ts
```diff
@@ -112,6 +113,45 @@ export default class JsonDatasetLoader {
...
+    index: number
+  ): Promise<TrackData | undefined> {
+    const promises = [];
+    promises.push(this.reportLoadProgress(this.loadToBuffer(FeatureDataType.U32, metadata.trackIds)));
+    promises.push(this.reportLoadProgress(this.loadToBuffer(FeatureDataType.U32, metadata.trackEdges)));
+    promises.push(this.reportLoadProgress(this.loadToBuffer(FeatureDataType.U32, metadata.nodeEdges)));
+
+    if (promises.length === 0) {
+      console.warn(`Track ${index}: No data files specified for track edges.`);
+      return undefined;
```
rugeli: is this `if` condition ever reachable? Looks like the promises get pushed unconditionally, so the `promise.length` is always 3 even when the metadata is undefined. should the metadata fields be null checked first? (Oh I saw you resolved a similar copilot comment on this, so maybe they're validated somewhere that I missed?)

rugeli (approved):
just one question about the guard, and looks like you already have an answer. Otherwise, lgtm!

src/colorizer/dataset_loaders/JsonDatasetLoader.ts
ShrimpCryptid: Ah, this was a holdover from when I was originally conditionally pushing promises only for defined files. Good catch!

PR #951, opened 2026-07-01
fix: Load initial viewer state before React render (lineage pt. 2a)

ShrimpCryptid:
Tiny fix. Initial viewer settings are now pulled from the URL before the first render, which prevents the UI from jumping when the dataset is loaded.

This also fixes a bug in a follow-up PR when making changes to the visibility of tabs on the UI.


meganrm (approved):

toloudis (approved):
LGTM

PR #954, opened 2026-07-02
fix: Fixed React startup errors due to misconfigured components

ShrimpCryptid:
Problem
=======
Fixes a number of errors shown in the console on TFE startup.

*Estimated review size: tiny, 5-10 minutes*

Solution
========
- Added ref forwarding to `IconButton.tsx`.
- Added ref forwarding to `TooltipWithSubtitle` and changed it to a default export.
  - Updated component imports.
- Replaced a deprecated prop in `AccessibleDropdown.tsx`, and renamed it to `KeyboardNavigableDropdown.tsx`.
- Fixed React static flag warnings in `LoadingSpinner.tsx` caused by incorrect hook usage.

## Type of change

Screenshots (optional):
-----------------------
| Before | After | 
| --- | --- |
| {180B07CF-71F3-42E2-98C5-5129A07A8A1F} | {B47786D7-AE60-4F59-9B43-B0C11F5FD261} |

(The two remaining errors are due to the cookie banner having problems when running in a sandboxed environment)

### DIFF
--- a/src/components/Buttons/IconButton.tsx
+++ b/src/components/Buttons/IconButton.tsx
@@ -144,7 +144,7 @@ const StyledButton = styled(Button)<{ $type: IconButtonProps["type"]; $sizePx?: 
  */
 const IconButton = forwardRef<HTMLButtonElement, PropsWithChildren<IconButtonProps>>(function IconButton(
   props: PropsWithChildren<IconButtonProps>,
-  ref
+  ref: React.Ref<HTMLButtonElement>
 ): ReactElement {
   const themeContext = useContext(AppThemeContext);
 
--- a/src/components/Buttons/IconButton.tsx
+++ b/src/components/Buttons/IconButton.tsx
@@ -1,5 +1,5 @@
 import { Button, ConfigProvider } from "antd";
-import React, { type PropsWithChildren, type ReactElement, useContext } from "react";
+import React, { forwardRef, type PropsWithChildren, type ReactElement, useContext } from "react";
 import styled, { css } from "styled-components";
 
 import { AppThemeContext } from "src/styles/AppStyle";
@@ -139,15 +139,19 @@ const StyledButton = styled(Button)<{ $type: IconButtonProps["type"]; $sizePx?: 
  *
  * @example
  * <IconButton onClick={myClickHandler}>
- *    <PauseOutlined /> // From Antd Icons
+ *    <PauseOutlined />
  * </IconButton>
  */
-export default function IconButton(props: PropsWithChildren<IconButtonProps>): ReactElement {
+const IconButton = forwardRef<HTMLButtonElement, PropsWithChildren<IconButtonProps>>(function IconButton(
+  props: PropsWithChildren<IconButtonProps>,
+  ref
+): ReactElement {
   const themeContext = useContext(AppThemeContext);
 
   return (
     <ConfigProvider theme={{ components: { Button: { colorPrimaryActive: themeContext.color.button.hover } } }}>
       <StyledButton
+        ref={ref}
         type="primary"
         $type={props.type || "primary"}
         $sizePx={props.sizePx}
@@ -166,4 +170,5 @@ export default function IconButton(props: PropsWithChildren<IconButtonProps>): R
       </StyledButton>
     </ConfigProvider>
   );
-}
+});
+export default IconButton;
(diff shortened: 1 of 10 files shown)


src/components/Dropdowns/KeyboardNavigableDropdown.tsx
```diff
@@ -149,10 +149,10 @@ const DropdownContentWrapper = styled.div`
 `;
 
 /**
- * A keyboard-accessible wrapper around the Antd `Dropdown` component, with support for button styling, labels, and tooltips.
+ * A keyboard-navigable wrapper around the Antd `Dropdown` component, with support for button styling, labels, and tooltips.
```
ShrimpCryptid, line 152: I think there are a bunch of other ARIA behaviors that we're not currently handling, so I feel a little self-conscious saying this is a fully accessible component. Hence the new name!

src/components/Dropdowns/KeyboardNavigableDropdown.tsx
```diff
@@ -266,7 +270,7 @@ export default function AccessibleDropdown(inputProps: AccessibleDropdownProps):
         disabled={props.disabled}
         open={dropdownOpenProp}
         getPopupContainer={componentContainerRef.current ? () => componentContainerRef.current! : undefined}
-        dropdownRender={(_menus: ReactNode) => {
+        popupRender={(_menus: ReactNode) => {
```
ShrimpCryptid, line 273: Bugfix was here, removing a deprecated prop.

src/components/Buttons/IconButton.tsx
```diff
@@ -1,5 +1,5 @@
 import { Button, ConfigProvider } from "antd";
-import React, { type PropsWithChildren, type ReactElement, useContext } from "react";
+import React, { forwardRef, type PropsWithChildren, type ReactElement, useContext } from "react";
```
ShrimpCryptid, line 2: Added ref forwarding. `IconButton` was sometimes used with Ant components that wrap other components, like Tooltips and Popover components, which expect to be able to get a ref to a child Button or other HTML Element. `IconButton` wasn't forwarding the ref, so the wrapping component was falling back to using `findDOMNode` which throws errors in React strict mode.

src/components/Tooltips/TooltipWithSubtitle.tsx
```diff
@@ -1,5 +1,6 @@
 import { Tooltip, type TooltipProps } from "antd";
-import React, { type ReactElement, type ReactNode, useContext } from "react";
+import type { TooltipRef } from "antd/es/tooltip";
+import React, { forwardRef, type ReactElement, type ReactNode, useContext } from "react";
```
ShrimpCryptid, line 3: Also added ref forwarding here

src/components/Tooltips/TooltipWithSubtitle.tsx
```diff
@@ -37,4 +42,5 @@ export function TooltipWithSubtitle(props: TooltipWithSubtitleProps): ReactEleme
       {props.children}
     </Tooltip>
   );
-}
+});
+export default TooltipWithSubtitle;
```
ShrimpCryptid, line 46: Changed this to a default export, which necessitated a bunch of import updates

src/components/LoadingSpinner.tsx
```diff
@@ -92,7 +92,8 @@ export default function LoadingSpinner(inputProps: PropsWithChildren<LoadingSpin
   // Delay showing progress bar slightly; this fixes a visual bug where the loading spinner
   // would flash the progress bar at 100% right as it vanished.
   const canShowProgressBar = props.progress !== undefined && props.progress !== null;
-  const showProgressBar = canShowProgressBar && useDebounce(canShowProgressBar, VANISH_DURATION_MS);
```
ShrimpCryptid, line 95: bug: `useDebounce` hook was called conditionally. Hooks should be at the top-level of the function component and always called.

meganrm (approved):

src/components/Dropdowns/HelpDropdown.tsx
```diff
@@ -86,7 +86,7 @@ export default function HelpDropdown(): ReactElement {
 
   return (
     <div>
-      <AccessibleDropdown
+      <KeyboardNavigableDropdown
```
rugeli, line 89: the new name feels much clear!

rugeli (approved):

PR #956, opened 2026-07-06
fix: Fixed unnormalized scaling in TFE 2D canvas calculations

ShrimpCryptid:
Problem
=======
Fixes a bug where the scale bar in TFE was incorrectly scaled on the newer Endo datasets. This was due to a calculation error for datasets where the frames had differing aspect ratio from the default TFE viewport.

Solution
========
- Fixed inversed calculation of frame scaling constants in `viewport/utils.ts`.

## Type of change

Steps to Verify:
----------------
1. Open broken main build. The frame should be 660 micrometers wide, but according to the scalebar it is only 330ish.  https://timelapse.allencell.org/viewer?dataset=https%3A%2F%2Fvast-files.int.allencell.org%2Fusers%2Fpeyton.lee%2Ftest-data%2Fjess-endo-scaling%2Fmanifest.json
2. Open the same dataset on the fixed build. The scalebar will show that the frame is roughly 660 micrometers wide. https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-956/viewer?dataset=https%3A%2F%2Fvast-files.int.allencell.org%2Fusers%2Fpeyton.lee%2Ftest-data%2Fjess-endo-scaling%2Fmanifest.json

I've also tested and visually verified that the scaling of other objects drawn with these values (track paths, vectors, etc.) are correct.

Screenshots (optional):
-----------------------

The frame should be ~667 micrometers wide.

| Before ❌ | After ✅ |
| --- | --- |
| image  | image |

### DIFF
--- a/src/colorizer/viewport/utils.ts
+++ b/src/colorizer/viewport/utils.ts
@@ -23,9 +23,9 @@ export function get2DCanvasScaling(
   if (canvasAspect > frameAspect) {
     // Canvas has a wider aspect ratio than the frame, so proportional height is
     // 1 and we scale width accordingly.
-    unscaledFrameSizeInCanvasCoords.x = canvasAspect / frameAspect;
+    unscaledFrameSizeInCanvasCoords.x = frameAspect / canvasAspect;
   } else {
-    unscaledFrameSizeInCanvasCoords.y = frameAspect / canvasAspect;
+    unscaledFrameSizeInCanvasCoords.y = canvasAspect / frameAspect;
   }
 
   // Get final size by applying the current zoom level, where `zoomMultiplier=2`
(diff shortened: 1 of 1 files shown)


src/colorizer/viewport/utils.ts
```diff
@@ -23,9 +23,9 @@ export function get2DCanvasScaling(
   if (canvasAspect > frameAspect) {
     // Canvas has a wider aspect ratio than the frame, so proportional height is
     // 1 and we scale width accordingly.
-    unscaledFrameSizeInCanvasCoords.x = canvasAspect / frameAspect;
+    unscaledFrameSizeInCanvasCoords.x = frameAspect / canvasAspect;
   } else {
-    unscaledFrameSizeInCanvasCoords.y = frameAspect / canvasAspect;
+    unscaledFrameSizeInCanvasCoords.y = canvasAspect / frameAspect;
   }
```
ShrimpCryptid, line 29: The math here was inverted from what it should be. 


Image

TFE attempts to show the frame as large as possible within a fixed aspect ratio viewport. In the case where the canvas is wider than the frame (e.g. `canvasAspect > frameAspect`), the height scale should be 1, and the width scale should be less than `1`. 

With the old math, which was `canvasAspect / frameAspect`, the calculated `x` scale would incorrectly  be greater than `1`, since `canvasAspect` is larger. 

I've flipped these, which corrected the scalebar issue.

frasercl (approved):

BrianWhitneyAI (approved):

PR #959, opened 2026-07-07
feat: Show track lineage tab (lineage pt. 2b)

### LINKED ISSUE #945: Lineage: Load + visualize track-level lineage data

ShrimpCryptid:
Problem
=======
Closes #945, "Lineage: Load + visualize track-level lineage data". 

This PR adds a new tab for lineage data, which is shown only if track edges or node edges are provided. The tab adds a tree view of the parent-child relationships between tracks in the dataset.

Much of the code was adapted from an initial solution created by Max Hess (@MaksHess), found here: https://github.com/MaksHess/d3-tree/

*Estimated review size: large, 50-60 minutes*

Solution
========
- Adds a new `LineageTab` that displays a `TreeLineageView`. In the future, multiple different view types can be supported.
  - Tracks can be clicked to select them from the tree lineage view, with support for multi-track selection.
  - Information about the track is shown when the track is hovered.
  - The lineage tab is hidden if no lineage data is provided.
- Added helper methods for calculating lineage data and lineage relationships form track edge data introduced in #950.

Misc:
- Moved `TooltipCard.tsx` out of `CanvasHoverTooltip.tsx`

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-959/viewer?dataset=https%3A%2F%2Fvast-files.int.allencell.org%2Fusers%2Fpeyton.lee%2Fliberali-data%2Ftfe%2Fmanifest.json&tab=lineage
2. Interact with the lineage tree view by clicking on tracks to select them.

Screenshots (optional):
-----------------------

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/9b4fc419-b188-49d9-812a-f05d8fdc3707

{FDDB460A-23AF-4F32-9CA1-6E5995117FB1}

-----------------------
1. `src/components/Tabs/Lineage/types.ts`
2. `lineage_utils.ts`
3. `LineageTab.tsx`
4. `TreeLineageView.tsx`


src/components/Tooltips/CanvasHoverTooltip.tsx
```diff
@@ -13,26 +13,14 @@ import { AppThemeContext } from "src/styles/AppStyle";
...
 import HoverTooltip from "./HoverTooltip";
+import { TooltipCard } from "./TooltipCard";
 
 type CanvasHoverTooltipProps = {
   lastValidHoveredId: PixelIdInfo;
   showObjectHoverInfo: boolean;
   annotationState: AnnotationState;
 };
 
-const ObjectInfoCard = styled.div`
```
ShrimpCryptid, line 23: Moved into `TooltipCard.tsx`

src/components/Tabs/Lineage/lineage_utils.ts
```diff
@@ -0,0 +1,103 @@
...
+  const trackIds = Array.from(data.trackIdToTrackInfo.keys());
+  const idToChildren = new Map<number, number[]>(trackIds.map((id) => [id, []]));
+  const idToChildrenRenderable = new Map<number, number[]>(trackIds.map((id) => [id, []]));
+  const idToParents = new Map<number, number[]>(trackIds.map((id) => [id, []]));
+
+  /**
+   * Edges to a node where the node already has a parent (i.e. edges that would
+   * create the second/nth parent of a merge node).
+   */
+  const multiparentEdges: [number, number][] = [];
```
ShrimpCryptid, line 66: Image

Multiparent edges are like the ones shown in dotted-orange here.

ShrimpCryptid: If we didn't have special handling for these kinds of nodes, 208 and 210 would both have separate 419 nodes after them, which duplicates every following node as well (420 and 421).

src/components/Tabs/Lineage/LineageTab.tsx
```diff
@@ -0,0 +1,145 @@
+import * as d3 from "d3";
```
ShrimpCryptid, line 1: Wraps the TreeLineageView, and adds a hover tooltip: 

Image

src/components/Tabs/Lineage/LineageViews/TreeLineageView.tsx
```diff
@@ -0,0 +1,264 @@
...
+    rootNode = trackIdToTrackInfo.get(rootNodeIds[0])!;
+  } else {
+    // Multiple root nodes, make a dummy root node that is the parent of all root nodes
+    rootNode = { id: DUMMY_ROOT_NODE_ID, length: 0, startTime: 0 };
+    // Add dummy track info for the dummy root node
+    trackIdToTrackInfo.set(rootNode.id, rootNode);
+    idToChildren.set(rootNode.id, rootNodeIds);
+  }
+
+  const root = d3.hierarchy<TrackInfo>(
```
toloudis, line 58: Curious about this chunk of code (lines 58-70) ...is this assembling a tree?  is this a function you need to benchmark for performance/scalability?  Would there be any way to make it faster?  And could it be done without using built-in d3 commands? (Like, could you build it yourself into some simple data structure and then hand that over to d3)

toloudis (approved):
Approved! although I have a sinking feeling that most of it changes in the next few PRs...

src/components/Tabs/Lineage/lineage_utils.ts
```diff
@@ -0,0 +1,103 @@
+import * as d3 from "d3";
+
+import type { Dataset } from "src/colorizer";
+
+import type { LineageData, LineageDataRelationships, TrackInfo } from "./types";
+
+export function getLineageData(dataset: Dataset): LineageData {
```
toloudis, line 7: this kinda looks like a member function of Dataset

src/components/Tabs/Lineage/LineageViews/TreeLineageView.tsx
ShrimpCryptid: Yes, this builds the tree datastructure for d3 rendering! I don't think that performance is an issue because I only construct this once per dataset, and it's doing a tree traversal where each node is a track (so at max like an O(N) operation over the dataset, but likely much smaller). 

I probably could construct the tree from scratch if performance became an issue (and I actually will be modifying the tree slightly for alignment in part 6 of this PR chain).

ShrimpCryptid:
> Approved! although I have a sinking feeling that most of it changes in the next few PRs...

Not too much, actually! A few calculations get moved outside of this component for reuse.

src/components/Tabs/Lineage/lineage_utils.ts
ShrimpCryptid: Hm, I mostly agree, but I'm also thinking about if we ever allow user edits? If so I don't know if I'd want those edits to be applied directly to the Dataset...

Okay, actually, talking myself out of this. Yeah, I think putting the lineage data in Dataset makes sense, since we're already allowing multiple versions of track data to be stored.

src/components/Tabs/Lineage/LineageViews/TreeLineageView.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  let rootNode: TrackInfo;
+  if (rootNodeIds.length === 0) {
+    console.warn("No root nodes found in lineage data, skipping tree rendering.");
+    return;
+  } else if (rootNodeIds.length === 1) {
+    rootNode = trackIdToTrackInfo.get(rootNodeIds[0])!;
+  } else {
+    // Multiple root nodes, make a dummy root node that is the parent of all root nodes
+    rootNode = { id: DUMMY_ROOT_NODE_ID, length: 0, startTime: 0 };
```
ascibisz, line 52: why do we need this dummy root node? does `d3.hierarchy` require a single root, or does it just makes things easier?

ascibisz (commented):
It looks good and it seems to work well for me! I haven't used d3 before, so I don't know how effective I was at reviewing the TreeLineageView.tsx, but it all seems to work as I would expect

ascibisz (approved):
It looks good and it seems to work well for me! I haven't used d3 before, so I don't know how effective I was at reviewing the TreeLineageView.tsx, but it all seems to work as I would expect

src/components/Tabs/Lineage/LineageViews/TreeLineageView.tsx
ShrimpCryptid: Yup, your assumption was right! d3.hierarchy requires a single root node for a tree.

src/components/Tabs/Lineage/lineage_utils.ts
ShrimpCryptid: I'm going to move this in a future PR because I don't want to propagate the changes across all (eep!) 6 open PRs in this chain. See #969!

PR #961, opened 2026-07-09
refactor: Calculate tree hierarchy outside of TreeLineageView (lineage pt. 4)

ShrimpCryptid:
Problem
=======
Small refactor. This change moves the calculation for the tree hierarchy outside of the TreeLineageView. In a follow-up PR, parts of the tree will be selectively rendered by a different view component, which will also need to calculate a hierarchy. 

Calculating the hierarchy once and passing into both allows them to render with the same branch ordering, and reduces duplicate computation!

Solution
========
- Moved d3 hierarchy calculation from `TreeLineageView` into a new `getTreeHierarchy` function in `lineage_utils`.
- Added a constants file.
- Removed `SharedLineageProps` type, since it was largely unused in future PRs.

Steps to Verify:
----------------
1. Open PR preview and verify that the lineage tree is loading: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-961/viewer?dataset=https://vast-files.int.allencell.org/users/peyton.lee/liberali-data/tfe/manifest.json&tab=lineage


ShrimpCryptid:
My bad, I forgot to update the description with a working PR preview! I updated it with a working link :)

src/components/Tabs/Lineage/lineage_utils.ts
```diff
@@ -101,3 +104,52 @@ export function getDefaultZoomTransform(
...
+  const { idToChildrenRenderable, idToParents } = relationships;
+  const trackIdToTrackInfo = new Map(data.trackIdToTrackInfo);
+  const idToChildren = new Map(idToChildrenRenderable);
+
+  // All nodes with no parents
+  const rootNodeIds = [...idToParents.entries()].filter(([, parents]) => parents.length === 0).map(([id]) => id);
+
+  let rootNode: TrackInfo;
+  if (rootNodeIds.length === 0) {
+    console.warn("No root nodes found in lineage data, skipping tree rendering.");
```
rugeli, line 135: I saw this warning in the console on the preload path, seems like it might be a bug(?) the tree renders fine.

rugeli (approved):
the refactor makes sense, and I see this feature is a work in progress, so it looks good! I only noticed one warning that persists in the console when opening the preload dataset which might worth looking.

ShrimpCryptid:
> the refactor makes sense, and I see this feature is a work in progress, so it looks good! I only noticed one warning that persists in the console when opening the preload dataset which might worth looking.

Oh hmm, that might be because there's no initial track lineage data? I'll take a look.

src/components/Tabs/Lineage/constants.ts
```diff
@@ -0,0 +1 @@
+export const DUMMY_ROOT_NODE_ID = -1;
```
frasercl, line 1: Do you anticipate this file growing beyond this size in the future? If the definition of the item you're importing is as large as the import statement... why not just inline the value?

src/components/Tabs/Lineage/LineageTab.tsx
```diff
@@ -57,6 +57,10 @@ export default function LineageTab(): ReactElement {
   const lineageRelationships = useMemo(() => {
     return getLineageRelationships(lineageData);
   }, [lineageData]);
+  const hierarchy = useMemo(() => {
+    return getTreeHierarchy(lineageData, lineageRelationships);
+  }, [lineageData, lineageRelationships]);
```
frasercl, line 62: Now THIS is what `useMemo` was made for!

Minor/optional thing that I can't comment on at the actual site of the potential code change: because `EMPTY_LINEAGE_DATA` gets a `satisfies` rather than a proper type annotation, the type of `lineageData` is inferred as a slightly awkward union instead of just `LineageData`. Is there a reason it can't just be typed as a `LineageData` instead?

src/components/Tabs/Lineage/LineageTab.tsx
```diff
@@ -113,9 +117,10 @@ export default function LineageTab(): ReactElement {
 
   const selectedTracks = useMemo(() => new Set(tracks.keys()), [tracks]);
 
-  const lineageViewProps: SharedLineageViewProps = {
+  const lineageViewProps = {
     container: containerRef,
     data: lineageData,
+    hierarchy: hierarchy,
```
frasercl: This and several others can just get implicit names if you want: `hierarchy: hierarchy,` -> `hierarchy,`

frasercl (approved):

src/components/Tabs/Lineage/LineageTab.tsx
ShrimpCryptid: Fixing this, thanks!

src/components/Tabs/Lineage/constants.ts
ShrimpCryptid: This constant gets imported in the track detail view later. I do agree that the mostly-empty constants file is a little silly... this value might move to `colorizer/constants` later if I do a refactor I'm planning on.

PR #963, opened 2026-07-17
refactor: Calculate coparents (lineage pt. 5a)

ShrimpCryptid:
Problem
=======
Part 5a of 7?? for the lineage tree feature. 

This PR adds simple calculations for determining groupings of coparents, and updates a few other utility methods with bugfixes.

Solution
========
- Changed track length calculation to be based on duration and not by number of IDs.
- Return null when default zoom transform cannot be calculated.
- Added calculation for coparents, parents that share direct children in the case of merge nodes.
  - Added unit tests.
- Added function for filtering lineage data to a subset for visualization purposes.

## Type of change

-----------------------
1. `types.ts`
2. `lineage_utils.ts`


src/components/Tabs/Lineage/lineage_utils.ts
```diff
@@ -19,21 +19,31 @@ export function getLineageData(dataset: Dataset): LineageData {
   }
 
   const allTracks = new Set<number>();
-  const trackIdToTrackInfo = new Map<number, TrackInfo>();
+  const trackToTimeMinMax = new Map<number, { min: number; max: number }>();
```
ShrimpCryptid, line 22: Fix here! Before, the duration (length) of a track was calculated based on the number of IDs assigned to it. However, some tracks can be missing IDs for certain timesteps, which means that they are present for more timesteps than the number of IDs. 

This changes the length calculation to be based on the range of timesteps the track lasts for.

src/components/Tabs/Lineage/types.ts
```diff
@@ -24,4 +24,10 @@ export type LineageDataRelationships = {
   multiparentEdges: [number, number][];
   /** A map from a track ID to its parent track IDs. */
   idToParents: Map<number, number[]>;
+
+  /**
+   * Maps from an ID to a set of its co-parents (other direct parent of its
+   * direct children, which happens in the case of merge nodes).
+   */
+  idToCoparents: Map<number, Set<number>>;
```
ShrimpCryptid, line 33: There are a set of rules I set up for the track detail view in a later PR, which manage how tracks are collapsed and expanded. If a node is expanded, all of its parents must also be visible. If it is hidden, all of its children must also be hidden.

This gets slightly complicated by the data provided by Max, where it's possible for a child node to have multiple parents. If you expand/collapse a parent node that shares a child with another parent node (coparent), that parent should also be expanded/collapsed. 

This new logic will be used in the next two PRs for governing these interactions.

src/components/Tabs/Lineage/lineage_utils.ts
```diff
@@ -153,3 +194,33 @@ export function getTreeHierarchy(
 
   return root;
 }
+
+/**
+ * Returns only the subset of lineage data that includes the specified track
+ * IDs and their related parents and children.
+ */
```
ShrimpCryptid, line 206: Used for expanding/collapsing the tree in a later PR. Filtering the existing tree is important for keeping the ordering of leaves the same!

ascibisz (approved):
looks good to me!

frasercl (approved):
Looks good!

PR #971, opened 2026-08-03
feat: Keep selected tracks visible (lineage pt. 8)

ShrimpCryptid:
Problem
=======
Added an automatic zoom framing in two track views that shows where newly-selected tracks are in the lineage tree. Also fixed a bug where the track detail view was not framed correctly on initial render.

Note: There's a known bug where the track detail view doesn't always respond. (edit: this gets fixed in #978  !)

*Estimated review size: medium, 25-30 minutes*

Solution
========
- Updated zoom handlers to run in the next React render after SVG rendering is performed.
  - Fixed bug where initial zoom ran when SVG elements had not yet been rendered.
- Added behavior where, on new track selection, the lineage views will update the zoom + pan to reframe the new selection onscreen.
  - Added associated utility methods + hook.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-971/viewer?dataset=https://vast-files.int.allencell.org/users/peyton.lee/liberali-data/tfe/manifest.json&tab=lineage&track=58
2. Zoom into the different lineage views. Click on different tracks on each-- the other view will update to match.
3. Click on tracks in the main viewport. The two lineage view panels will also pan to keep the new tracks in view.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/754f1cdb-dd94-409e-a24f-e837fbd23de3

{10B22E68-964B-4F2A-AE6A-93A1BDC5FA20}


src/components/Tabs/Lineage/types.ts
```diff
@@ -32,3 +34,10 @@ export type LineageDataRelationships = {
    */
   idToCoparents: Map<number, Set<number>>;
 };
+
+export type LineageNodeSelection = d3.Selection<
+  SVGGElement | d3.BaseType,
+  d3.HierarchyPointNode<TrackInfo>,
+  SVGGElement,
+  TrackInfo
+>;
```
ShrimpCryptid, line 43: Consolidated from `TreeLineageView` and `TrackDetailLineageView`

src/components/Tabs/Lineage/LineageViews/TrackDetailLineageView.tsx
```diff
@@ -485,10 +492,14 @@ export default function LineageTrackDetailView(props: TrackDetailLineageViewProp
   const svgRef = useRef<SVGSVGElement>(null);
   const groupRef = useRef<SVGGElement>(null);
   const nodeGroupRef = useRef<SVGGElement>(null);
-  const nodeSelectionRef = useRef<NodeSelection | undefined>(undefined);
+  const nodeSelectionRef = useRef<LineageNodeSelection | undefined>(undefined);
 
   const trackIds = useMemo(() => new Set(props.selectedTracks.keys()), [props.selectedTracks]);
 
+  // Flag that triggers a zoom reset once the tree completes an initial render.
+  const [hasRenderedTree, setHasRenderedTree] = useState(false);
```
ShrimpCryptid: This fixes the bug where the initial zoom of the viewport was incorrect.

src/components/Tabs/Lineage/LineageViews/TrackDetailLineageView.tsx
```diff
@@ -548,6 +559,7 @@ export default function LineageTrackDetailView(props: TrackDetailLineageViewProp
     d3
       .zoom<SVGSVGElement, unknown>()
       .scaleExtent([0.1, 10])
+      .interpolate(d3.interpolate)
```
ShrimpCryptid, line 551: This changes the interpolation method d3 uses when updating the zoom-- `d3.interpolate` causes it to use linear interpolation, instead of the "fancier" (but more jarring) default transition style that zooms in and out.

src/components/Tabs/Lineage/LineageViews/TrackDetailLineageView.tsx
```diff
@@ -503,26 +515,13 @@ export default function LineageTrackDetailView(props: TrackDetailLineageViewProp
...
-
-  const newTracks = useMemo(() => {
-    const newTracks = new Set<number>();
-    for (const trackId of props.selectedTracks.keys()) {
-      if (!prevTracks.current.has(trackId)) {
-        newTracks.add(trackId);
-      }
-    }
-    return newTracks;
-  }, [props.selectedTracks]);
```
ShrimpCryptid, line 518: Copied this logic into a repeatable hook so it can be used in `TreeLineageView` as well!

src/components/Tabs/Lineage/LineageViews/TrackDetailLineageView.tsx
```diff
@@ -573,12 +573,15 @@ export default function LineageTrackDetailView(props: TrackDetailLineageViewProp
     }
     const initialTransform = getDefaultZoomTransform(svgNode, gNode);
     if (initialTransform) {
-      zoom.current.transform(svg, initialTransform);
+      if (durationMs === 0) {
+        // Apply immediately
+        zoom.current.transform(svg, initialTransform);
+      } else {
+        svg.transition().duration(durationMs).call(zoom.current.transform, initialTransform);
```
toloudis, line 581: why bother? 😉 
I only say this to ask whether extra complexity is worth the benefit

toloudis (approved):
Two main thoughts:
1. if you suddenly had to port this code to a new framework that is not React, how hard would it be?  This can guide you around deciding what should be in the tsx files vs non-component business logic files.
2. it could be interesting to just make a .ts file that handles all the svg updating logic, and semi-hides it in a class abstraction like LineageTreeDrawable. maybe only exposing the root node of the svg

src/components/Tabs/Lineage/LineageViews/TrackDetailLineageView.tsx
ShrimpCryptid: My gut instinct is that the little bit of animation makes it easier to contextualize where you are on the tree? But if anyone hates it it'll be easy to remove 😅

src/components/Tabs/Lineage/LineageViews/TreeLineageView.tsx
```diff
@@ -234,6 +245,20 @@ export default function TreeLineageView(props: TreeLineageViewProps): ReactEleme
...
+  useEffect(() => {
+    setNeedsTrackZoomReframe(true);
+    updateTracks(props.selectedTracks);
+  }, [props.selectedTracks]);
+  useEffect(() => {
+    if (needsTrackZoomReframe) {
+      frameTracksInView(svgRef.current, nodeRef.current, newTracks, zoom.current);
+      setNeedsTrackZoomReframe(false);
+    }
+  }, [needsTrackZoomReframe, newTracks]);
```
frasercl, line 259: Making sure I understand the timeline of actions this sets up:

1. `props.selectedTracks` changes, triggers the first effect. Component re-renders.
2. `setNeedsTrackZoomReframe` triggers a second re-render, for the sole purpose of triggering the second effect. `frameTracksInView` runs.
3. Calling `setNeedsTrackZoomReframe` to reset `needsTrackZoomReframe` back to `false` triggers a third re-render, which does nothing.

Surely there's a better way? At the very least, `needsTrackZoomReframe` can be an incrementing counter or a boolean that toggles only once and triggers effect number 2 on both its rising and falling edge, to avoid the redundant third re-render. But why does the action inside effect number 1 need to happen in an effect? Effects already run after the DOM is updated, I thought. `updateTracks` just sets a ref. Is there no way this can be just one effect, that responds to DOM update after just one component render rather than two (and a redundant third)?

ShrimpCryptid: I think you're right that we can reduce it to 2 React renders, but I don't think it's feasible to reduce it further...

I tried turning step 1 (the SVG render) into a `useMemo`, and got a bunch of strange rendering artifacts as a result. It needs the refs for the SVG and G nodes to be set anyways, so I think it makes sense for it to be a render still. 

The zoom reset functions have to happen in a render after the SVG render, since they rely on the SVG nodes that are created during that render to be actually present on the DOM for measuring their rects. I did change the flag into a number, so we don't incur the third render on the boolean reset.

frasercl (approved):
I'm liking what I see in #978 and don't have any other big notes here!

PR #980, opened 2026-08-26
feat: Expand/collapse nodes in track detail view (select all pt. 3)

ShrimpCryptid:
Problem
=======
Extends #986 with additional options for the track detail view in the lineage graph! You can now right click to expand or collapse the children of a node.

Solution
========
- Also added the ability to expand or collapse all children in `TrackDetailLineageView`.
- Added helper methods for expanding children of a node.

## Type of change

* New feature (non-breaking change which adds functionality)

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-980/viewer?dataset=https%3A%2F%2Fvast-files.int.allencell.org%2Fusers%2Fpeyton.lee%2Fliberali-data%2Ftfe%2Fmanifest.json&track=486%3A0%2C566%3A1%2C561%3A2&t=649&tab=lineage
2. Right click on any node in the detail view on the bottom. Click the option to expand all children.
3. Right click again. There should now be an option to collapse all children instead.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/d3a454b4-e4be-4e77-b171-15310f94b718

{F42E1929-F952-4195-8B66-FD7794723D06}

PR #983, opened 2026-08-28
feat: Add option to change coloring mode for lineage group selections (select all pt. 4)

ShrimpCryptid:
Problem
=======
Extension off of #980, "feat: Select all parents/children."

This change adds the option to control how TFE colors selections of parent or child subtrees. By default, each selection is given a "random" color from the track selection palette (this is actually a repeating cycle of colors, but they're intended to feel random so it's easy to distinguish each track). 

However, I anticipate scientists wanting to see, say, all children of one track colored with one color, and all children of another track colored a different color. This change adds a new setting that causes the default colors to be overridden, and new colors, one for each group of tracks, to be used instead.

Solution
========
- Added a new state field, `colorTracksByGroup`, and a related setter.
  - When enabled on a dataset that has lineage data, the `trackColors` and `isSelectedLut` are updated to show to color groups.
  - Updated other track setters to also update `trackColors` and `isSelectedLut` with overridden color behaviors.
- Added a checkbox in the LineageGraphTab and SegmentationSettings for enabling/disabling the behavior.
- Added URL serialization.
- Added unit tests.

## Type of change

Steps to Verify:
----------------
1. Open PR preview: https://allen-cell-animated.github.io/timelapse-colorizer/pr-preview/pr-983/viewer?dataset=https%3A%2F%2Fvast-files.int.allencell.org%2Fusers%2Fpeyton.lee%2Fliberali-data%2Ftfe%2Fmanifest.json&track=track=486%3A0%2C566%3A1%2C561%3A2%2C482%3A3%2C560%3A4%2C483%3A5%2C568%3A6%2C571%3A7%2C427%3A8%2C487%3A9%2C567%3A10%2C554%3A11%2C556%3A0%2C426%3A1&t=649&tab=lineage
2. Enable the "Color by track groups" checkbox. All adjacent selected tracks will share the same color.
3. Click on any node that's connected to the selected groups. The new nodes will also have the same colors as the rest of the group.

Screenshots (optional):
-----------------------
**Video talk-through (🔊):**

https://github.com/user-attachments/assets/18ebb46f-de1a-49ca-874c-ba30577125e6

https://github.com/user-attachments/assets/4342919a-4a3b-4310-8aff-4b46dd0177d1

| Before | After |
| --- | --- |
| {CCC3CA34-2148-40DD-AB23-F9012A1E403E} | {DE734A49-FFB2-4F91-85B0-7392ABC162E9} |

{0F7DBCF8-DFC0-438A-AF08-13F3B3C4F31D}

{84FC43BA-F290-4821-A1C0-EB039730884F}

Keyfiles:
-------
1. `lineage_utils.ts`
2. `track_slice.ts`


src/components/Tabs/Plots/LineageGraph/lineage_utils.ts
```diff
@@ -433,6 +435,16 @@ export function getLineageContextMenuItems(
         onClick: hoveredId !== null ? () => callbacks.deselectNodeAndChildren(hoveredId) : undefined,
         visible: areTrackAndAllChildrenSelected,
       },
+      {
+        label: "Selection options...",
+        children: [
+          {
+            label: (!data.applyTrackColorToRelatives ? "☑ " : "☐ ") + "Use random outline colors",
+            onClick: () => callbacks.setApplyTrackColorToRelatives(!data.applyTrackColorToRelatives),
```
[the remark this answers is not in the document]
ShrimpCryptid: Unfortunately I don't think Ant has an API for setting these properties on individual menu items....

src/components/Tabs/Plots/LineageGraph/LineageGraphTab.tsx
```diff
@@ -142,15 +145,25 @@ export default function LineageGraphTab(props: LineageGraphTabProps): ReactEleme
       const trackIds = Array.from(trackIdSet);
 
       if (selected) {
-        const tracks = trackIds
+        const trackObjects = trackIds
           .map((id) => dataset?.getTrack(id))
           .filter((track): track is Track => track !== undefined);
-        addTracks(tracks);
+        let colorId: number | undefined;
+        if (applyTrackColorToRelatives) {
```
meganrm: this means they all have the same color? I might name this something slightly clearer. Like "relativesInheritColor", "sameColorRelatives". I know that's what this technically says, but I get lost in the sauce of it a bit

src/components/Tabs/Plots/LineageGraph/LineageGraphTab.tsx
```diff
@@ -142,15 +145,25 @@ export default function LineageGraphTab(props: LineageGraphTabProps): ReactEleme
...
+          const { tracks, trackToColorId } = useViewerStateStore.getState();
+          // Get base track color
+          const baseTrackColorId = trackToColorId.get(trackId) ?? getNextColorId(tracks, trackToColorId);
+          colorId = applyTrackColorToRelatives ? baseTrackColorId : undefined;
```
meganrm: aren't we already in a 'if (applyTrackColorToRelatives)' block? so wouldn't this always be true?

src/components/Tabs/Plots/LineageGraph/LineageGraphTab.tsx
```diff
@@ -142,15 +145,25 @@ export default function LineageGraphTab(props: LineageGraphTabProps): ReactEleme
...
+          // Remove any existing track colors if all relatives should have the
+          // same color
+          removeTracks(trackIds);
+        }
+        addTracks(trackObjects, colorId);
```
meganrm: to make this clearer, I think I would just send in undefined. or add a comment. You don't really to define colorId outside this block

src/state/slices/config_slice.ts
```diff
@@ -331,6 +342,8 @@ export const createConfigSlice: StateCreator<ConfigSlice, [], [], ConfigSlice> =
         outlinePaletteKey: key,
       };
     }),
+  setApplyTrackColorToRelatives: (applyTrackColorToLineageSelection) =>
+    set({ applyTrackColorToRelatives: applyTrackColorToLineageSelection }),
```
frasercl: Why is the name of the argument different than the name of the property it gets set to?

src/state/slices/config_slice.ts
```diff
@@ -207,6 +214,7 @@ export type ConfigSliceActions = {
   setEdgeMode: (edgeMode: DrawMode) => void;
   setOpenTab: (openTab: TabType) => void;
   setPlotTab: (plotTab: PlotTabType) => void;
+  setApplyTrackColorToRelatives: (applyTrackColorToLineageSelection: boolean) => void;
```
frasercl: Same question

frasercl (changes requested):
I was also a bit puzzled that switching the outline color settings didn't immediately apply to previous selections. Maybe that makes sense with the UX though? If users are interested in applying a uniform color to one lineage subtree while applying random colors to another, it might be right. Otherwise, I'd expect computation of track colors to behave like a `useMemo` or `useEffect`, which could trigger the relevant computation when `applyTrackColorToRelatives` or the set of selected tracks change. Can that sort of computation be decoupled from the rest of the selection though? Can it be determined as a pure function from the set of selected tracks and the value of `applyTrackColorToRelatives`?

src/state/slices/track_slice.ts
```diff
@@ -33,8 +33,9 @@ export type TrackSliceActions = {
   /**
    * Adds one or more tracks to the current track selection.
    * @param tracks The track or array of tracks to add.
+   * @param colorIdx Optional color index to assign to the added tracks.
    */
-  addTracks: (tracks: Track | Track[]) => void;
+  addTracks: (tracks: Track | Track[], colorIdx?: number) => void;
```
frasercl, line 39: Another spot where type and value have confusingly different arg names. Thought for a second that `colorOverride` below was a 24-bit color as a single `number`.

frasercl (commented):
I hit "request changes" by accident just now, so here's a second review to make the scary red icon go away.

frasercl (approved):
Perhaps this will work?

src/state/slices/track_slice.ts
```diff
@@ -168,41 +263,46 @@ export const createTrackSlice: StateCreator<TrackSlice & ConfigSlice, [], [], Tr
...
-    tracks = Array.isArray(tracks) ? tracks : [tracks];
-
-    // Ensure colors array matches length of tracks
-    colors = colors ?? [];
-    while (colors.length < tracks.length) {
-      // Fill with ascending indices; will be wrapped to the palette size
-      colors.push(colors.length);
-    }
+  setTracks: (tracks: Track | Track[], colors?: number[]) =>
+    set((state) => {
```
ShrimpCryptid: Turn off whitespace here! I wrapped this in the `set` function a little earlier than before, so I could access current state.

src/colorizer/utils/lineage_utils.ts
```diff
@@ -283,3 +285,45 @@ export function getDescendants(
...
+    const allGroups = Array.from(new Set([...parentGroups, ...childGroups]));
+
+    if (allGroups.length === 0) {
+      // Not in contact with an existing group, so create a new one
+      const newGroup: Set<number> = new Set([trackId]);
+      groups.add(newGroup);
+      idToGroup.set(trackId, newGroup);
+    } else {
+      // Merge all groups into an existing one. Existing parent groups take
+      // priority.
```
ShrimpCryptid, line 319: I've been a little too deep in the algorithm sauce.... I think technically the correct user behavior here would be to merge into the oldest set, but it would add more complexity + more computation here. So this is fine!

src/components/Tabs/Plots/LineageGraph/LineageGraphTab.tsx
```diff
@@ -193,8 +185,18 @@ export default function LineageGraphTab(props: LineageGraphTabProps): ReactEleme
...
+          <Checkbox checked={colorTracksByGroup} onChange={(e) => setColorTracksByGroup(e.target.checked)}>
+            <LabelWithHint
+              hintProps={{
+                title: "Groups of related tracks will use the same color when selected in the graph and viewport.",
+              }}
+            >
+              Color by track groups
+            </LabelWithHint>
+          </Checkbox>
+        </FlexRowAlignCenter>
```
ShrimpCryptid, line 199: Image

src/components/Tabs/Settings/SegmentationSettings.tsx
```diff
@@ -1,8 +1,11 @@
+import { Tooltip } from "antd";
```
ShrimpCryptid, line 1: Image

---

Image

src/state/slices/track_slice.ts
```diff
@@ -1,9 +1,23 @@
 import type { Color } from "three";
 import type { StateCreator } from "zustand";
 
-import { type ColorRamp, MAX_FEATURE_CATEGORIES, SelectionOutlineColorMode, type Track } from "src/colorizer";
+import {
+  type ColorRamp,
+  Dataset,
```
frasercl: Lint warns this import is only used as a type

